PR #32, opened 2013-06-22
Header files of packages gazebo_ros and gazebo_plugins are installed to the wrong location

meyerj:
The packages gazebo_ros and gazebo_plugins install their header files to `${CATKIN_PACKAGE_INCLUDE_DESTINATION}` as in the catkin example at [ros.org](http://ros.org/wiki/catkin/CMakeLists.txt#Optional_Step:_Specifying_Installable_Directories), but as the include directory contains another subdirectory named gazebo or gazebo_plugins headers are installed to
- include/gazebo_ros/gazebo/gazebo_ros_api_plugin.h
- include/gazebo_plugins/gazebo_plugins/gazebo_ros_camera_utils.h
- ...

This patch changes the installation target directory to `${CATKIN_GLOBAL_INCLUDE_DESTINATION}`, so that header files end up in
- include/gazebo/gazebo_ros_api_plugin.h
- include/gazebo_plugins/gazebo_ros_camera_utils.h
- ...

We are using the `GazeboRosCameraUtils` class in our [hector_gazebo](http://www.ros.org/wiki/hector_gazebo) stack and include the header using the `#include <gazebo_plugins/gazebo_ros_camera_utils.h>` which would fail without this patch.

See also my question regarding installation of header files in catkin packages on [ROS answers](http://answers.ros.org/question/65716/which-is-the-correct-way-to-install-header-files-in-catkin-packages/).


meyerj:
Recreated as #35.

PR #97, opened 2013-08-07
gazebo_ros_openni_kinect plugin: adds publishing of the camera info again (fixes #95)

bit-pirate:

PR #98, opened 2013-08-07
now explicitly publish camera info on new depth frame

piyushk:
based on @bit-pirate's comment in https://github.com/ros-simulation/gazebo_ros_pkgs/issues/95


piyushk:
I just realized that #97 has the same fix. closing this pull request.

PR #109, opened 2013-09-12
add OGRE-Paging as dependency since gazebo is doing paging.

hsu:


gazebo_plugins/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ include (FindPkgConfig)
 if (PKG_CONFIG_FOUND)
   pkg_check_modules(XML libxml-2.0)
   pkg_check_modules(OGRE OGRE)
+  pkg_check_modules(OGRE-Paging OGRE-Paging)
```
nkoenig: Use OGRE-Terrain. It's more inclusive than OGRE-Paging

PR #120, opened 2013-09-25
Segfaults in camera gazebo plugins due to uninitialized shared pointers

meyerj:
The new shared pointer members introduced in bd078a095da8723347a26fbb007c64bbf49dbb22 (`image_connect_count_`, `image_connect_count_lock_` and `was_active_`) have not properly been initialized in most camera plugins except gazebo_ros_camera.cpp and gazebo_ros_multicamera.cpp. Gazebo therefore segfaults when someone connects to an image subscriber.

cb950af9410e80b833dc5749eac6fbeb900b4cbd added initialization of `image_connect_count_` in the openni plugin, but the other two pointers have still been left uninitialized. The pointers must be initialized before `GazeboRosCameraUtils::LoadThread()` is executed, as they are dereferenced in `GazeboRosCameraUtils::ImageConnect()`.

Additionally in the multi camera plugin the shared pointers have been constructed from pointers to member variables, which is forbidden as the shared_ptr cannot take ownership of this pointer.

I only tested this patch with the gazebo_ros_camera and gazebo_ros_openni_kinect plugins!

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_multicamera.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_multicamera.h
@@ -54,9 +54,9 @@ namespace gazebo
     /// Bookkeeping flags that will be passed into the underlying
     /// GazeboRosCameraUtils objects to let them share state about the parent
     /// sensor.
-    private: int imageConnectCount;
-    private: boost::mutex imageConnectCountLock;
-    private: bool wasActive;
+    private: boost::shared_ptr<int> image_connect_count_;
+    private: boost::shared_ptr<boost::mutex> image_connect_count_lock_;
+    private: boost::shared_ptr<bool> was_active_;
   };
 }
 #endif
--- a/gazebo_plugins/src/gazebo_ros_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera.cpp
@@ -66,11 +66,7 @@ void GazeboRosCamera::Load(sensors::SensorPtr _parent, sdf::ElementPtr _sdf)
   this->depth_ = this->depth;
   this->format_ = this->format;
   this->camera_ = this->camera;
-  this->image_connect_count_ = boost::shared_ptr<int>(new int);
-  *this->image_connect_count_ = 0;
-  this->image_connect_count_lock_ = boost::shared_ptr<boost::mutex>(new boost::mutex);
-  this->was_active_ = boost::shared_ptr<bool>(new bool);
-  *this->was_active_ = false;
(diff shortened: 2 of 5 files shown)


j-rivero:
I was able to setup a catkin workspace using this pull request and build the whole drcsim software stack. Launch atlas_v3 allowed me to visualize images from both situational cameras and stereo.

The problem is that I was not able to reproduce the segfault on my system with master version of this repository.

I'm running quantal, ros-groovy and gazebo-2.0.0.

meyerj:
Does atlas uses the ros_openni_depth_camera or ros_openni_kinect plugin? It is quite obvious when looking at the source of [gazebo_ros_depth_camera.cpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/9a61cfa53117ffab043bc7d4d1bf0da2f426885f/gazebo_plugins/src/gazebo_ros_depth_camera.cpp) that the `image_connect_count_` shared pointer is used but never initialized. That's why I moved the initialization to [gazebo_ros_camera_utils.cpp](https://github.com/meyerj/gazebo_ros_pkgs/blob/f75b0c302bbf1127c70100a07e5349065a58cb8a/gazebo_plugins/src/gazebo_ros_camera_utils.cpp) where the pointers actually are defined. gazebo_ros_camera and gazebo_ros_multicamera already had their own initializations and therefore probably worked even without this patch.

I am using precise with ROS hydro and Gazebo 1.9.1, but this bug should affect all versions that use the hydro-devel branch of this repo.

fmder:
I do have this same segfault with Gazebo 1.9.1.

j-rivero:
@meyerj you are right. Atlas, AFAIK, does not use openni_kinect but camera and multicamera plugins. So yes, they are working with both, patch and hydro-devel version. And indeed, `image_connect_count_` looks to me that is being used with being initialized. Summarizing, all I can confirm with my test is that the patch is not breaking current camera  and multicamera plugins.

@fmder can you confirm that the bug is fixed for you with this patch?

fmder:
@j-rivero I just finished testing everything, the patch do solve the problem.

fmder:
This bug also affects groovy-devel, as hydro-devel has been merged in groovy lately

j-rivero:
The plan is not to use groovy-devel any more and kill it in very few hours, since the release of the 2.3.3 version will use the same code than hydro-devel. More details are in [this pull request](https://github.com/ros-gbp/gazebo_ros_pkgs-release/pull/2).

So the bug should be fixed for both version. To be released in 2.3.3, hopefully in few hours.

PR #124, opened 2013-10-03
Env hook so that ROS packages can export gazebo model paths

jonbinney:
I've been trying to find a clean way for ROS packages which contain gazebo models to have the right directory added to GAZEBO_MODEL_PATH. (It's possible there's an obvious way to do this that I'm missing). I think this solution is pretty reasonable though. The end result is that ROS packages can just specify their gazebo models directory in their package.xml, and when the user sources devel/setup.sh, the right directories will be added to GAZEBO_MODEL_PATH.

How this works:

Any ROS package which has gazebo models does two things: it depends on gazebo_ros, and it puts the following in its package.xml:

``` xml
 <export>
    <gazebo_ros gazebo_model_path="my_models_directory"/>
  </export>
```

To take advantage of this, i've added a catkin_env_hook in the gazebo_ros CMakeLists.txt which tells catkin that when sourcing devel/setup.sh, it should also source gazebo_ros/env-hooks/setup_gazebo_env.sh. That then calls the make_models_path python script, which looks at all packages which depend on gazebo_ros, pulls out the appropriate export tags, and creates the GAZEBO_MODEL_PATH.

I realize that this may seem a bit complicated, but I haven't found any good alternative, and it enables a very nice workflow. ROS users can create packages that contain gazebo models, specify it in a clean way in their package.xml, and then when they source the standard catkin setup script, the gazebo environment variables will be set appropriately.

@davetcoleman, does this sound useful to you? I'm open to other suggestions about how to do this.


scpeters:
This looks like a nice feature. We currently provide this in drcsim through a [setup.sh script](https://bitbucket.org/osrf/drcsim/src/ec7d7c63e7b9d92a60edfa5e7057b9acbfebb570/drcsim_model_resources/cmake/setup.sh.in?at=default), which works but is easy to forget. It would absurd, though, for every ROS package to require sourcing a setup.sh

One question: any thoughts about GAZEBO_RESOURCE_PATH or GAZEBO_PLUGIN_PATH?

jonbinney:
I guess you could have a similar tag in the export section for gazebo_resource_path and for gazebo_plugin_path - i just did the model path right now but i can add the others if we think this is a good idea.

scpeters:
We can start with the model path and add the others if it seems to be working well. I was just mentioning the others since they are used in the drcsim setup.sh script.

jonbinney:
What is the GAZEBO_RESOURCE_PATH used for?

scpeters:
I know GAZEBO_RESOURCE_PATH is used for world files, possibly textures. I'll have to check with @nkoenig

jonbinney:
There's also the question of whether the paths in the package.xml's should override any existing values for these variables, or just be appended to them. Normally the setup.sh overrides any existing ROS environment variables, so thats how i've done things so far, but i could see use cases where appending might be preferable.

scpeters:
For GAZEBO_MODEL_PATH, I think appending is preferable.

jonbinney:
Yeah, i think you may be right. The downside of appending is that it can cause unexpected things to happen. If someone switches from one ROS workspace to another, they'll have leftover model paths from the old workspace.

scpeters:
I guess I could go either way on appending vs overriding.

I currently have bitbucket.org/osrf/gazebo_models cloned for my local machine and I add that folder to GAZEBO_MODEL_PATH in my `.bashrc`. I would still want that in the GAZEBO_MODEL_PATH variable after invoking the catkin script.

jonbinney:
Makes sense. I'll modify this pull request to do appending, and to handle the resource path.

jonbinney:
Changed a few things:
- now also handling GAZEBO_RESOURCE_PATH and GAZEBO_PLUGIN_PATH
- directories supplied by ros packages are now prepended to any existing path, instead of replacing it

scpeters:
Sounds good, though I haven't had a chance to test yet.

jonbinney:
@jon-weisz - you've used ROS with gazebo - how did you handle the setting the gazebo paths from ROS packages?

jon-weisz:
I used gazebo 1.5 . The newer version seems to have changed the 

{simulator_gazebo|gazebo_ros_pkgs}/gazebo/scripts/gazebo script and eliminated the setup.sh

The last line in the old version did this:

```
. `rospack find gazebo`/scripts/setup.sh && `rospack find gazebo`/gazebo/bin/gzserver $final
```

Which used to set up the relevant resource paths by doing this:

```
export GAZEBO_RESOURCE_PATH=`rospack plugins --attrib="gazebo_media_path" gazebo | awk 'BEGIN { getline; printf $2 }; { printf ":"$2 }'`
export GAZEBO_PLUGIN_PATH=`rospack plugins --attrib="plugin_path" gazebo | awk 'BEGIN { getline; printf $2 }; { printf ":"$2 }'`
export OGRE_RESOURCE_PATH=`rospack find ogre`/ogre/lib/OGRE
```

Now, the gazebo/share/gazebo/setup.bash seems like it does something like that:
https://bitbucket.org/osrf/gazebo/src/4f31431d1ddc0901ed3dba9ab7852ed09ce99304/cmake/setup.sh.in?at=gazebo_1.9 .

It looks to me like that does not append, and instead replaces the current paths.

Maybe add a new setup.sh that calls  gazebo/share/gazebo/setup.bash and then uses rospack plugins to append the way it used to for all of those paths?

jonbinney:
Hmmm... just found the gazebo_ros_paths plugin in gazebo_ros: https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_ros/src/gazebo_ros_paths_plugin.cpp

It appears to look for exports of gazebo paths in package.xml's, and add them to the gazebo paths. So the env hook stuf in this pull request is probably unnecessary. @davetcoleman - does that sound right to you?

davetcoleman:
Sorry for coming late into this conversation. This PR seems like a great idea and would remove the need for users to export an environmental variable/edit their .bashrc whenever they download a new robot_description package. It replaces the [Environmental Variable](http://gazebosim.org/wiki/Tutorials/1.9/Using_roslaunch_Files_to_Spawn_Models#Environment_Variable) section of the tutorials I wrote.

The `gazebo_ros_paths` plugin I'll admit I've never fully understood, even though I cleaned up that script a lot this past summer. @hsu made it and would understand it much better, but it seems like that script might be where the functionality of this pull request should go.

jon-weisz:
Based on that code, and that plugin being loaded in already by the gazebo
launch script, it seems to me that your export should already have worked,
without any env hooks. Is there a specific case where it was failing?

jonbinney:
It looks to me like that `gazebo_ros_paths` plugin might actually be a cleaner way to do things. It has all the advantages of this pull requrest, but additionally, it avoids adding the paths for ROS packages to the actual gazebo environment variables. This is good because there won't be leftover paths if you switch between ROS workspaces.

@jon-weisz i didn't even check whether it worked without the env hook, since I didn't know that functionality existed. I'll try it now, and if it works, I'll add something about it in the tutorial.

@hsu do you see any problems with using the `gazebo_ros_path` plugin as the main way for ROS packages to get their models and plugins into gazebo?

jonbinney:
Hmmm... the `gazebo_ros_paths` plugin doesn't seem to be expanding paths exported by packages. So when i put the following in my package.xml:

```
 <export>
    <gazebo_ros gazebo_model_path="my_models_directory"/>
  </export>
```

It literally adds "my_models_directory" to the GAZEBO_MODEL_PATH, instead of resolving it relative to the package directory. Doesn't seem like the right behavior.

davetcoleman:
I haven't myself seen this feature used, so perhaps its broken and this should be fixed in it.

jonbinney:
I'll take a stab at fixing it, and then do a new pull request. I'm closing this pull request since the plugin seems like a better way to go.

jon-weisz:
@jonbinney: You're looking for the awesomely documented `${prefix}` macro

i.e. 

```
 <export>
    <gazebo_ros gazebo_model_path="${prefix}/my_models_directory_relative_to_this_package" />
  </export>
```

https://github.com/ros/rospack/blob/groovy-devel/src/rospack.cpp#L2142

jonbinney:
@jon-weisz  I WAS looking for that! @dirk-thomas the `${prefix}` syntax isn't documented anywhere that I could find. Is it an official part of the package.xml format?

I think I still need to fix the `gazebo_ros_paths` plugin to use GAZEBO_RESOURCE_PATH instead of GAZEBO_MEDIA_PATH.

jon-weisz:
@jonbinney http://docs.ros.org/independent/api/rospkg/html/manifest_xml.html#id38

jonbinney:
Yeah its in the old manifest.xml docs but not the package.xml docs

jonbinney:
Hmmm.... also realizing that`${prefix}` is probably going to have problems with the plugin paths on installed catkin packages. Ideally plugins would be install to `/opt/ros/hydro/lib/<package_name>`, but in installs `${prefix}` will resolve to the share directory.

dirk-thomas:
`${prefix}` should resolve to the folder where the plugin manifest is stored in.

jonbinney:
@dirk-thomas i meant gazebo plugins, not ROS plugins. Do gazebo plugins also have a plugin manifest? From what i read it sounded like the compiled libraries need to be directly in the GAZEBO_PLUGIN_PATH.

dirk-thomas:
I thought you meant ROS plugins - for Gazebo I don't know anything about it...

jonbinney:
Yeah, but it brings up a need within the ROS package.xml. Before, adding `${prefix}` let packages export paths relative to the source directory, and also worked in the install space. Now, it will only work for paths that will end up in the share directory after installation. This is similar to the `$(find )` problem that roslaunch xml syntax has.

PR #129, opened 2013-10-11
Spawn model pose fix

etsardou:
Altered function GazeboRosApiPlugin::updateSDFAttributes in order to apply the user-defined robot pose (given by arguments when spawninga model) relatively to the sdf file where the model is included.

For example when a robot had an sdf pose `[0 0 0.18 0 0 0]` and when spawning it the args `-x 3 -y -3 -z 0.5` were given the following error occured : 

`Error [Param.cc:181] Unable to set value [0 0 0.180 0 0 03 -3 0.5 0 -0 0] for key[pose]`

Since one can not detect if a value is not set in arguments or it was set equal to 0, I got the two poses and added them relatively. Now when gazebo starts no errors are thrown and the models spawn in place.

If you think this approach is not correct, please inform me!

Manos

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -1984,6 +1984,8 @@ void GazeboRosApiPlugin::updateSDFAttributes(TiXmlDocument &gazebo_model_xml, st
     // Pose is set at bottom of function
   }
 
+	
+
   // Set the pose value for both SDFs types here
   if (pose_element)
   {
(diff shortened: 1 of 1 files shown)


gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1984,6 +1984,8 @@ void GazeboRosApiPlugin::updateSDFAttributes(TiXmlDocument &gazebo_model_xml, st
     // Pose is set at bottom of function
   }
 
+	
+
```
hsu, line 1988: please remove these empty lines

hsu:
The scalar addition will not work correctly in all cases.  Please see pull request #148 and pull request #150

hsu:
Do you have a good test for this?

PR #134, opened 2013-10-24
gazebo_ros_control: Use the model NodeHandle to get the robot_description parameter

meyerj:
I replaced the default NodeHandle `nh_` by the model's NodeHandle `model_nh_` in `GazeboRosControlPlugin::getURDF()`. Otherwise it would not be possible to set a per-robot description on the parameter server. I cannot use the `robotParam` parameter for that as I would like to use the same plugin config for all models. The `robotNamespace` parameter is set automatically by the [spawn_model](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_ros/scripts/spawn_model) script based on the namespace in which the script itself has been called.

Due to the usage of `searchParam()` the plugin still uses the global `/robot_description` parameter if `<robotNamespace>/robot_description` is not set.

### DIFF
--- a/gazebo_ros_control/include/gazebo_ros_control/gazebo_ros_control_plugin.h
+++ b/gazebo_ros_control/include/gazebo_ros_control/gazebo_ros_control_plugin.h
@@ -81,7 +81,6 @@ public:
 protected:
 
   // Node Handles
-  ros::NodeHandle nh_; // no namespace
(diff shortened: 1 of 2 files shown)


meyerj:
Just noticed that this is a duplicate of #112.

PR #135, opened 2013-10-24
gazebo_ros_control: The position and velocity hardware interfaces are now fully supported.

jim-rothrock:
Added support for the position hardware interface. Completed support for the velocity hardware interface.

This pull request depends on ros-controls/ros_control pull request #124 being accepted.


progtologist:
Tested it with the position hardware interface and is functioning perfectly!
Great work!

jim-rothrock:
ros-controls/ros_control pull request #124 has been merged, so this pull request can be merged.

davetcoleman:
pull request #124 does not appear to have been merged but rather closed.

davetcoleman:
Have you considered the need to switch between control modes during runtime? I just added a related feature to ros_control: https://github.com/ros-controls/ros_control/pull/128

jim-rothrock:
ros-controls/ros_control pull request #124 is here: https://github.com/ros-controls/ros_control/pull/124. It was merged, then closed.

jim-rothrock:
> Have you considered the need to switch between control modes during runtime?

No, I didn't know that was possible.

davetcoleman:
Ah, Github syntax auto-linked to the wrong repo. 

Switching control modes during runtime can be added later when the need arises, but yes it is possible and sometimes necessary:
http://wiki.ros.org/ros_control/Tutorials/Switching%20control%20modes%20in%20hardware

davetcoleman:
I don't have the time/initiative to build this but it looks fine to me.

gazebo_ros_control/src/default_robot_hw_sim.cpp
```diff
@@ -209,29 +227,46 @@ class DefaultRobotHWSim : public gazebo_ros_control::RobotHWSim
...
+          sim_joints_[j]->SetForce(0,joint_effort_command_[j]);
+      
+          // Output debug every nth msg
+          if( !(j % 10) && false)
+          {
+            ROS_DEBUG_STREAM_NAMED("robot_hw_sim","SetForce " << joint_effort_command_[j]);
+          }
+          break;
+        case POSITION:
+          sim_joints_[j]->SetAngle(0,joint_position_command_[j]);
```
adolfo-rt: When you send a discrete jump to `SetAngle(...)`, does the joint actually replicate the discrete jump, or are there some underlying dynamics in between (if so, which)?. I can't really tell from the official doc for [setAngle](http://gazebosim.org/api/dev/classgazebo_1_1physics_1_1Joint.html#a4debd0c6857db7e7461103523129f9e4) if this is the case or not.

jim-rothrock:
> When you send a discrete jump to SetAngle(...), does the joint actually replicate the discrete jump, or are there some underlying dynamics in between...

I don't know. @ros-simulation, could someone enlighten us? One hint that dynamics are involved is that joint->SetMaxForce() must be called with an appropriate value in order for joint->SetAngle() and joint->SetVelocity() to work.

My simulated robot is using position-controlled and velocity-controlled joints exclusively, and everything moves as expected. In particular, rotating the drive wheels with joint->SetPosition() or joint->SetVelocity() causes to the robot to move on the ground plane. I was concerned that there might be no interaction with the ground, causing the wheels to skid.

gazebo_ros_control/src/default_robot_hw_sim.cpp
davetcoleman: @hsu

hsu: [`SetAngle(...)`](https://bitbucket.org/osrf/gazebo/src/70b7d55cfaf6aa0f41459814d1a840f2f278fe94/gazebo/physics/Joint.cc?at=default#cl-548) causes a discrete jump without dynamics.  I feel an underlying PID that uses `SetForce(...)` to try and achieve the desired joint position might make a better behavior here.

jim-rothrock: Does SetVelocity() exhibit the same behavior? I.e. does it cause an instantaneous change in velocity, without dynamics?

adolfo-rt: > SetAngle(...) causes a discrete jump without dynamics. I feel an underlying PID that uses SetForce(...) to try and achieve the desired joint position might make a better behavior here.

What we currently do is precisely that. If a joint has a position interface, we look for a PID specification in the ROS parameter server, and bail out if none is found. Here's the [PID loading](https://github.com/pal-robotics/reemc_simulation/blob/hydro-devel/reemc_hardware_gazebo/src/reemc_hardware_gazebo.cpp#L185) and [use](https://github.com/pal-robotics/reemc_simulation/blob/hydro-devel/reemc_hardware_gazebo/src/reemc_hardware_gazebo.cpp#L245) for a position-only robot. I originally intended to generalize and add this to gazebo_ros_control, but have not had the time. We use the PID class from ros_control instead of that from Gazebo because loading a controller configuration from file with the former is really simple.

adolfo-rt: > Does SetVelocity() exhibit the same behavior? I.e. does it cause an instantaneous change in velocity, without dynamics?

I believe from experimentation that it respects the effort limits, because the max force affects the max accelerations you can achieve. I haven't checked the underlying implementation.

jim-rothrock: I'll modify the code in this pull request so that position-controlled and velocity-controlled joints use PID controllers instead of calling SetAngle() or SetVelocity(). The PID parameters will be read from the ROS Parameter Server.

jim-rothrock:
PID controllers are now used instead of calling SetAngle() or SetVelocity(). The PID gain values are read from the Parameter Server.

adolfo-rt:
A late note. I think that setting velocities directly (without passing through a PID) works rather well in Gazebo. As long as you set a reasonable effort limit (we read it from URDF), it should behave as expected. We control our simulated mobile bases like this and so far have not had any trouble.

jim-rothrock:
The code now falls back to SetAngle() and SetVelocity() if PID gains are not found.

jim-rothrock:
This pull request is ready to be merged whenever someone has a chance to do so.

davetcoleman:
I just noticed you added the PID controllers for position and velocity commands. What was the rationale behind this? Gazebo already has built in PID controllers for these two control modes, and otherwise it seems like PID is in the domain of a controller that runs on top of this gazebo-ros_control interface. I probably already missed the discussion, but are we sure this belongs here?

jim-rothrock:
The PID controllers are for emulating position-controlled and velocity-controlled joints. For example, if a motor is controlled by a Roboteq motor controller, the Roboteq reads the optical encoder directly and has its own PID loop. ROS just sends position or velocity commands to the Roboteq over a USB connection. Originally, I implemented position control by calling joint->SetAngle(), and velocity control by calling joint->SetVelocity(). However, @adolfo-rt and @hsu were concerned that joint->SetAngle() could cause problems because it does not use forces (see the "outdated diff" thread from two months ago). In fact, I have confirmed that if joint->SetAngle() is used to control _any_ of my robot arm's joints, gripper friction does not work when picking up an object. It's still useful for controlling base steering joints, though.

The optional PID controllers allow one to use SetAngle() and SetVelocity() for simplicity, or full PID control if SetAngle() or SetVelocity() are insufficient. Basically, I added PID control to implement what @adolfo-rt described here:

"What we currently do is precisely that. If a joint has a position interface, we look for a PID specification in the ROS parameter server, and bail out if none is found. Here's the PID loading and use for a position-only robot. I originally intended to generalize and add this to gazebo_ros_control, but have not had the time. We use the PID class from ros_control instead of that from Gazebo because loading a controller configuration from file with the former is really simple."

davetcoleman:
Thanks for the explanation, I do recall hearing about his. Essentially, SetAngle() and SetVelocity() "freeze" the Gazebo simulation and artificially move the bodies before continuing. 

While this PR is a good fix for the time being, would you agree that it would be better for Gazebo to either fix its SetAngle() and SetVelocity() functions, or add similar version that have PID controllers in them? The thinking behind this `gazebo_ros_control` package was that it would be a pretty thin pass-through wrapper that exposes Gazebo actuators as ros_control `hardware_interfaces`.

jim-rothrock:
> would you agree that it would be better for Gazebo to either fix its SetAngle() and SetVelocity() functions, or add similar version that have PID controllers in them?

Ideally, there would be a Gazebo controller plugin type that could be attached to a joint. The plugin would receive the values passed to SetAngle() and SetVelocity(), and implement closed-loop control via calls to SetForce(). Even more ideally, such a plugin could also support transmissions, which would at the very least involve overriding GetAngle() in addition to SetAngle() (for reduction gears). Perhaps it is possible to make such a plug-in with Gazebo's existing plugin system, but it was not obvious to me how to do it.

> it would be a pretty thin pass-through wrapper that exposes Gazebo actuators

I think the issue here is that Gazebo's joints are not really "actuators" as I think of them. An actuator might have a control loop and a transmission controlling the joint.

adolfo-rt:
> Gazebo already has built in PID controllers for these two control modes

I believe that the reason why control_toolbox's PIDs are used instead of Gazebo's impementation is that loading the controller configuration (gains and clampings) is straightforward and already implemented for the former (via ROS params).

jim-rothrock:
> I believe that the reason why control_toolbox's PIDs are used instead of Gazebo's impementation is that loading
> the controller configuration (gains and clampings) is straightforward and already implemented for the former (via 
> ROS params).

Yes, and the PID parameters can be tuned with dynamic_reconfigure.

gazebo_ros_control/src/default_robot_hw_sim.cpp
```diff
@@ -209,30 +271,93 @@ class DefaultRobotHWSim : public gazebo_ros_control::RobotHWSim
...
       {
-        ROS_DEBUG_STREAM_NAMED("robot_hw_sim","SetForce " << joint_effort_command_[j]);
+        case EFFORT:
+          {
+            const double effort = joint_effort_command_[j];
+            sim_joints_[j]->SetForce(0, effort);
+            // Output debug every nth msg
+            if( !(j % 10) && false)
+            {
+              ROS_DEBUG_STREAM_NAMED("robot_hw_sim","SetForce " << effort);
```
adolfo-rt: I'd remove this high-frequency log message, even if it's of DEBUG severity.

progtologist: That if statement is actually never true.

adolfo-rt: Correct, I missed the `&& false`. Dead code is an even better reason to justify removal.

adolfo-rt:
I'm hesitant to accept the PR as I don't have the cycles to test all possible combinations of HW interfaces and PID absence/presence combinations, and the package does not have a test suite to help here. @jim-rothrock and others, how extensively have the contents of this PR request been exercised in practice?. The code looks correct, maybe a bit too much conditionals (if and switch) for my taste, but nothing blocking.

progtologist:
I have tested that PR and I am currently using it with position-position, velocity-position and effort-position controllers. I have found no problems so far with that combinations. If others can confirm its' functionality then I would say it can be integrated and if issues arise, fixes can be proposed.

davetcoleman:
I've reviewed the code but not tested it (no time). I agree with
progtologist we should just merge it so it doesn't linger and get forgotten.

:: dave | 251.463.2345 c

gazebo_ros_control/src/default_robot_hw_sim.cpp
jim-rothrock: It's now been removed.

jim-rothrock:
> how extensively have the contents of this PR request been exercised in practice?

I use it daily with position-position and velocity-velocity joints, with and without PID control. I've done significant testing with position-effort and velocity-effort joints. I've done moderate testing with other permutations.

> maybe a bit too much conditionals (if and switch) for my taste

It would be cleaner to create classes for Joint, EffortJoint, PositionJoint, etc., and use virtual functions instead of switch statements, as well as replacing the separate joint_names_, joint_types_, etc. vectors with a single vector of Joint pointers, but I decided to follow the pattern that was established in the existing code. If someone decides to add support for various transmission types, it probably will be time for a major code reorganization, or a different class derived from gazebo_ros_control::RobotHWSim.

adolfo-rt:
Ha!, I just realized I'm not a collaborator in this repo, so I can't merge the PR.

PR #138, opened 2013-10-30
initialize count lock in kinect plugin

jonbinney:
Otherwise the gazebo_ros_openni_kinect plugin crashes. Should this actually be initialized in the GazeboRosCameraUtils class? It looks like other subclasses do it themselves, but not all of them.


hsu:
if I am not mistaken, it looks like all except `depth_image_connect_count_` are initialized by [this bit of initialization](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_plugins/src/gazebo_ros_camera_utils.cpp#L237)?

jonbinney:
Interesting - i'll try to reproduce the error next week and get a gdb backtrace to see exactly which one is causing the problem

jonbinney:
Closing this for now, since I don't see this problem anymore.

PR #144, opened 2013-11-05
Fixed #125: gazebo_ros_control: controlPeriod greater than the simulation period causes unexpected results

### LINKED ISSUE #125: gazebo_ros_control: controlPeriod greater than the simulation period causes unexpected results.
I am running my Gazebo simulation at 1 kHz. This seems like overkill for my effort-based joint position and velocity controllers (servos often use a 50 Hz PID loop), so I set the gazebo_ros_control control frequency to 100 Hz, like this:

<pre>
  &lt;xacro:property name="ctrl_period" value="0.01"/&gt;

  &lt;gazebo&gt;
    &lt;plugin name="gazebo_ros_control" filename="libgazebo_ros_control.so"&gt;
      &lt;controlPeriod&gt;${ctrl_period}&lt;/controlPeriod&gt;
    &lt;/plugin&gt;
  &lt;/gazebo&gt;
</pre>

This caused my simulated vehicle's shock absorbers to stop working, and its steering to become sluggish. I mitigated (but didn't completely fix) the problem by making this modification to gazebo_ros_control_plugin.cpp:

<pre>
void GazeboRosControlPlugin::Update()
{
  // Get the simulation time and period
  gazebo::common::Time gz_time_now = parent_model_->GetWorld()->GetSimTime();
  ros::Time sim_time_ros(gz_time_now.sec, gz_time_now.nsec);
  ros::Duration sim_period = sim_time_ros - last_sim_time_ros_;

#ifdef NOTDEF
  // Check if we should update the controllers
  if(sim_period >= control_period_) {
    // Store this simulation time
    last_sim_time_ros_ = sim_time_ros;

    // Update the robot simulation with the state of the gazebo model
    robot_hw_sim_->readSim(sim_time_ros, sim_period);

    // Compute the controller commands
    controller_manager_->update(sim_time_ros, sim_period);

    // Update the gazebo model with the result of the controller
    // computation
    robot_hw_sim_->writeSim(sim_time_ros, sim_period);
  }
#else
  robot_hw_sim_->readSim(sim_time_ros, ros::Duration(0.001));

  if (sim_period >= control_period_)
  {
    last_sim_time_ros_ = sim_time_ros;
    controller_manager_->update(sim_time_ros, sim_period);
  }

  robot_hw_sim_->writeSim(sim_time_ros, ros::Duration(0.001));
#endif
}
</pre>

It seems as if the original code and the replacement I wrote should be roughly equivalent. My hypothesis is that the SetForce() calls in DefaultRobotHWSim::writeSim() affect Gazebo's current time step only. However, I've found no SetForce() documentation that describes that behavior.

For the moment, I'm going to refrain from setting controlPeriod in my URDF file, so readSim(), update, and writeSim() will be called during each Gazebo time step. I'm posting this issue in case someone wants to investigate it.

meyerj:
This is a fix for #125.

I also added a Reset() method, which is required to prevent that controllers are called with negative update periods when the world is reset and sim time starts at zero again.

### DIFF
--- a/gazebo_ros_control/include/gazebo_ros_control/gazebo_ros_control_plugin.h
+++ b/gazebo_ros_control/include/gazebo_ros_control/gazebo_ros_control_plugin.h
@@ -72,6 +72,9 @@ public:
   // Called by the world update start event
   void Update();
 
+  // Called on world reset
+  virtual void Reset();
+
   // Get the URDF XML from the parameter server
   std::string getURDF(std::string param_name) const;
 
--- a/gazebo_ros_control/include/gazebo_ros_control/gazebo_ros_control_plugin.h
+++ b/gazebo_ros_control/include/gazebo_ros_control/gazebo_ros_control_plugin.h
@@ -114,7 +114,8 @@ protected:
 
   // Timing
   ros::Duration control_period_;
-  ros::Time last_sim_time_ros_;
+  ros::Time last_update_sim_time_ros_;
+  ros::Time last_write_sim_time_ros_;
 
 };
 
--- a/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp
+++ b/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp
@@ -196,23 +196,24 @@ void GazeboRosControlPlugin::Update()
   // Get the simulation time and period
   gazebo::common::Time gz_time_now = parent_model_->GetWorld()->GetSimTime();
   ros::Time sim_time_ros(gz_time_now.sec, gz_time_now.nsec);
-  ros::Duration sim_period = sim_time_ros - last_sim_time_ros_;
+  ros::Duration sim_period = sim_time_ros - last_update_sim_time_ros_;
 
   // Check if we should update the controllers
   if(sim_period >= control_period_) {
     // Store this simulation time
-    last_sim_time_ros_ = sim_time_ros;
+    last_update_sim_time_ros_ = sim_time_ros;
 
     // Update the robot simulation with the state of the gazebo model
     robot_hw_sim_->readSim(sim_time_ros, sim_period);
 
     // Compute the controller commands
     controller_manager_->update(sim_time_ros, sim_period);
-
-    // Update the gazebo model with the result of the controller
-    // computation
-    robot_hw_sim_->writeSim(sim_time_ros, sim_period);
   }
+
+  // Update the gazebo model with the result of the controller
+  // computation
+  robot_hw_sim_->writeSim(sim_time_ros, sim_time_ros - last_write_sim_time_ros_);
+  last_write_sim_time_ros_ = sim_time_ros;
 }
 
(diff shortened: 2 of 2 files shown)


hsu:
Thanks @meyerj , do you have a simple test I can test this pr against?

hsu:
thanks for adding the test.  But while testing `spawn_can.launch`, the can accelerates upwards for me initially.  I modified:

```
diff --git a/gazebo_ros_control_test/src/anti_gravity_controller.cpp b/gazebo_ros_control_test/src/anti_gravity_controller.cpp
index 61c394c..c792a9d 100644
--- a/gazebo_ros_control_test/src/anti_gravity_controller.cpp
+++ b/gazebo_ros_control_test/src/anti_gravity_controller.cpp
@@ -45,7 +45,7 @@ public:

   virtual void update(const ros::Time &time, const ros::Duration &period)
   {
-    gazebo::math::Vector3 gravity(0.0, 0.0, 9.80);
+    gazebo::math::Vector3 gravity(0.0, 0.0, 9.81*0.39);
     hw_->setWrench(gravity);
   }
```

and it works now.

meyerj:
I was aware that the test did not work yet and wanted to fix it today. Thanks @hsu for finding the bug.
The use case is similar to what I added to the [ros_control branch](https://github.com/tu-darmstadt-ros-pkg/hector_quadrotor/tree/ros_control) of [hector_quadrotor](http://wiki.ros.org/hector_quadrotor) to control a quadrotor.

I also noticed that the white bounding box around the can is shifted about half a meter when the object is selected:


This is not the case for the original coke can in the Gazebo model database. Is this related to #148?

hsu:
@meyerj I don't see the shift in bounding box when I load and select your model in 33eac071c694c887c4d9d7d48480d5b2e831f990.  I realized that #148 was not a complete fix for model spawn pose, can you try with the rest of model spawn pose fix in #150?

meyerj:
@hsu: I already used the latest revision 635276623e77fd5d395aa274410b65c7a2f13da1 of gazebo_ros_pkgs for the screenshot, where #150 has been merged. The bounding box error is still there. Perhaps it's because the model is sdf and not urdf?

The can is spawned at z=1.0 when you my spawn_can.launch file. If I change this value to **z=0**, the bounding box is still slightly shifted.


I am not sure, but this might be a timing issue? If the simulation is paused, the bounding box seems to be displayed correctly in the z=0 case. The cans that are spawned directly from the Gazebo UI also don't have the shift. The hovering_can sdf model was just copied from the model database, because I haven't found a way to include it and add a plugin.

I am using Gazebo version 1.9.1 in ROS hydro on Ubuntu precise installed from the standard ROS package repositories.

Please don't waste time just because of the bounding box. The visual and collision model seems to be correct in all these cases. Another minor bug I discovered during testing this is that models that are deleted and respawned with the same name are not visible in the 3D view until it is restarted. They only show up in the model list.

PR #148, opened 2013-11-09
fix spawn initial pose.  When model has a non-zero initial pose and user...

hsu:
... specified initial model spawn pose, add the two.


hsu:
tested by

```
roslaunch gazebo_ros shapes_world.launch 
```

followed by

```
rosservice call /gazebo/delete_model "model_name: 'sphere'" 
```

Issue #155, opened 2013-12-16
gazebo_plugins gazebo_ros_bumper no contact seen


ugocupcic:
When adding a gazebo_ros_bumper sensor to my urdf, I'm not able to get a contact from the published ROS topic (states stays empty, `contact_size() = 0` on line 206), even though the contact seems to be detected by Gazebo, as shown in the attached screenshot.

![error_bumper](https://f.cloud.github.com/assets/557182/1755789/6eea74d6-666a-11e3-8256-5fbf5ff27e4a.png)

I'm running ROS hydro on ubuntu 12.04. I tried both the default package installation, and a version installed from sources (both gazebo and gazebo plugins). You can find the urdf used here: https://gist.github.com/ugocupcic/7989409

Any help is greatly appreciated!

ugocupcic:
After more digging, it appears that the issue comes from a different naming convention between gazebo and the ros plugin (?): the collision element we try to access in Gazebo is named `__default__`, but when imported from urdf, the name is `linkname_collision` by default. 

I'm reporting this on Gazebo's bug tracker: https://bitbucket.org/osrf/gazebo/pull-request/882/fixing-the-contactsensor-in-gazebo/diff

scpeters:
I saw your gazebo pull request; sorry I didn't look at your issue sooner.

I your issue can be resolved by adding some additional tags inside your `<gazebo><sensor>`. Here's an [example from atlas_description in drcsim](https://bitbucket.org/osrf/drcsim/src/default/atlas_description/urdf/atlas_v3.gazebo#cl-227):

```
  <gazebo reference="r_foot">
    <!-- contact sensor -->
    <sensor type="contact" name="r_foot_contact_sensor">
      <update_rate>1000.0</update_rate>
      <always_on>1</always_on>
      <contact>
        <collision>r_foot_collision</collision>
        <topic>/r_foot_contact</topic>
      </contact>
    </sensor>
  </gazebo>
```

You'll want the `<always_on>` tag, and you'll also want to include the names of each collision element under `<contact><collision>`.

ugocupcic:
Thanks for that clean solution! I spent ages trying to figure out what went wrong with those bumpers again....

I'd really suggest updating at least that link:
 http://gazebosim.org/wiki/Tutorials/1.9/ROS_Motor_and_Sensor_Plugins#Bumper

I found it very hard to find proper up to date information on that front (so much so that I opened those pull requests!). Thanks for the help.

[...]

PR #156, opened 2013-12-18
[gazebo_plugins] Fix gazebo plugins bumper

ugocupcic:
In addition to the fix that needs to be done on gazebo (https://bitbucket.org/osrf/gazebo/pull-request/882/fixing-the-contactsensor-in-gazebo), the parent sensor (gazebo `ContactSensor`) of the `GazeboRosBumper` has to be activated, otherwise we're not getting any data.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_bumper.cpp
+++ b/gazebo_plugins/src/gazebo_ros_bumper.cpp
@@ -208,7 +208,7 @@ void GazeboRosBumper::OnContact()
     // For each collision contact
     // Create a ContactState
     gazebo_msgs::ContactState state;
-    /// \TODO:
+    /// \TODO: 
     gazebo::msgs::Contact contact = contacts.contact(i);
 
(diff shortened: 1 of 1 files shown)


scpeters:
I just commented in #155 about this, but I'll mention it here too. You can make the contact sensor active through an `<always_on>` tag in the `<gazebo><sensor>` hierarchy (see [this example from atlas_description in drcsim](https://bitbucket.org/osrf/drcsim/src/default/atlas_description/urdf/atlas_v3.gazebo#cl-227)).

That said, I'm not opposed to this pull request; I don't know why the contact sensors are off by default in gazebo. I'm just pointing out that there's a workaround.

ugocupcic:
If we don't set `<always_on>` in the xml, the sensor is not publishing the data (without that patch) unless we visualise the topic in gazebo. 

The problem with that pull request is that it overrides the always_on behaviour: it always publishes the data.

PR #157, opened 2013-12-24
Very small fix in gazebo_ros_api_plugin

po1:
While harmless, a self-assignment clutters the already big source file and has been removed. It has been reported by cppcheck:

```
[gazebo_ros_pkgs/gazebo_ros/src/gazebo_ros_api_plugin.cpp:1496]: warning: Redundant assignment of 'target_force' to itself. [cppcheck: selfAssignment]
```

If this assignment was having an intended side-effect, I strongly suggest that this side effect be reworked into a more explicit form.

Also, my editor (vim) could not resist the urge of fixing a few newline characters (or trailing white spaces), so I committed them separately.


hsu:
looks good to me.

PR #165, opened 2014-02-12
[gazebo_plugins] made depth camera and openni kinect plugins behave according to REP 117

stwirth:
[REP-117](http://www.ros.org/reps/rep-0117.html) defines how distance measurements should be represented in ROS. The depth images coming from gazebo do not follow this convention.
In the plugin, all values below a threshold parameter are set to `std::numeric_limits<float>::quiet_NaN()`. According to the REP all "too close" measurements should return -inf instead.
Gazebo returns 0.0 as depth value for pixels that do not contain any object. Those are translated into NaNs as well but should be set to +inf instead.

This PR changes the behavior of the depth camera and the openni kinect plugin to match the REP.

What I don't like much is the check for `depth == 0.0f` to get the pixels that shoud have +inf. Gazebo itself should return +inf for pixels that have no object behind them.
Maybe the fragment shader that is used [here](https://bitbucket.org/osrf/gazebo/src/0c7d0c7b9a221aa512aaa6ec88c5e9104b62ca2e/gazebo/rendering/DepthCamera.cc?at=default#cl-95) should be adapted?

@chadrockey Could you check the REP conformity please?

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
@@ -394,13 +394,19 @@ bool GazeboRosDepthCamera::FillPointCloudHelper(
       pcl::PointXYZRGB point;
       point.x      = depth * tan(yAngle);
       point.y      = depth * tan(pAngle);
-      if(depth > this->point_cloud_cutoff_)
+      if (depth == 0.0f) // gazebo returns 0.0 when no object was rendered 
+      {
+        // set "no returns" to +infinity according to REP-117
+        point.z = std::numeric_limits<float>::infinity();
+      }
+      else if(depth > this->point_cloud_cutoff_)
       {
         point.z    = depth;
       }
-      else //point in the unseeable range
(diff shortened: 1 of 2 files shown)


chadrockey:
@stwirth Unfortunately, with the way the actual hardware works (for Primesense based devices) there's no guarantee which value the measurements actually represents.  Given the deadband, Primesense devices fail to correlate the values and return a value of 0 in this area.  (REP states -Inf for this case).  However, this is the same value as 0/I did not receive data (REP states +Inf).  Finally, it's also the same case for a non-reflective black surface (REP states NaN).  Since there's no context aware way to separate the three cases outside of simulation, Gazebo did the most accurate thing for a Primesense device by returning all NaNs.

Some phase of flight cameras can return -Inf.  In this case, REP 117 can be fully implemented and the user chooses what to do with the various values.

I would say make depth camera conform to the REP but allow openni kinect to follow the primesense standard.

jonbinney:
I think this can probably be closed since it matches the existing behavior matches the hardware.

PR #185, opened 2014-04-16
[gazebo_ros_control] Indigo compatibility

adolfo-rt:
These changesets require ros-controls/ros_control#156.

### DIFF
--- a/gazebo_ros_control/package.xml
+++ b/gazebo_ros_control/package.xml
@@ -18,7 +18,6 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <build_depend>roscpp</build_depend> 
-  <build_depend>gazebo_ros</build_depend> 
   <build_depend>control_toolbox</build_depend> 
   <build_depend>controller_manager</build_depend> 
   <build_depend>pluginlib</build_depend> 
--- a/gazebo_ros_control/CMakeLists.txt
+++ b/gazebo_ros_control/CMakeLists.txt
@@ -4,7 +4,6 @@ project(gazebo_ros_control)
 # Load catkin and all dependencies required for this package
 find_package(catkin REQUIRED COMPONENTS
   roscpp
-  gazebo_ros
   control_toolbox
   controller_manager
   hardware_interface 
@@ -20,7 +19,6 @@ find_package(gazebo REQUIRED)
 catkin_package(
   CATKIN_DEPENDS
     roscpp
-    gazebo_ros
     controller_manager
     pluginlib
     transmission_interface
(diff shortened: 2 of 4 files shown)


davetcoleman:
+1 reviewed.

adolfo-rt:
Please don't merge this PR yet.

@davetcoleman You might be interested in testing these changes once ros-controls/ros_control#156 is merged. I have only checked that they build so far. I might have time to do a bit more next week. I'm not a user of the default plugin, and there's no test suite to validate against, so actual user validation would be valuable.

I tried to adapt the `DefaultRobotHWSim` implementation to support transmissions with multiple joints and multiple hardware interfaces, but realized that the class needs an important refactoring and didn't continue. I opened #186 to keep track of this issue, which is important in its own right.

gazebo_ros_control/package.xml
```diff
@@ -27,7 +26,6 @@
   <build_depend>urdf</build_depend>
   
   <run_depend>roscpp</run_depend> 
-  <run_depend>gazebo_ros</run_depend>
```
scpeters, line 30: I think it still needs the `run_depend` on gazebo_ros (see [gazebo_ros_control_plugin.cpp:74-80](https://github.com/pal-robotics/gazebo_ros_pkgs/blob/gazebo-ros-control-indigo/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp#L74-L80)).

davetcoleman, line 30: Agreed.

adolfo-rt:
PR modified to preserve the runtime dependency on `gazebo_ros`.

adolfo-rt:
The pending ros_control for Indigo PR has been merged. Yet ros_control and friends have not yet been released into Indigo, only `control_msgs` is. Let's get that done and then merge this one.

davetcoleman:
I created the branches for it yesterday, but someone else needs to test it on an indigo machine.

scpeters:
It builds for me on saucy/indigo, but I haven't run any code. Any recommendations for testing?

adolfo-rt:
I would say that this tutorial is the best there is now, would you agree @davetcoleman?

http://gazebosim.org/wiki/Tutorials/1.9/ROS_Control_with_Gazebo

davetcoleman:
Agreed, not the best documentation on non-simulation

sent from my phone

dave :: 251.463.2345

scpeters:
Sorry I haven't managed to do any run-time testing. Since it builds for me, I would propose merging this so that we can make releases and then resolve any issues as they are reported.

+1

adolfo-rt:
OK, I'm merging. Building from source should work, but take into account that the ros-controls repos have not yet been released into the buildfarm for indigo. It's on my todo list. As of today, all critical PRs on the ros-controls organization for indigo are either merged or submitted.

scpeters:
Thanks @adolfo-rt !

PR #188, opened 2014-05-08
added osx support for gazebo start scripts

markusachtelik:


scpeters:
Makes sense to me. @wjwwood what do you think?

wjwwood:
That looks ok.

PR #190, opened 2014-05-15
Add verbose parameter

clynamen:
Add verbose parameter for --verbose gazebo flag


scpeters:
Looks good to me. What is the policy now about hydro vs. indigo? Do we target hydro and merge forward or vice versa?

clynamen:
I don't know actually, I did not find anything about the branching model on the ros wiki. 
However since this is a new feature and not a bugfix I guess this could be merged directly in indigo.

hsu:
+1

PR #199, opened 2014-05-30
change equality operator in rosrun scripts to remove error message [: Linux: unexpected operator

ghost:


jonbinney:
This gets rid of the error on my machine as well, but I don't understand why. All bash documentation I can find says that "==" is valid. Do you know if it has been removed in recent versions of bash?

ghost:
I found this on stackoverflow. To make it a bash script, the shebang should be changed to `!/bin/bash`. With the current shebang, it is executed as a shell script, and the syntax is slightly different.

jonbinney:
Ah, I see. The `gdbrun` script does have `/bin/bash` in the shebang, so i had assumed that the others did too. Makes sense now, and does look like posix scripts should use a single equals sign.

+1 for merging this

davetcoleman:
Or should we just change the script to a bash script? Seems like the better solution @hsu ?

hsu:
If we don't need to additional features offered by `bash`, I'd say we go with `sh`

jbohren:
+1 to merge this PR for POSIX compliance

PR #203, opened 2014-05-31
Fix repo names in package.xmls (groovy-devel branch)

jonbinney:
Addresses https://github.com/ros-simulation/gazebo_ros_pkgs/issues/200 for the groovy-devel branch


scpeters:
+1, thanks!

PR #205, opened 2014-06-09
gazebo_plugins: Added updateRate parameter to the gazebo_ros_imu plugin

fsuarez6:
This PR adds the `<updateRate>` parameter to the `gazebo_ros_imu` plugin. I added it because at the moment the plugin runs as fast as possible.

Usage example:

``` xml
<gazebo>
  <plugin name="imu_sensor" filename="libgazebo_ros_imu.so">
    <updateRate>100.0</updateRate>
    <topicName>imu_sensor</topicName>
    <bodyName>IMU_LINK</bodyName>
  </plugin>
</gazebo>
```

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_imu.cpp
+++ b/gazebo_plugins/src/gazebo_ros_imu.cpp
@@ -114,6 +114,15 @@ void GazeboRosIMU::LoadThread()
   }
   else
     this->offset_.rot = this->sdf->Get<math::Vector3>("rpyOffset");
+  
+  if (!this->sdf->HasElement("updateRate"))
+  {
+    ROS_DEBUG("imu plugin missing <updateRate>, defaults to 0.0"
+             " (as fast as possible)");
+    this->update_rate_ = 0;
+  }
+  else
+    this->update_rate_ = this->sdf->GetElement("updateRate")->Get<double>();
 
 
   // Make sure the ROS node for Gazebo has already been initialized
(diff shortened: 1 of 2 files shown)


gazebo_plugins/src/gazebo_ros_imu.cpp
```diff
@@ -114,6 +114,15 @@ void GazeboRosIMU::LoadThread()
   }
   else
     this->offset_.rot = this->sdf->Get<math::Vector3>("rpyOffset");
+  
+  if (!this->sdf->HasElement("updateRate"))
+  {
+    ROS_DEBUG("imu plugin missing <updateRate>, defaults to 0.0"
+             " (as fast as possible)");
+    this->update_rate_ = 0;
```
jonbinney: Small thing - using "0.0" instead of "0" will make it clearer to someone reading the code that this is a float.

gazebo_plugins/src/gazebo_ros_imu.cpp
```diff
@@ -187,6 +196,13 @@ bool GazeboRosIMU::ServiceCallback(std_srvs::Empty::Request &req,
 // Update the controller
 void GazeboRosIMU::UpdateChild()
 {
+  common::Time cur_time = this->world_->GetSimTime();
+  
+  // rate control
+  if (this->update_rate_ > 0 &&
+      (cur_time-this->last_time_).Double() < (1.0/this->update_rate_))
```
jonbinney: Style comment - need a couple other spaces "(1.0 / this->update_rate_)"

jonbinney:
The code looks good to me, but I haven't tested it.

Issue #210, opened 2014-06-11
Point cloud from libgazebo_ros_openni_kinect.so is not organized


mmurooka:
The width and height fields of point cloud from libgazebo_ros_openni_kinect.so is not organized as follow:

``` bash
$ rostopic echo -n 1 /xtion/depth/points/width
 307200
 ---
$ rostopic echo -n 1 /xtion/depth/points/height
 1
 ---
```

Height should be 480 and width should be 640.
My sensor definition is as bellow:

``` xml
  <gazebo reference="xtion_depth_frame">
    <sensor name="xtion" type="depth">
      <update_rate>20</update_rate>
      <camera>
        <horizontal_fov>1.09606677025</horizontal_fov>
        <image>
          <format>R8G8B8</format>
          <width>640</width>
          <height>480</height>
        </image>
        <clip>
          <near>0.5</near>
          <far>9</far>
        </clip>
      </camera>
      <plugin filename="libgazebo_ros_openni_kinect.so" name="xtion_camera_controller">
        <alwaysOn>true</alwaysOn>
        <updateRate>20</updateRate>
        <imageTopicName>xtion/rgb/image_raw</imageTopicName>
        <cameraInfoTopicName>xtion/rgb/camera_info</cameraInfoTopicName>
        <depthImageTopicName>xtion/depth/image_raw</depthImageTopicName>
        <depthImageCameraInfoTopicName>xtion/depth/camera_info</depthImageCameraInfoTopicName>
        <pointCloudTopicName>xtion/depth/points</pointCloudTopicName>
        <frameName>xtion_depth_optical_frame</frameName>
        <distortion_k1>0.0</distortion_k1>
        <distortion_k2>0.0</distortion_k2>
        <distortion_k3>0.0</distortion_k3>
        <distortion_t1>0.0</distortion_t1>
        <distortion_t2>0.0</distortion_t2>
      </plugin>
    </sensor>
  </gazebo>
```

garaemon:
[this pull](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/211) request should solve the issue

mmurooka:
pull request was merged.

PR #211, opened 2014-06-13
publish organized pointcloud from openni plugin

garaemon:
publish pointcloud from openni as organized pointcloud.

see https://github.com/ros-simulation/gazebo_ros_pkgs/issues/210


jonbinney:
This seems like it could break code that expects a 1d array of points - maybe it shouldn't go into hydro?

garaemon:
> This seems like it could break code that expects a 1d array of points

I think the point cloud keeps 1d array even though you apply this patch.

jonbinney:
I see what you mean - the actual array in the message will be 1d. Not sure if anyone has code which expects width and height to not be set, but it seems reasonable to call this a bugfix rather than a new feature.

garaemon:
> Not sure if anyone has code which expects width and height to not be set

I think there may be several codes expect organized pointcloud structure and I don't think there is a code which expects `height=1`

Issue #213, opened 2014-06-30
Diff Drive plugin forces tf odom frame.


curranw:
For situations where the user needs to create their own simulated odom, like from laser scans, it conflicts with the diff drive odom tf frame published by gazebo.

http://answers.ros.org/question/175013/removingdeletingignoring-odom-tf-data/

jonbinney:
You could just use a launchfile to remap the odom that gazebo publishes to some other topic like odom_diff_drive, right?

jonbinney:
Actually there is a param that you can put in the sdf/urdf plugin called "odometryTopic" which you can just set to anything besides "odom"

curranw:
The issue is the TF, not the odom publisher. There is no way to kill a tf frame, and the odom tf connects to base_footprint automatically using the plugin. If you rename it, the tf frame is still there.

jonbinney:
Ah, I see. Makes sense.

mikepurvis:
Not sure if you're referring to `diff_drive_controller`, but it now supports disabling publishing the odom tf: https://github.com/ros-controls/ros_controllers/issues/118

j-rivero:
Looking at Mike's comment looks like we can close the issue. Please reopen if you think that the problem is still present.

PR #214, opened 2014-06-30
Issue #213

curranw:
Added an XML argument to the Diff Drive plugin to allow the turning on/off of tf/topic publishing for odometry.


hsu:
This changes the default behavior to turn off publishing tf without `<publishTF>`.  Is this expected?

curranw:
No that was a mistake. Default behavior switched. Now the change shouldn't affect anyone unless explicitly changed.

hsu:
Thanks for the update. Looks good.

hsu:
merging in #230

PR #221, opened 2014-07-31
Fix build for gazebo4

nkoenig:


adolfo-rt:
Shouldn't this PR target `indigo-devel`?. The fix for `gazebo_ros_control_plugin.cpp` is already there.

It has already been suggested before, but we should change the repository's default branch to `indigo-devel`, especially now that indigo has been released.

j-rivero:
I've tested the pull request under Trusty for:
- gazebo4-prerelease + Indigo
- gazebo3 + Indigo
- gazebo2 + Indigo
- gazebo + hydro

It is compiling just fine for me on all the combinations.

@adolfo-rt we are still producing ROS wrappers for gazebo new versions (gazebo3, gazebo4) in Hydro and it would help to have the fixes applied to both branches, hydro-devel and indigo-devel.

Thanks.

hsu:
`GAZEBO_VERSION_MAJOR` should be `GAZEBO_MAJOR_VERSION`, similarly for others.

gazebo_ros_control/src/default_robot_hw_sim.cpp
```diff
@@ -288,7 +294,11 @@ class DefaultRobotHWSim : public gazebo_ros_control::RobotHWSim
           break;
 
         case POSITION:
+#if GAZEBO_VERSION_MAJOR >= 4
```
hsu: should be `GAZEBO_MAJOR_VERSION`

adolfo-rt:
> @adolfo-rt we are still producing ROS wrappers for gazebo new versions (gazebo3, gazebo4) in Hydro and it would help to have the fixes applied to both branches, hydro-devel and indigo-devel.

OK, I was not aware of this.

PR #227, opened 2014-08-28
check physics engine type before calling set_physics_properties and get_...

hsu:
...physics_properteis


caguero:
@hsu , should we target this pull request to indigo-devel too?

caguero:
+1

hsu:
done for indigo-devel in #232

PR #228, opened 2014-08-28
Fix get physics properties non ode

hsu:


j-rivero:
Given the list of commits should we point this PR to indigo-devel instead of hydro-devel?

j-rivero:
If I use this branch, the problem when using simbody is gone. Thanks.

hsu:
@j-rivero you're right, should go to indigo-devel, thanks.

hsu:
reopened with indigo-devel as #232

PR #238, opened 2014-09-05
Fixing handling of non-world frame velocities in setModelState.

ayrton04:
For issue #240


gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1287,7 +1287,7 @@ bool GazeboRosApiPlugin::setModelState(gazebo_msgs::SetModelState::Request &req,
 
       //std::cout << " debug : " << relative_entity->GetName() << " : " << frame_pose << " : " << target_pose << std::endl;
       //target_pose = frame_pose + target_pose; // seems buggy, use my own
-      target_pose.pos = frame_pos + frame_rot.RotateVector(target_pos);
+      target_pose.pos = model->GetWorldPose().pos + frame_rot.RotateVector(target_pos);
```
cheffe112, line 1290: This seems to break using ```reference_frame``` for me, the model starts drifting away, see #638.

PR #239, opened 2014-09-05
Fixing handling of non-world frame velocities in setModelState.

ayrton04:
For issue #240

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -1284,6 +1284,11 @@ bool GazeboRosApiPlugin::setModelState(gazebo_msgs::SetModelState::Request &req,
       //target_pose = frame_pose + target_pose; // seems buggy, use my own
       target_pose.pos = frame_pos + frame_rot.RotateVector(target_pos);
       target_pose.rot = frame_rot * target_pose.rot;
+     
+      // Velocities should be commanded in the requested reference
+      // frame, so we need to translate them to the world frame
+      target_pos_dot = frame_rot.RotateVector(target_pos_dot);
+      target_rot_dot = frame_rot.RotateVector(target_rot_dot);
     }
     /// @todo: FIXME map is really wrong, need to use tf here somehow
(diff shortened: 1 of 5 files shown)


hsu:
Tested with pioneer2dx model per issue #240 , looks good to me.
It'll be great if you can also add a launch file for pioneer 2dx with a test ros nodes that demonstrates the problem under a directory in `gazebo_ros/test/`? Thanks!

ayrton04:
I'll try to throw something together soon, sure.

hsu:
thanks @ayrton04.

hsu:
+1

since there isn't a test yet, one more approval before I merge this.

ayrton04:
Sorry, I've been buried lately. I'll get to it soon, I promise!

EDIT: Did you want the test in gazebo_ros/test or gazebo_plugins/test?

hsu:
please put it in gazebo_plugins/test, thanks.

ayrton04:
Please don't merge this or the indigo-devel PR yet. I may have discovered a different issue.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1282,8 +1282,13 @@ bool GazeboRosApiPlugin::setModelState(gazebo_msgs::SetModelState::Request &req,
 
       //std::cout << " debug : " << relative_entity->GetName() << " : " << frame_pose << " : " << target_pose << std::endl;
       //target_pose = frame_pose + target_pose; // seems buggy, use my own
-      target_pose.pos = frame_pos + frame_rot.RotateVector(target_pos);
+      target_pose.pos = model->GetWorldPose().pos + frame_rot.RotateVector(target_pos);
```
ayrton04, line 1285: Part of the problem is that frame_pos is the pose of the relative_entity (e.g., the "chassis" link in the Pioneer model). This means that when we call model->SetWorldPose(target_pose) on line 1310, we'll setting the model's position using the position of the relative_entity. If relative_entity has a non-zero offset from the model origin, then we'll be repeatedly moving the model by the amount of the offset. In the case of the pioneer, the model has a Z offset of 0.16 meters, so every time this function is called, the robot climbs 16 centimeters into the air. If you call it with great enough frequency, the robot flies.

ayrton04:
Should be fixed. I added the new test as a ROS test in the CMakelists file.

PR #242, opened 2014-09-12
Specify physics engine in args to empty_world.launch

scpeters:
As of gazebo3, there is support for multiple physics engines in gazebo using the `-e` parameter. This pull request exposes that parameter to the empty_world launch file. The following should load the empty world in bullet:

```
roslaunch gazebo_ros empty_world.launch physics:=bullet verbose:=true
```


hsu:
+1

should we also do this for hydro-devel?  given we are still releasing hydro + gazebo 3+ for drcsim.

scpeters:
See #254 for hydro

PR #259, opened 2014-10-06
Header for GetModelState service request

maxbader:
Hi

I added a general header to the GetModelState.srv and some lines to the gazebo_ros_api_plugin.cpp which to get a actural time stamp to a model request.
I did this to fix the model state to the simlated time which is needed in many cases
I think my work would also be downward compatibel.

Greetings
Markus

PS.: I would suggest to do the same thing for the ModelStates.msg or ModelState.msg

### DIFF
--- a/gazebo_msgs/srv/GetModelState.srv
+++ b/gazebo_msgs/srv/GetModelState.srv
@@ -4,6 +4,10 @@ string relative_entity_name          # return pose and twist relative to this en
                                      # be sure to use gazebo scoped naming notation (e.g. [model_name::body_name])
                                      # leave empty or "world" will use inertial world frame
 ---
+Header header                        # Standard metadata for higher-level stamped data types.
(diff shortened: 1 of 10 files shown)


maxbader:
Anybody interested?

adolfo-rt:
Take into account that modifying non-comment parts of a message specification will change its MD5 sum. Even if the changes you propose are relevant, they would have to be applied to the next ROS distro (jade-turtle) for message compatibility reasons.

hsu:
created jade-devel branch, please retarget. thanks!

hsu:
Hi @maxbader can you recreate the pull request to target jade-devel?
thanks.

maxbader:
Hi @hsu I created an other pull request for jade-devel. have a look at the pull request nr #302.
If 302 you can reject this one or I will close it.

scpeters:
#302 merged instead

PR #261, opened 2014-10-09
add launch files for spawning models at gazebo world

YoheiKakiuchi:
Add launch files for spawning models, such as box, cylinder, model in database.


hsu:
What is the specific use case you need these launch files for?
thanks.

furushchev:
+1

furushchev:
I think this launch file is for spawning many kinds of models from gazebo database by changing only launch file.
This is very useful when writing test codes that needs specific scenario.

scpeters:
The box and cylinder launch files do not have inertia values specified. Can you add reasonable values?

ahcorde:
Very old PR from 2014, targeting a not supported ROS version, closing this

PR #263, opened 2014-10-17
spawn_model: adding -b option to bond to the model and delete it on sigint

jbohren:


jbohren:
As an aside, the indentation in this python file is really messed up.

jbohren:
Superseded by #264

PR #273, opened 2014-11-21
GetModelState.srv with a Header (Timestamp)

maxbader:
Hi

I anhanced the GetModelState service with header information because it is important for the client to have a timestamp with the model.

Greetings
Markus

### DIFF
--- a/gazebo_msgs/srv/GetModelState.srv
+++ b/gazebo_msgs/srv/GetModelState.srv
@@ -4,6 +4,10 @@ string relative_entity_name          # return pose and twist relative to this en
                                      # be sure to use gazebo scoped naming notation (e.g. [model_name::body_name])
                                      # leave empty or "world" will use inertial world frame
 ---
+Header header                        # Standard metadata for higher-level stamped data types.
+                                     # * header.seq holds the number of requests since the plugin started
+                                     # * header.stamp timestamp related to the pose
+                                     # * header.frame_id not used but currently filled with the relative_entity_name
 geometry_msgs/Pose pose              # pose of model in relative entity frame
 geometry_msgs/Twist twist            # twist of model in relative entity frame
 bool success                         # return true if get successful
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -776,6 +776,24 @@ bool GazeboRosApiPlugin::getModelState(gazebo_msgs::GetModelState::Request &req,
   }
   else
   {
+    /**
+     * @brief creates a header for the result
+     * @author Markus Bader markus.bader@tuwien.ac.at
+     * @date 21th Nov 2014
+     **/
+    {
+      std::map<std::string, unsigned int>::iterator it = access_count_get_model_state_.find(req.model_name);
+      if(it == access_count_get_model_state_.end()) {
+	access_count_get_model_state_.insert( std::pair<std::string, unsigned int>(req.model_name, 1) );
+	res.header.seq = 1;
+	
+      } else {
+	it->second++;
+	res.header.seq = it->second;
+      }
+      res.header.stamp = ros::Time::now();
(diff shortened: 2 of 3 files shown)


gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -776,6 +776,24 @@ bool GazeboRosApiPlugin::getModelState(gazebo_msgs::GetModelState::Request &req,
...
+     * @author Markus Bader markus.bader@tuwien.ac.at
+     * @date 21th Nov 2014
+     **/
+    {
+      std::map<std::string, unsigned int>::iterator it = access_count_get_model_state_.find(req.model_name);
+      if(it == access_count_get_model_state_.end()) {
+	access_count_get_model_state_.insert( std::pair<std::string, unsigned int>(req.model_name, 1) );
+	res.header.seq = 1;
+	
+      } else {
```
ffurrer: stick to the same `if` `else` style as in the rest of the file.

```
if (condition) // With the space after the if
{
  ...
}
else
{
  ...
}
```

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -776,6 +776,24 @@ bool GazeboRosApiPlugin::getModelState(gazebo_msgs::GetModelState::Request &req,
...
+	it->second++;
```
ffurrer: pre-increment

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -776,6 +776,24 @@ bool GazeboRosApiPlugin::getModelState(gazebo_msgs::GetModelState::Request &req,
...
   {
+    /**
+     * @brief creates a header for the result
+     * @author Markus Bader markus.bader@tuwien.ac.at
+     * @date 21th Nov 2014
+     **/
+    {
+      std::map<std::string, unsigned int>::iterator it = access_count_get_model_state_.find(req.model_name);
+      if(it == access_count_get_model_state_.end()) {
+	access_count_get_model_state_.insert( std::pair<std::string, unsigned int>(req.model_name, 1) );
```
ffurrer: no spaces after `(`

ffurrer:
scolgtm

hsu:
+1

ffurrer:
yup nice :)

hsu:
@tfoote Is it ok to release with message data structure change?

gazebo_msgs/srv/GetModelState.srv
```diff
@@ -4,6 +4,10 @@ string relative_entity_name          # return pose and twist relative to this en
                                      # be sure to use gazebo scoped naming notation (e.g. [model_name::body_name])
                                      # leave empty or "world" will use inertial world frame
 ---
+Header header                        # Standard metadata for higher-level stamped data types.
+                                     # * header.seq holds the number of requests since the plugin started
```
tfoote, line 8: The seq value should not be used. If this is an important value you should actually have a field for it. 

Since you're not using the frame_id, you should consider just putting in a named time field.

tfoote:
@hsu changing a service is not strictly forbidden. If it's purely additions like this, as existing code won't fail to compile anymore, but it will change the ABI of anyone building on top requiring a recompile. 

As it's only a service, message migration rules are not an issue. 

The most common approach for this would be to merge it into the pending jade release. If a backport is necessary it can be done, but the communication should be broad enough that all expected effected users hear it and know what they need to update.

hsu:
@tfoote Given jade turtle has already been released, should we target k-turtle due ~5/16?

hsu:
looks like everything has been merged from https://github.com/maxbader/gazebo_ros_pkgs/tree/jade-devel-GetModelState into jade-devel, is this still needed?

scpeters:
@hsu merged #302 into jade-devel on Jan 21st

I would say let's skip it for `indigo-devel` since it's a bit of a change

Issue #274, opened 2014-11-21
cv::resize missing in trusty / indigo


hsu:
```
/home/hsu/projects/catkin_drcsim_drcsim4/src/gazebo_ros_pkgs/gazebo_plugins/src/gazebo_ros_video.cpp: In member function ‘
void gazebo::VideoVisual::render(const cv::Mat&)’:
/home/hsu/projects/catkin_drcsim_drcsim4/src/gazebo_ros_pkgs/gazebo_plugins/src/gazebo_ros_video.cpp:94:7: error: ‘resize’
 is not a member of ‘cv’
       cv::resize(*image_ptr, converted_image, cv::Size(width_, height_));
       ^
make[2]: *** [gazebo_ros_pkgs/gazebo_plugins/CMakeFiles/gazebo_ros_video.dir/src/gazebo_ros_video.cpp.o] Error 1
make[2]: Leaving directory `/home/hsu/projects/catkin_drcsim_drcsim4/build'
make[1]: *** [gazebo_ros_pkgs/gazebo_plugins/CMakeFiles/gazebo_ros_video.dir/all] Error 2
make[1]: Leaving directory `/home/hsu/projects/catkin_drcsim_drcsim4/build'
make: *** [all] Error 2
```

kev-the-dev:
@j-rivero this was fixed.

j-rivero:
closing, thanks!

PR #275, opened 2014-11-21
change header to use opencv2/opencv.hpp issue #274

hsu:


scpeters:
Does this imply any changes about opencv version dependencies (ie. does package.xml need to change)?

j-rivero:
I don't think so: gazebo_plugins is depending on cv_bridge. Which currently resolve in Indigo to:

```
ros-indigo-cv-bridge:
 --> Depends: libopencv-core2.4 , libopencv-dev
```

The change in the PR is getting rid of `cv.h` in favor of `opencv.hpp`.

```
libopencv-dev:
 --> libopencv-core-dev

jrivero@nium ~ $ dpkg -S /usr/include/opencv/cv.h
libcv-dev:amd64: /usr/include/opencv/cv.h
jrivero@nium ~ $ dpkg -S /usr/include/opencv2/opencv.hpp 
libopencv-core-dev:amd64: /usr/include/opencv2/opencv.hpp
```

The change looks good to me since the package selected by cv_bridge is libopencv-core-dev.

PR #279, opened 2014-11-27
temporary hack to **fix** the -J joint position option (issue #93), sleeping for 1 second to avoid race condition. this branch should only be used for debugging, merge only as a last resort.

k-okada:


k-okada:
hydro version of #278

PR #286, opened 2014-12-28
gazebo_ros_control: respect joint limits while measuring joint angle

xqms:
Previously, the logic used to reconstruct joint angles from
Gazebo measurements could end up with joint angles outside the limits.
For example, this occurs when setting joint angles with the
`set_model_configuration` service call.
Example:

lower limit: -2.792
upper limit: +2.792
argument to `set_model_configuration` call: -2.222
calculated joint angle: 4.061

This confuses the PID controllers from `ros_controllers`.

Now, use `angles::shortest_angular_distance_with_limits()`
for revolute joints to calculate a delta which results in an angle
inside our limits.

### DIFF
--- a/gazebo_ros_control/src/default_robot_hw_sim.cpp
+++ b/gazebo_ros_control/src/default_robot_hw_sim.cpp
@@ -246,6 +246,16 @@ void DefaultRobotHWSim::readSim(ros::Time time, ros::Duration period)
     {
       joint_position_[j] = sim_joints_[j]->GetAngle(0).Radian();
     }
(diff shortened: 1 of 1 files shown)


xqms:
Okay, please don't merge this yet. I'm seeing problems with drifting joint angles now, which is also not good. I'll investigate and get back to you with more information...

hsu:
close due to inactivity, please reopen when the drifting problem is understood / resolved. I would advice adding a test to verify the behavior at that time as well. Thanks.

PR #287, opened 2015-01-01
add option to change package:// to model:// when loading urdf file

k-okada:
this feature is required if we want to use gzweb, but ofcourse you have to set GAZEBO_MODEL_PATH correctly


k-okada:
closed due to  wrong target, see new PR  -> https://github.com/ros-simulation/gazebo_ros_pkgs/pull/288

PR #288, opened 2015-01-01
add option to change package:// to model:// when loading urdf file

k-okada:
this feature is required if we want to use gzweb, but ofcourse you have to set GAZEBO_MODEL_PATH correctly

### DIFF
--- a/gazebo_ros/scripts/spawn_model
+++ b/gazebo_ros/scripts/spawn_model
@@ -21,6 +21,7 @@
 import rospy, sys, os, time
 import string
 import warnings
+import re
 
 from gazebo_ros import gazebo_interface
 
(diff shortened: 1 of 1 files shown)


hsu:
@k-okada do you have an example? tutorial? thanks.

130s:
Related discussion https://bitbucket.org/osrf/gzweb/issue/45/creating-local-model-of
@hsu

hsu:
+1

gazebo_ros/scripts/spawn_model
```diff
@@ -243,6 +248,9 @@ class SpawnModel():
           print "Error: user specified param or filename is an empty string"
           sys.exit(0)
 
+        if self.package_to_model:
+            model_xml = re.sub("<\s*mesh\s+filename\s*=\s*\"package:/?/?","<mesh filename=\"model://", model_xml)
```
scpeters, line 252: this only works for attributes with double quotes `"`. I think it should be updated to handle single quotes `'` as well

hsu, line 252: #445

PR #294, opened 2015-02-08
spawn_model: adding joint rosparam

jbohren:
this avoids having to add a whole ton of arguments, instead it's just a rosparam

### DIFF
--- a/gazebo_ros/scripts/spawn_model
+++ b/gazebo_ros/scripts/spawn_model
@@ -83,8 +83,9 @@ class SpawnModel():
         self.wait_for_model_exists   = False
         self.urdf_format             = False
         self.sdf_format              = False
-        self.joint_names             = []
-        self.joint_positions         = []
+        self.joint_names             = rospy.get_param('~initial_joint_names',[])
+        self.joint_positions         = rospy.get_param('~initial_joint_positions',[])
+        self.bond                    = False
 
     def parseUserInputs(self):
         # get goal from commandline
@@ -194,7 +195,6 @@ class SpawnModel():
     def callSpawnService(self):
 
         # wait for model to exist
-        rospy.init_node('spawn_model')
 
         if not self.wait_for_model == "":
           rospy.Subscriber("%s/model_states"%(self.gazebo_namespace), ModelStates, self.checkForModel)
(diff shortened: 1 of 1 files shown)


adolfo-rt:
Style comment: Could you move the changes to `gazebo_plugins/src/gazebo_ros_moveit_planning_scene.cpp` to a separate commit?. It only removes trailing whitespaces.

jbohren:
Yeah, sorry about that. I will.

jbohren:
Fixed.

gazebo_ros/scripts/spawn_model
```diff
@@ -288,6 +288,7 @@ if __name__ == "__main__":
         print usage()
     else:
         print("spawn_model script started") # make this a print incase roscore has not been started
+        rospy.init_node('spawn_model')
```
hsu, line 291: what's the reasoning behind moving this line down?

jbohren, line 291: Can't get ros params without initializing the node.

hsu, line 291: To clarify... the `init_node` call originally exists on line 197, does it make a difference whether `rospy.init_node(...)` is called before or after line 199-200 block?  What happens if we left `rospy.init_node(...)` where it was before?

jbohren, line 291: If you left `rospy.init_node()` inside of `SpawnModel.callSpawnService()`, then it would not get called before `SpawnModel.__init__()` and then would not get called before `rospy.get_param()`.

hsu, line 291: ah, sorry, I misread the indent. got it.

hsu, line 291: got an example that uses this new feature?  thanks.

jbohren, line 291: Example here: https://github.com/jhu-lcsr/lcsr_barrett/blob/devel/launch/sim/wam7_sim.launch#L56

hsu, line 291: got it. added issue #301. I hope one day to enable unit / regression tests for this stack.

Issue #169, opened 2014-02-18
calling /gazebo/reset_simulation seems to break plugins. 


BenBlumer:
I'm running Ubuntu 12.04 with Gazebo 1.9.3 on ROS Hydro. 

I have a model loaded that has a collision sensor.  When I echo the rostopic, I see the expected output.  But when I call `/gazebo/reset_simulation`, that output hangs for a long time -- up to a minute.  This is despite the fact that it only takes 5 seconds to launch Gazebo and get the sensor output to start with.

[I've made a simple proof of concept and uploaded the workspace to github.](https://github.com/BenBlumer/gazebo_issue_demo) 

To replicate, 
- Download and catkin_init the workspace
- `roslaunch BREAKROS_gazebo BREAKROS.launch`
- In another terminal, `rostopic echo /sensored_block_bumper`
- In another terminal `rosservice call /gazebo/reset_simulation`
- You'll see that Gazebo actually resets itself pretty quickly, but the sensor output hangs.

I suspect it's a bug because I had this issue on a real model, I've replicated it with a minimal example (http://answers.gazebosim.org/question/5488/calling-gazeboreset_simulation-breaks-contact/), and it seems like others have had similar problems as well (http://answers.gazebosim.org/question/5194/differential-drive-ros-plugin-stops-responding/).

peci1:
I can't confirm this on Gazebo 2.2/Ubuntu 14.04/ROS Indigo.

meyerj:
I can confirm this issue if a non-zero `updateRate` parameter is set. Some plugins are caching the last update time and only publish if the difference to the current time is greater than `1.0/update_rate_` (The exact variable names are different for different plugins).

Example from [gazebo_ros_imu.cpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_plugins/src/gazebo_ros_imu.cpp):

``` cpp
  // rate control
  if (this->update_rate_ > 0 &&
      (cur_time - this->last_time_).Double() < (1.0 / this->update_rate_))
    return;
```

[...]

PR #298, opened 2015-03-03
[gazebo_ros_diff_drive] force call SetMaxForce

k-okada:
Since this Joint::Reset in gazebo/physics/Joint.cc reset MaxForce to zero and ModelPlugin::Reset is called after Joint::Reset at gazebo2,

this will solve problem such as http://answers.ros.org/question/114931/differential-drive-ros-plugin-stops-responding-after-reset_simulation-or-reset_world/
and partially, https://github.com/ros-simulation/gazebo_ros_pkgs/issues/169


scpeters:
+1

PR #299, opened 2015-03-03
fixed mistake at calculation of joint velocity

sabrina-heerklotz:
Hi,

To calculate the velocity of the joints the wheel radius is used instead of the wheel diameter.

Regards,
Sabrina

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_skid_steer_drive.cpp
+++ b/gazebo_plugins/src/gazebo_ros_skid_steer_drive.cpp
@@ -308,10 +308,10 @@ namespace gazebo {
 
       // Update robot in case new velocities have been requested
       getWheelVelocities();
-      joints[LEFT_FRONT]->SetVelocity(0, wheel_speed_[LEFT_FRONT] / wheel_diameter_);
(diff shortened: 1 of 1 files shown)


k-okada:
sorry for asking very basic question, I thought original code is correct by

```
l = 2 * pi * radius
v[m/s] = r[rad/s] * 2 * radius
r = v / diameter
```

Do I misunderstand something?

sabrina-heerklotz:
Hi,

Sorry for my late reply!

I think your second equation is not correct. Please have a look at my equations:

l = 2 \* pi \* radius
v[m/s] = dx / dt
r[rad/s] = dphi / dt

assuming that there is no acceleration
for a full rotation you get dx = l and dphi = 2 \* pi:
v = l / dt          => dt = l / v = 2 \* pi \* radius / v
r = 2 \* pi /dt   => dt = 2 \* pi / r

eliminating dt gives:
2\* pi \* radius / v = 2 \* pi / r
radius /v = 1/r => r = v / radius

I also checked my code with the simulator:
Using my equation my robot is moving forward with the given linear speed and not with the half of it :)

Regards,
Sabrina

k-okada:
Yeah, exactly, you're correct.

## 

◉ Kei Okada

hsu:
looks good
+1

PR #264, opened 2014-10-21
spawn_model: adding -b option to bond to the model and delete it on sigint

jbohren:

PR #303, opened 2015-03-13
Jhu lcsr forks bond

hsu:
resolving conflict from pulls #264


caguero:
Hydro is EoL. Retargeted to Kinetic in #435,

PR #306, opened 2015-04-01
hog: making only the hand of god link zero-g. 

jbohren:
this lets you use the hand of god on a single link in a gravity-affected model

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_hand_of_god.cpp
+++ b/gazebo_plugins/src/gazebo_ros_hand_of_god.cpp
@@ -102,11 +102,10 @@ namespace gazebo
     // Store the model
     model_ = _parent;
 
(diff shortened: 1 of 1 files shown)


hsu:
should we add an optional sdf element for the plugin to indicate if either the model or the link gravity should be turned off? And default to the old behavior (model)?

scpeters:
We can get this in kinetic if you respond to John's feedback by Wednesday.

hsu:
+1

I am OK for merging into hydro-devel branch, and merge forward. Or if you want to retarget for indigo or jade that works for me too.

This pr does not affect api, so I think we can always merge it into kinetic later right?

j-rivero:
we don't plan to merge forward hydro. Not a bugfix so lets put this in jade directly.

Issue #162, opened 2014-02-07
Consider adding sonar or infrared sensor plugins


hsu:
Inspired by [this pose](http://answers.gazebosim.org/question/5303/is-there-a-gazebo-plugin-that-simulates-sonar-and/?answer=5354#post-id-5354).

Consider moving plugin from [pal_gazebo_plugin](https://github.com/pal-robotics/pal_gazebo_plugins/blob/hydro-devel/src/gazebo_ros_range.cpp) here.

Also, potentially add a new ROS plugin for the [sonar implementation in gazebo](https://bitbucket.org/osrf/gazebo/src/b1dea9012cb37b5afc2336fce5d9f30e7cbc475f/gazebo/sensors/SonarSensor.cc?at=default).

AlexReimann:
Fork with only the ros range message gazebo plugin (without dependencies on other pal stuff for quick tryout):  https://github.com/AlexReimann/pal_gazebo_plugins

Should I make pull request to add this to the gazebo plugin packages?

@po1 Could you confirm if the gazebo_ros_range plugin is stable?

jack-oquin:
+1

bit-pirate:
+1

Ping @adolfo-rt

bmagyar:
+1 from me, and for the question of stability, yes, we've been using this ever since the copyright notice (2013)

adolfo-rt:
+1 for the feature, which means rebasing #311 against at least `indigo-devel`.

j-rivero:
Thanks to the friends of Pal this is now implemented.

PR #311, opened 2015-04-08
Adds range plugin for infrared and ultrasound sensors from pal robotics

AlexReimann:
Pull request for #162 

Only tested if it works, I didn't even look at the source code.

All is contributed from pal robotics

### DIFF
--- a/gazebo_plugins/CMakeLists.txt
+++ b/gazebo_plugins/CMakeLists.txt
@@ -27,6 +27,7 @@ find_package(catkin REQUIRED COMPONENTS
   diagnostic_updater
   camera_info_manager
   moveit_msgs
+  std_msgs
 )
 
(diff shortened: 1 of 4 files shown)


bmagyar:
Good job, however the install rule wasn't updated with the new library. This only works if you compile it and use it from devel space. Do you mind adding it to this PR?
I will make a new PR against indigo and jade as well with the changes.

AlexReimann:
I added it to the install rule. thanks for doing the PR for indigo / jade!

130s:
We may want to close this since Hydro is EOL and https://github.com/ros-simulation/gazebo_ros_pkgs/pull/335 for Indigo is merged.

scpeters:
hydro is EOL

PR #325, opened 2015-05-03
imu supports frameName

leftys:
Imu cannot publish in frames other than bodyName. Thus doesn't work with nested tf like /robot/part1/part2.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_imu.cpp
+++ b/gazebo_plugins/src/gazebo_ros_imu.cpp
@@ -124,6 +124,13 @@ void GazeboRosIMU::LoadThread()
   else
     this->update_rate_ = this->sdf->GetElement("updateRate")->Get<double>();
 
+  if (!this->sdf->HasElement("frameName"))
+  {
+    ROS_INFO("Laser plugin missing <frameName>, defaults to <bodyName>");
+    this->frame_name_ = link_name_;
+  }
+  else
+    this->frame_name_ = this->sdf->Get<std::string>("frameName");
 
   // Make sure the ROS node for Gazebo has already been initialized
   if (!ros::isInitialized())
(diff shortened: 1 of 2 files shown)


gazebo_plugins/src/gazebo_ros_imu.cpp
```diff
@@ -124,6 +124,13 @@ void GazeboRosIMU::LoadThread()
   else
     this->update_rate_ = this->sdf->GetElement("updateRate")->Get<double>();
 
+  if (!this->sdf->HasElement("frameName"))
+  {
+    ROS_INFO("Laser plugin missing <frameName>, defaults to <bodyName>");
```
chapulina: `imu` instead of `Laser`?

chapulina:
Thanks for the contribution! 

Do you happen to have a simple world example using the imu plugin? I noticed there is none in the package yet. It would be great if we could add one to `gazebo_ros_pkgs/gazebo_plugins/test/test_worlds`

leftys:
Hi! I am not using gazebo at the time and forgot most of its usage, so I can't add imu example usage. I fixed the 'Laser' typo however.

scpeters:

scpeters:
the CI is clean now after a rebuild (there was a spurious `apt-get` error)

scpeters:
+1

j-rivero:
@chapulina what you think?

chapulina:
The changes look alright, but I don't know how to test this...

scpeters:
There is an example urdf at http://answers.gazebosim.org/question/4717/imu-sensor-plugin-libgazebo_ros_imuso-strange-data/ but I don't know if it works properly. Or we could add one to the pendulum in gazebo_ros_demos

chapulina:
Ok... We can add an example in the future then. +1

PR #326, opened 2015-05-12
Pull request mimic_joint plugin

vsamy:
It should works for gazebo 4.0 and more recent versions.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_mimic_joint.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_mimic_joint.h
@@ -1,0 +1,80 @@
+#ifndef GAZEBO_ROS_FT_HH
+#define GAZEBO_ROS_FT_HH
+
+#include <gazebo/physics/physics.hh>
+#include <gazebo/common/Plugin.hh>
+#include <gazebo/common/Events.hh>
+#include <ros/ros.h>
+
+namespace gazebo
+{
+/** \defgroup MimicJoint XML Reference and Example
+
+\brief Ros Gazebo Ros MimicJoint Plugin.
+This model allow to mimic a joint without the use of the dynamics
+
+Example Usage:
+
+\verbatim
+<gazebo>
+  <plugin name="mimic_joint" filename="libgazebo_ros_mimic_plugin.so">
+    <jointName>joint_name</jointName>
+    <mimicJoint>mimic_joint_name</mimicJoint>
+    <multiplier>1.0</multiplier>
+  </plugin>
+</gazebo>
+\endverbatim
+\{
+*/
+
+
+/// \brief GazeboRosFT controller
+/// This is a controller that simulates a 6 dof force sensor
+class MimicJoint: public ModelPlugin
+{
+  /// \brief Constructor
+  /// \param parent The parent entity must be a Model
+  public: MimicJoint();
+
+  /// \brief Destructor
+  public: virtual ~MimicJoint();
+
+  /// \brief Load the controller
+  public: void Load( physics::ModelPtr _parent, sdf::ElementPtr _sdf );
+
+  /// \brief Update the controller
+  protected: virtual void UpdateChild();
+  
+  /// \brief multiplier
+  private: double multiplier_;
+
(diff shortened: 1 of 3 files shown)


gazebo_plugins/include/gazebo_plugins/gazebo_ros_mimic_joint.h
```diff
@@ -0,0 +1,80 @@
+#ifndef GAZEBO_ROS_FT_HH
```
adolfo-rt: Check this header guard. It looks as if it was copy-pasted from another plugin.

adolfo-rt:
Consider supporting not only the mimic joint multiplier, but also an offset (or bias) value. This would allow for instance to represent [URDF-style mimic tags](http://wiki.ros.org/urdf/XML/joint).

vsamy:
i updated the files to take into account the bias and rename the header guard. 
I tested it on gazebo and it works fine

gazebo_plugins/include/gazebo_plugins/gazebo_ros_mimic_joint.h
```diff
@@ -0,0 +1,83 @@
...
+  public: void Load( physics::ModelPtr _parent, sdf::ElementPtr _sdf );
+
+  /// \brief Update the controller
+  protected: virtual void UpdateChild();
+  
+  /// \brief multiplier
+  private: double multiplier_;
+
+  /// \brief multiplier
+  private: double bias_;
```
adolfo-rt: Why not use `offset`, which is the same terminology used in the URDF tags?

gazebo_plugins/include/gazebo_plugins/gazebo_ros_mimic_joint.h
```diff
@@ -0,0 +1,83 @@
...
+This model allow to mimic a joint without the use of the dynamics
+
+Example Usage:
+
+\verbatim
+<gazebo>
+  <plugin name="mimic_joint" filename="libgazebo_ros_mimic_plugin.so">
+    <jointName>joint_name</jointName>
+    <mimicJoint>mimic_joint_name</mimicJoint>
+    <multiplier>1.0</multiplier>
```
adolfo-rt, line 23: Please update example with offset specification

vsamy:
hi, i changed the bias name to be consistent with the urdf-style. The example is also updated.
Do you see anything which could be improved ?

adolfo-rt:
I'll let the Gazebo devs comment on the actual plugin implementation. I'm not familiar enough with the Gazebo coding style and practices to comment on that.

vsamy:
It seems that the script is highly unstable when using with dynamics.
It should ok for kinematics simulation though.

vsamy:
I updated the plugin so that the plugin is not using SetPosition which makes everything unstable.
I created simple proportionnal gain and use the SetForce method. It works fine on my simulations.

scpeters:
Can you resolve the conflicts?

vsamy:
What conflicts are you talking about ?
It's an old plugin. Normally, a SetPosition should be better for mimic joints but it does not on Gazebo. I didn't dig further. What exactly do you want ?

scpeters:
Sorry, I meant could you rebase the branch or merge with `jade-devel`. The github UI says "This branch has conflicts that must be resolved"

j-rivero:
@scpeters is right, we need to get the conflicts resolved before we can merge this PR.

vsamy:
Hi,

Sorry, i had completely forgot about this.
My main problem with those files is that, even it is created for handling mimic joints it is using a proportionnal controller. 
Unfortunately this is no mimic joint. We should use a "SetPosition" like function but after trying the results were totally random that's why i used a more common controller. 

Do you still want to merge this ?

j-rivero:
> My main problem with those files is that, even it is created for handling mimic joints it is using a proportionnal controller.
> Unfortunately this is no mimic joint. We should use a "SetPosition" like function but after trying the results were totally random that's why i used a more common controller.
> 
> Do you still want to merge this ?

Thanks Vincent.

@hsu, @scpeters, @nkoenig may I have your opinion about Vicent question?

PR #328, opened 2015-06-12
plugins: Fixing race condition that prevents stereo cameras from broadcasting

jbohren:
Related to https://github.com/ros-simulation/gazebo_ros_pkgs/pull/17

It appears that if both of these image transport publishers are created at the same time, they won't be able to load the same transport plugins. This is probably due to a bug in `image_transport` itself.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_camera_utils.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera_utils.cpp
@@ -258,6 +258,10 @@ event::ConnectionPtr GazeboRosCameraUtils::OnLoad(const boost::function<void()>&
 // Load the controller
 void GazeboRosCameraUtils::LoadThread()
 {
+  // Synchronize loading for multicamera setup, without this, not all transport
+  // topics may be advertised. This might be due to a bug in image_transport
+  boost::mutex::scoped_lock lock(*this->image_connect_count_lock_);
+
   // Exit if no ROS
   if (!ros::isInitialized())
   {


jackieokay:
I wasn't able to reproduce the issue on Indigo with two depth cameras:

https://gist.github.com/jacquelinekay/065d906a5888f3cdac4fde5ee410a900

From inspecting the output of rosparam list, dynparam list, and rostopic list, everything looks ok.

It's possible this issue was fixed in Indigo in `image_transport`, or it's possible that my test case isn't enough to reproduce the race condition. If you can provide me with a world file that reproduces the issue in Indigo I'll forward-port this to Indigo and possible Jade and Kinetic. Otherwise I'm a little hesitant to introduce random mutexes throughout the code.

hsu:
I wonder if it's related to [this patch](https://bitbucket.org/osrf/drcsim/commits/2cf768ccbc9e8d4cd5b8daa72e8099c515a7864f#Lros/atlas_msgs/gazebo_ros_camera_utils.cppT239), or seen [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_plugins/src/gazebo_ros_camera_utils.cpp#L270)

jbohren:
> It's possible this issue was fixed in Indigo in image_transport, or it's possible that my test case isn't enough to reproduce the race condition. If you can provide me with a world file that reproduces the issue in Indigo I'll forward-port this to Indigo and possible Jade and Kinetic.

I doubt it was fixed in `image_transport`, since I don't see any references to this problem or thread-safety in the commit log.

>  Otherwise I'm a little hesitant to introduce random mutexes throughout the code.

Yeah I wouldn't want to introduce random mutexes throughoug the code either, but this patch resolved the race condition completely for me at a time when it was occurring very frequently. I put it there based on the effect I was seeing and what looked like the weakest part of the code:
- In `GazeboRosMultiCamera::Load()`, multiple `GazeboRosCameraUtils` objects are instantiated
- For each of these objects, `GazeboRosMultiCamera::Load()` calls `GazeboRosCameraUtils::Load()` which subsequently spawns a thread from `GazeboRosCameraUtils::LoadThread()`

Therefore, multiple calls to `GazeboRosCameraUtils::LoadThread()` are very likely to happen concurrently. When adding a mutex to synchronize this loading, it kept the two `ImageTransport`s from being instantiated at the same time, and immediately resolved the problem. It appeared that instantiating multiple `ImageTransport`s at the same time was not thread-safe (with respect to transport loading), which is why I said that it was probably a problem within `image_transport`.

It probably went deeper than that, and was actually a problem in `pluginlib` or `class_loader`. I would have expected it to have been resolved by https://github.com/ros/class_loader/pull/16 (which I just found now), since that issue appears to describe exactly the problem that I was observing. One camera would have some transports, and the other camera would have the remaining transports. It's surprising that I would have still been observing this problem a whole year after it was fixed in `class_loader` 0.2.4, but it's possible that the system install on the machine I was using was out of date.

Since you can't repro it, I'll close it unless I see it again with indigo/stock libs.

PR #329, opened 2015-06-15
ros_api_plugin: Throttle clock message to 100Hz by default, make it parameterizable by a rosparam

jbohren:
By default, if running gazebo with a 1Khz update rate, the gazebo ROS API plugin will attempt to publish the simulation time to the `/clock` topic at that rate. This means that Gazebo opens one TCP socket for _each_ ROS node, and tries to publish `/clock` messages at 1KHz to each node. This leads to a serious resource waste across an entire ROS system.

This PR throttles that publication to 100Hz by default, and makes it configurable with the ROS parameter `/sim_time_rate`.

When publishing at 100Hz, this has the following effects (on an Intel Xeon E5-1650 workstation):
- Gazebo itself uses 50% less CPU time 
- All trivial or idle python processes consume 1% CPU time instead of a minimum of 10% CPU time

This bug is related to: http://answers.ros.org/question/29425/rospy-horrible-performance-with-sim_time/?answer=29456#post-id-29456

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -30,8 +30,9 @@ namespace gazebo
 GazeboRosApiPlugin::GazeboRosApiPlugin() :
   physics_reconfigure_initialized_(false),
   world_created_(false),
+  plugin_loaded_(false),
   stop_(false),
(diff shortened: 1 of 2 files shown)


scpeters:
see also #345

hsu:
Hydro is EOL. #345 should also address this issue. Please retarget if needed.

PR #331, opened 2015-06-25
Fix crash due to world disconnect event

iche033:
With recent changes in gazebo event system, event disconnection within the event callback no longer works, otherwise it causes a crash.

reproducible by compiling against latest gazebo default branch and running:

```
roslaunch gazebo_ros empty_world
```

This PR moves the disconnection of the `worldCreated` event to the desctructor.

If this looks good, I can create a separate PR targeted at `jade-devel` if needed


nkoenig:
+1

nkoenig:
https://bitbucket.org/osrf/gazebo/pull-request/1790 should fix the problem

nkoenig:
+1

PR #332, opened 2015-06-28
gazebo_ros_control: Fix DefaultRobotHWSim puts robotNamespace twice

akio:
DefaultRobotHWSim::initSim() member function uses both
namespaced NodeHandle(model_nh) and robot_namespace string to create
parameter names.

For example,  if a robotNamespace is "rrbot",
DefaultRobotHWSim tries to get parameters from following names:
- /rrbot/rrbot/gazebo_ros_control/pid_gains/*
- /rrbot/rrbot/joint_limits/*

This commit change these names to:
- /rrbot/gazebo_ros_control/pid_gains/*
- /rrbot/joint_limits/*

### DIFF
--- a/gazebo_ros_control/src/default_robot_hw_sim.cpp
+++ b/gazebo_ros_control/src/default_robot_hw_sim.cpp
@@ -65,7 +65,7 @@ bool DefaultRobotHWSim::initSim(
 {
   // getJointLimits() searches joint_limit_nh for joint limit parameters. The format of each
   // parameter's name is "joint_limits/<joint name>". An example is "joint_limits/axle_joint".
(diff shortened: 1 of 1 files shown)


adolfo-rt:
You are right. The `model_nh` argument [is already namespaced](https://github.com/akio/gazebo_ros_pkgs/blob/fix-ros-control-param-ns/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp#L141). Thanks for the fix.

130s:
Is this supposed to be cherry-picked toward `jade-devel` too?

adolfo-rt:
Yes, #339.

PR #335, opened 2015-07-09
Adds range plugin for infrared and ultrasound sensors from PAL Robotics

bmagyar:
Rebase of #311 to indigo-devel + added install rule for new plugin and squashed 'em together.

Solves #162 on ROS Indigo.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_range.cpp
+++ b/gazebo_plugins/src/gazebo_ros_range.cpp
@@ -1,0 +1,336 @@
+/*
+ * Software License Agreement (Modified BSD License)
+ *
+ *  Copyright (c) 2013-2015, PAL Robotics, S.L.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of PAL Robotics, S.L. nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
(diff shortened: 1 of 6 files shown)


bmagyar:
@j-rivero

gazebo_plugins/src/gazebo_ros_range.cpp
```diff
@@ -0,0 +1,403 @@
...
+    ROS_INFO("Laser plugin missing <frameName>, defaults to /world");
+    this->frame_name_ = "/world";
+  }
+  else
+    this->frame_name_ = this->sdf->Get<std::string>("frameName");
+
+  if (!this->sdf->HasElement("topicName"))
+  {
+    ROS_INFO("Laser plugin missing <topicName>, defaults to /world");
+    this->topic_name_ = "/world";
```
gavanderhoorn: Is this copy/pasta from the `frameName` check above? I'm really not a Gazebo plugin expert, but this seems like a strange topic name for a sonar sensor?

bmagyar: Oh it is indeed copy pasta with tomato sauce. Also the message there make no sense, I shall fix them shortly.

bmagyar:
@scpeters, what do you think?

scpeters:
+1

rethink-imcmahon:
+1 great addition @AlexReimann & @bmagyar

j-rivero:
Amazing! Thanks Ben! Would you mind to include a quick test world file in `gazebo_plugins/test/range`?

bmagyar:
Only a .world file? If I need anything else, please you point me at an example from the test folder.

j-rivero:
The point is to have something to demonstrate and run the code for testing and learning proposes. I think that a world file + launch file could be a good combo. Give a look to what Nate implemented in pull request for #338.

Automatic testing would be ideal, but we probably need to setup a proper mechanism in place first to be shared by the different plugins.

bmagyar:
Took some time to get back here, but I finally managed to add a world file + a launch file.

There are 2 boxes in the world, both with a range sensor. One looks at the other, the other looks into the horizon like a true western hero.

To launch and check:

`roslaunch gazebo_ros range.world.launch`
`rostopic echo /sonar`
`rostopic echo /sonar2`

There is a gaussian noise on the sensors, so the perfect 0.25 measurement hardly appears, but it works well.

Now I will try to add a rostest in a new commit.

j-rivero:
> Took some time to get back here, but I finally managed to add a world file + a launch file.

Thanks very much Bence for taking the time to implement it. I was able to build your patch built on top of pull request #350 using gazebo5 and gazebo6 on Trusty/Indigo. 

> Now I will try to add a rostest in a new commit.

I will pay the beers if you made that :) Since I'm planning on releasing (issue #335), I want to get this in, so I will merge it just after pull request #350. Feel free to send a new pull request for the rostest support.

bmagyar:
Great!
This means that gazebo plugins will be synced for now between indigo and
jade? Good news for my patch for the block laser ;)

j-rivero:
> This means that gazebo plugins will be synced for now between indigo and jade?

That should have been done in pull request #350 and this PR will go into jade in pull request #352

PR #314, opened 2015-04-15
Set GAZEBO_CXX_FLAGS to fix c++11 compilation errors

scpeters:
sdformat version 3 uses c++11 features in its header files. The necessary compiler flags are exported in `GAZEBO_CXX_FLAGS`, and these must be set in order to avoid compiler errors. This is most relevant for gazebo6.

See the following:
- [gazebo pull request 1573](https://bitbucket.org/osrf/gazebo/pull-request/1573/use-flags-in-example-code-cmakeliststxt/diff)
- [sdformat pull request 172](https://bitbucket.org/osrf/sdformat/pull-request/172/add-std-c-11-flag-to-sdf_configcmakein-and/diff)

PR #336, opened 2015-07-10
Use c++11

nkoenig:
Gazebo 5 and above require c++11.


scpeters:
Looks good. This was done for `indigo-devel` ( #314 ), but it seems that we haven't merged forward all the pull requests from indigo to jade. We might want to look at what's missing and merge forward.

+1

PR #345, opened 2015-08-10
support to change clock frequncy from ros param

k-okada:
When we use gazebo with ros application, we have serious performance issue, mainly due to 
http://answers.ros.org/question/29425/rospy-horrible-performance-with-sim_time/ problem.

As suggested avove answers, we'd like to run gazebo with lower clock rate (but not to slow down gazebo simulation step for physics stability)

This PR supports pub_clock_frequency param in order to change /clock frequency, see modified empty_world.launch for usage.


130s:
+1

hsu:
Please fix conflicts

hsu:
moved to #442

PR #353, opened 2015-08-21
Add rostest to accompany range plugin world

bmagyar:
I wrote this yesterday on the airplane. I am able to run the test individually, but when I ran `catkin_make run_tests` it failed on the set model test. It seems that the tests are launched all at once, so it attempts to spawn several gazebo instances which conflict with each other.

Take a look and let me know what do you think. The organization of files could be better, maybe there is time for a general reorganization so similar tests can be made for the other plugins as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ script:
   - cd ~/ros/ws_$REPOSITORY_NAME
   - catkin build -j4 --verbose --summary
   # Run tests
-  - catkin run_tests
+  - catkin run_tests -j1 # have to run only 1 test at a time to ensure no gazebo instances collide
   # catkin run_tests always returns 0, use catkin_test_results to check for errors
   # https://github.com/catkin/catkin_tools/issues/245
   - catkin_test_results
--- a/gazebo_plugins/CMakeLists.txt
+++ b/gazebo_plugins/CMakeLists.txt
@@ -342,5 +342,6 @@ if (CATKIN_ENABLE_TESTING)
   add_rostest_gtest(set_model_state-test
                     test/set_model_state_test/set_model_state_test.test 
                     test/set_model_state_test/set_model_state_test.cpp)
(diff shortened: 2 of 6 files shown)


j-rivero:
Hey Bence:

Could you please merge this branch with indigo-devel to get some travis fun for the new code?

Thanks.

bmagyar:
Hello señor, 
Done :+1:

bmagyar:
So, the 1 failing check happens when you run `catkin_make run_tests`.
Running the tests individually works but `run_tests` tries to start all of them at once. Since they all launch their own instance of gazebo, these collide and one is doomed to fail.

Any ideas how to make test execution linear? @j-rivero

j-rivero:
> Any ideas how to make test execution linear? @j-rivero

Looking into the log, it could be that the `-j` [parameter](https://cmake.org/cmake/help/v2.8.9/ctest.html#opt%3a-jjobs--paralleljobs) is controlling the number of testing jobs to launch, so we could try `catkin run_tests --catkin-make-args -j1` in the `travis.yml` file.

bmagyar:
Correct. Should I add that fix to this PR too?

j-rivero:
> Correct. Should I add that fix to this PR too?

green light, yes.

bmagyar:
All green! This should enable tests to be made for all the other plugins too (actually probably they only need to be added to the cmake)

j-rivero:
> All green! This should enable tests to be made for all the other plugins too (actually probably they only need to be added to the cmake)

I'm on it ;). Thanks Bence!

bmagyar:
If encounter weird issues have a look at what i did to the .test file of
the range plugin. It's a bit nasty but without explicitly setting the
gazebo plugin path it didn't take it up...

j-rivero:
> If encounter weird issues have a look at what i did to the .test file of the range plugin.

Umm interesting, thanks for the tip.

PR #357, opened 2015-09-12
Minor: Added a missing variable initialization inside Differential Drive

MirkoFerrati:
A variable inside the differential drive plugin was not initialized to zero in the constructor, this caused random speeds when spawning more than one differential drive plugin into gazebo (e.g. more than one iRobot Create).
In order to reply the error, add multiple robots that use a differential drive, and send them a cmd_vel with x>0.
Searching the code for the variable wheel_speed_instr_ will show that it was used and modified but never set (lines 275 and below), so even with a single robot the error could happen.


MirkoFerrati:
This fix should also be ported to jade_devel, but I am not able to test it on my indigo set-up.

scpeters:
+1

PR #362, opened 2015-10-06
[gazebo_ros] fixes #361

### LINKED ISSUE #361: [gazebo_ros] spawning too slow if lot of resources in URDF
We developed a tool that reads some extra tags in a URDF file. These tags are not valid URDF tags nor valid SDF tags, just additional XML tags that gazebo should ignore.
Unfortunately inside these tags we re-used the similar syntax for resources since our tool uses rviz actually which can also resolve resources. They look something like this :

```
 <tactile topic="tactile">
      <origin rpy="0 0 0" xyz="0 0 0"/>
      <geometry>
        <mesh filename="package://some_package_name/some_path/taxel.stl"/>
      </geometry>
    </tactile>
```

We have 200 of similar entries in our URDF and due to these additional tags, gazebo_ros model spawner gets so slow, more than 30 seconds, that our controllers spawner timeout and models are not cleanly loaded now.

I guess the culprit is this :
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_ros/src/gazebo_ros_api_plugin.cpp#L553

any `package://` is resolved, with no cache, even for non SDF/URDF tags. This takes ages and is completely useless as gazebo will not read these resolved resources anyway.

I think adding caching to the resource resolver (our 200 tags have the same package source for instance) would help, even for other models with a lot of resources (typically a hand with a lot of links).
Not taking care of non URDF/SDF tags would be even better but more complex to implement in the current structure (separated resolving and urdf/sdf spawning)

Do you think that is an easy fix ? I could probably try to add a very basic package name cache in this local resolver if you think that is the way to go.

best
Guillaume

guihomework:
Implemented a dictionary of resolved package path in the URDF spawner to speed up the process.
Also not starting all over finding new occurrences of package:// as suggested by @rhaschke

Not starting all over did not improve significantly the time, but the dictionary improved spawning from 15.5 sec down to 2.7 sec on our file and reduced loading time of the bare shadow hand (without 200 tactile elements) by more than 7 seconds.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -577,7 +577,7 @@ bool GazeboRosApiPlugin::spawnURDFModel(gazebo_msgs::SpawnModel::Request &req,
       ROS_DEBUG_ONCE("Package name [%s] has path [%s]", package_name.c_str(), package_path.c_str());
 
       model_xml.replace(pos1,(pos2-pos1),package_path);
-      pos1 = model_xml.find(package_prefix,0);
+      pos1 = model_xml.find(package_prefix, pos1);
     }
   }
   // ROS_DEBUG("Model XML\n\n%s\n\n ",model_xml.c_str());


gazebo_ros/src/gazebo_ros_api_plugin.h
```diff
@@ -28,6 +28,7 @@
 #include <signal.h>
 #include <errno.h>
 #include <iostream>
+#include <map>
```
rhaschke: If you don't need map in the header, you should include it in cpp only.

rhaschke:
[rospack PR #49](https://github.com/ros/rospack/pull/49) and [roslib PR #97](https://github.com/ros/ros/pull/97) will solve the caching issue deeper in the call chain.

hsu:
+1

one more approval and we can merge this. Thanks!

guihomework:
Removed the last commit about caching, as the caching issue is solved in a better fashion lower down the chain in rospack in https://github.com/ros/rospack/pull/49 , and produces very similar results in terms of timing.

The other commit fixing the reading the urdf from start everytime one package has been resolved should still be merged.

scpeters:
+1

PR #364, opened 2015-10-24
[gazebo_ros] Fix invalid signal name on OS X for script/gazebo

bgromov:
Fixes the following error on OS X

```
scripts/gazebo: line 30: kill: SIGINT: invalid signal specification
```

by feeding system-dependent argument to `kill` command. On OS X it is `kill -s INT` instead of `kill -s SIGINT`.

### DIFF
--- a/gazebo_ros/scripts/gazebo
+++ b/gazebo_ros/scripts/gazebo
@@ -2,8 +2,10 @@
 final="$@"
 
 EXT=so
+SIGNAL=SIGINT
 if [ $(uname) = "Darwin" ]; then
     EXT=dylib
+    SIGNAL=INT
 fi
 
 # add ros path plugin if it does not already exist in the passed in arguments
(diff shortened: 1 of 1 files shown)


scpeters:
+1

scpeters:
Thanks! I cherry-picked to `indigo-devel` in 03a1e27d59c34f1494c1bb586f85de701e8d6661

dirk-thomas:
For me on Xenial I do have the same problem. When running the `gazebo`script it fails to kill `gzserver` at the end with the following error message:

```
kill: invalid signal number or name: SIGINT
```

When changing the signal name to `INT` (as this PR does for OS X) it works for me on Xenial.

Another reference mentioning this: http://stackoverflow.com/questions/16423732/linux-shell-kill-signal-sigkill-kill

dirk-thomas:
@nkoenig just mentioned to me that the problem will be addressed by #472 (and a new patch release of Gazebo 7).

scpeters:
I'm not sure if #472 will help, since it wouldn't work with the upstream Ubuntu gazebo 7.0.0

what if we just switch to `kill -n 2` instead of trying to name the signal?

PR #368, opened 2015-11-18
Covariance for published twist in skid steer plugin

l0g1x:
As packages such as robot_localization require an associated non-zero covariance matrix to the input source, I thought I would submit the changes I made in order to allow for the skid steer plugin to work with robot_localization.


l0g1x:
is this merged into indigo-devel as well?

PR #370, opened 2015-11-20
Allow offsetting /clock sim_time by the current wall time on startup

paulbovbel:
Use Case: we're collecting, fusing and analyzing historical data, both with real robots and simulations (running at near-realtime). Since the data is collected across topics and timestamped with ros::Time, inconveniently all data from the simulator becomes impractical to filter using time criteria, since every iteration of the sim starts at January, 1970. 

This pull allows setting a /sim_wall_time_offset parameter, which will offset all clock messages with the current wall time.


paulbovbel:
o

paulbovbel:
Closing, not the right place to fix this since outgoing plugin messages will still have the non-offset timestamp.

PR #373, opened 2015-12-09
[gazebo_plugins] Publish organized point cloud from openni_kinect plugin

wkentaro:


wkentaro:
Actually I don't have time to identify the actual cause, but `FillPointCloudHelper` seems set height as 1 for the PointCloud2 message.
To get organized point cloud sensor data, I need this change.

furushchev:
Any update? This is indispensable for working pr2 with gazebo properly

furushchev:
This is good to me.
@hsu How do you think of it?

furushchev:
ping

furushchev:
@hsu ping
No maintainer of this package?

wkentaro:
I took a look at the code and found the actual cause of this problem.
I updated the commit, so please review and merge.

gazebo_plugins/src/gazebo_ros_openni_kinect.cpp
```diff
@@ -306,6 +306,7 @@ bool GazeboRosOpenniKinect::FillPointCloudHelper(
 {
   sensor_msgs::PointCloud2Modifier pcd_modifier(point_cloud_msg);
   pcd_modifier.setPointCloud2FieldsByString(2, "xyz", "rgb");
+  // convert to flat array shape, we need to reconvert later
   pcd_modifier.resize(rows_arg*cols_arg);
```
wkentaro, line 310: this converts `point_cloud_msg` as height=1 and width=SIZE, so we need reassign the value of height and width for the point cloud msg.

hsu:
sorry for the lat reply. Looks good, thanks.

PR #380, opened 2016-01-13
Fix gazebo6 deprecation warnings

scpeters:
Several RaySensor functions are deprecated in gazebo6
and are removed in gazebo7.
The return type is changed to use ignition math
and the function name is changed.
This adds ifdef's to handle the changes.


hsu:
+1

PR #375, opened 2015-12-18
Updated gazebo_ros_pkg to latest gazebo api (gazebo 7.0.1).

carlosmccosta:
- Gazebo classes in which the api changed:
  Joint
  RaySensor
- Added missing boost includes.

PR #381, opened 2016-01-15
Gazebo7 fixes

scpeters:
This fixes three separate build errors when using gazebo7 (related to #375):
- a27311b: a boost header was removed from a gazebo header and was being used implicitly by gazebo_ros_joint_state_publisher.cpp
- ce0ab10: resolve ambiguity between overloaded functions ( [upstream issue](https://bitbucket.org/osrf/gazebo/issues/1830) ).
- 194ebf8: sensor pointers changed from boost to std shared pointers. I've added a macro called `GAZEBO_SENSORS_USING_DYNAMIC_POINTER_CAST` to `gazebo_ros_utils.h` that provides a `using` statement so that `dynamic_pointer_cast` can be used more easily.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h
@@ -41,6 +41,14 @@
 #include <gazebo/sensors/Sensor.hh>
 #include <ros/ros.h>
 
+#ifndef GAZEBO_SENSORS_USING_DYNAMIC_POINTER_CAST
+# if GAZEBO_MAJOR_VERSION >= 7
+#define GAZEBO_SENSORS_USING_DYNAMIC_POINTER_CAST using std::dynamic_pointer_cast
+# else
+#define GAZEBO_SENSORS_USING_DYNAMIC_POINTER_CAST using boost::dynamic_pointer_cast
+# endif
+#endif
+
 namespace gazebo
 {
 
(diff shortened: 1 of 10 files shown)


gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h
```diff
@@ -41,6 +41,14 @@
 #include <gazebo/sensors/Sensor.hh>
 #include <ros/ros.h>
 
+#ifndef GAZEBO_SENSORS_USING_DYNAMIC_POINTER_CAST
+# if GAZEBO_MAJOR_VERSION >= 7
+#define GAZEBO_SENSORS_USING_DYNAMIC_POINTER_CAST using std::dynamic_pointer_cast
```
hsu, line 47: just for my own education, what's the reasoning behind using `using` as opposed to making `xxx::dynamic_pointer_cast` a macro?

hsu:
other than the comment about the macro, everything looks good and compiles. Thanks!

gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h
scpeters, line 47: In urdfdom we made [wrapper functions in types.h](https://github.com/ros/urdfdom_headers/blob/master/urdf_model/include/urdf_model/types.h#L72-L95) so you could call `urdf::dynamic_pointer_cast`. In this case, we haven't switched all of gazebo's pointers yet (sensors uses `std::` in gazebo7 while physics still uses `boost::`). It seemed like these `using` macros (which can be scoped) would be a more compact way to allow for different pointer types.

hsu:
+1

PR #389, opened 2016-02-11
[gazebo_plugins] Add vacuum gripper plugin

wkentaro:
this plugin is not perfect:
- apply forces to link's coords not collision
- do not apply friction actually

rrbot_sample: 
- code: https://github.com/wkentaro/gazebo_ros_demos/commit/93ab6f1b3bd02687436a9ef483bb6904917d4e67
- demo_movie: https://drive.google.com/file/d/0B9P1L--7Wd2vd0YzTkkwRFh3SG8/view?usp=sharing

baxter_sample:
- code: https://github.com/start-jsk/jsk_apc/blob/76db5dd1aab662f55a71e7e3157789ca1c018e21/jsk_2015_05_baxter_apc/robots/baxter.gazebo#L39-L69
- demo_movies
  - https://drive.google.com/file/d/0B9P1L--7Wd2va1JSQ1U4WkkzMnM/view?usp=sharing
  - https://www.youtube.com/watch?v=uV6XctamwEA

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_vacuum_gripper.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_vacuum_gripper.h
@@ -1,0 +1,161 @@
+/*
+ * Copyright (C) 2012-2014 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+/*
+   Desc: GazeboVacuumGripper plugin for manipulating objects in Gazebo
+ * Author: Kentaro Wada
+ * Date: 7 Dec 2015
+ */
+
+#ifndef GAZEBO_ROS_VACUUM_GRIPPER_HH
+#define GAZEBO_ROS_VACUUM_GRIPPER_HH
+
+#include <string>
+
+// Custom Callback Queue
+#include <ros/callback_queue.h>
+#include <ros/advertise_options.h>
+#include <ros/advertise_service_options.h>
+#include <std_srvs/Empty.h>
+
+#include <ros/ros.h>
+#include <boost/thread.hpp>
+#include <boost/thread/mutex.hpp>
+
+#include <gazebo/physics/physics.hh>
+#include <gazebo/transport/TransportTypes.hh>
+#include <gazebo/common/Plugin.hh>
+#include <gazebo/common/Events.hh>
+
+
+namespace gazebo
+{
+/// @addtogroup gazebo_dynamic_plugins Gazebo ROS Dynamic Plugins
+/// @{
+/** \defgroup GazeboRosVacuumGripper Plugin XML Reference and Example
+
+  \brief Ros Vacuum Gripper Plugin.
+
+  This is a Plugin that collects data from a ROS topic and applies wrench to a body accordingly.
+
+  Example Usage:
+    - left_end_effector will be the power point
+    - revolute type joint is necessary (fixed joint disappears on gazebo and plugin can't find the joint and link)
+  \verbatim
+    <link name="left_end_effector">
+      <gravity>0</gravity>
+      <visual>
+        <origin rpy="0 0 0" xyz="0 0 0"/>
+        <geometry>
+          <box size="0.01 0.01 0.01"/>
+        </geometry>
+        <material name="transparent">
+          <color rgba="0 0 0 0"/>
+        </material>
+      </visual>
+      <inertial>
+        <origin rpy="0 0 0" xyz="0.000000 0.000000 0.000000"/>
+        <mass value="0.0001"/>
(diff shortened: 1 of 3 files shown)


rethink-imcmahon:
Huge +1. This is so cool, @wkentaro! Well done generalizing the interface to a plugin as well.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_vacuum_gripper.h
```diff
@@ -0,0 +1,161 @@
+/*
+ * Copyright (C) 2012-2014 Open Source Robotics Foundation
```
scpeters, line 2: This date could be updated to 2015-2016

scpeters:
closing and re-opening to trigger CI builds

gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
```diff
@@ -0,0 +1,247 @@
...
+  for (size_t i = 0; i < models.size(); i++) {
+    if (models[i]->GetName() == link_->GetName() ||
+        models[i]->GetName() == parent_->GetName())
+    {
+      continue;
+    }
+    physics::Link_V links = models[i]->GetLinks();
+    for (size_t j = 0; j < links.size(); j++) {
+      math::Pose link_pose = links[j]->GetWorldPose();
+      math::Pose diff = parent_pose - link_pose;
```
scpeters, line 193: This uses the difference in link pose between the end effector and an object to determine whether they are close enough for force to be applied, but this depends on where the coordinate frames are defined. This might work for small objects, but it would be better to use collision checking (such as ray-casting) to determine the proximity to the actual surface, which won't be affected by how coordinate frames are defined.

Let me know if that didn't make sense and we can discuss further.

scpeters, line 193: now I realize that you said exactly this in the description of the pull request

gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
```diff
@@ -0,0 +1,247 @@
...
+        if (norm < 0.01) {
+          // apply friction like force
+          // TODO(unknown): should apply friction actually
+          link_pose.Set(parent_pose.pos, link_pose.rot);
+          links[j]->SetWorldPose(link_pose);
+        }
+        if (norm_force > 20) {
+          norm_force = 20;  // max_force
+        }
+        math::Vector3 force = norm_force * diff.pos.Normalize();
```
scpeters, line 210: it also applies the force in every direction, when it should apply it in a single direction only. We would need a way to specify this direction to implement this

gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
```diff
@@ -0,0 +1,247 @@
...
+        links[j]->SetAngularVel(link_->GetWorldAngularVel());
+        double norm_force = 1 / norm;
+        if (norm < 0.01) {
+          // apply friction like force
+          // TODO(unknown): should apply friction actually
+          link_pose.Set(parent_pose.pos, link_pose.rot);
+          links[j]->SetWorldPose(link_pose);
+        }
+        if (norm_force > 20) {
+          norm_force = 20;  // max_force
```
scpeters, line 208: it would be nice to be able to specify the `norm_force` and `norm` parameters via the plugin xml tags

scpeters:
This is a working plugin even if it isn't perfect, as you stated. I'll be ready to merge this if the copyright year is updated. We can create an issue to address the other comments I made (poses of coordinate frames instead of collision proximity, directionality of force, parameterization).

j-rivero:
> I'll be ready to merge this if the copyright year is updated.

I will create a new pull request to fix the problem with the copyright.

Thanks to everyone for the pull request.

gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
wkentaro, line 193: Actually, I don't know how to get the distance between robot link and object collision for now.
Do you know how to do this? Or example?

wkentaro:
> I will create a new pull request to fix the problem with the copyright.
> Thanks to everyone for the pull request.

Thanks.

gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
wkentaro, line 210: Yeah, it should be.
Maybe, It will be not single direction but restricted.

wkentaro:
> This is a working plugin even if it isn't perfect, as you stated. I'll be ready to merge this if the copyright year is updated. We can create an issue to address the other comments I made (poses of coordinate frames instead of collision proximity, directionality of force, parameterization).

Thanks for the review. I watch the issue on https://github.com/ros-simulation/gazebo_ros_pkgs/issues/446 and will work for this when I have time.

gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
abdullahmohiuddin: Wada San, 
This is a wonderful , plugin and very helpful. How can I do a specific modification, such as applying force to the links position (x,y,z) with an offset, (x+10,y,z)
This way, I can solve a specifc problem, that I have, i.e. to apply force at the corners of a large cylindrical object, rather than applying force on the center of the cylinder?

wkentaro: I think you should change the direction of the force.
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp#L210-L211
Currently the direction is from gripper link to object link.

Issue #387, opened 2016-02-03
Support to use gazebo and ROS parameters when using rosrun


peci1:
If I try to run `gzserver` from a roslaunch file, e.g. like this:

```
<node name="gzserver" pkg="gazebo_ros" type="gzserver" />
```

`gzserver` crashes with 

```
[DEBUG]: remap: __name => gzserver
[DEBUG]: Adding tcp socket [11] to pollset
[DEBUG]: UDPROS server listening on port [58980]
[DEBUG]: Publisher update for [/clock]:  already have these connections: 
[DEBUG]: Started node [/gzserver], pid [30771], bound on [peci1-Latitude-E6540], xmlrpc port [40902], tcpros port [60227], using [sim] time
[ INFO]: Finished loading Gazebo ROS API Plugin.
[DEBUG]: XML-RPC call [lookupService] returned an error (-1): [no provider]
[ INFO]: waitForService: Service [/gazebo/set_physics_properties] has not been advertised, waiting...
[DEBUG]: Accepted connection on socket [11], new socket [14]
[DEBUG]: Adding tcp socket [14] to pollset
[DEBUG]: TCPROS received a connection from [127.0.0.1:45466]
[DEBUG]: Connection: Creating TransportSubscriberLink for topic [/rosout] connected to [callerid=[/rosout] address=[TCPROS connection on port 60227 to [127.0.0.1:45466 on socket 14]]]
[DEBUG]: shutting down due to exit() or end of main() without cleanup of all NodeHandles
gzserver: /usr/include/boost/thread/pthread/recursive_mutex.hpp:101: boost::recursive_mutex::~recursive_mutex(): Assertion `!pthread_mutex_destroy(&m)' failed.
Aborted (core dumped)
```

I found out it is because of the `__name` and `__log` ROS parameters passed to `rosrun`. The commandline parameters actually are:

```
__name:=gzserver __log:=notimportant -s path/devel/lib/libgazebo_ros_paths_plugin.so -s path/devel/lib/libgazebo_ros_api_plugin.so
```

And I've found out that `gzserver` doesn't like the ROS arguments on its commandline (nor does `rosrun` remove them, which I'd expect).

I propose to patch the `gzserver`, `gzclient` and `gazebo` scripts by adding a line like:

```
final=`echo "$final" | sed 's/__[^ ]*:=[^ ]* \?//g'`
```

[...]

PR #390, opened 2016-02-16
[gazebo_ros] Remove ROS remapping arguments from gazebo_ros launch scripts.

peci1:
Solves issue https://github.com/ros-simulation/gazebo_ros_pkgs/issues/387 .

### DIFF
--- a/gazebo_ros/scripts/debug
+++ b/gazebo_ros/scripts/debug
@@ -1,6 +1,10 @@
 #!/bin/sh
 final="$@"
 
+# remove ros remapping arguments like "__name:=node_name"; otherwise, they'd 
+# be interpreted as a world name by gzserver, and it'd crash while loading it
+final=`echo "$final" | sed 's/[^ ]*:=[^ ]* \?//g'`
+
 EXT=so
 if [ $(uname) == "Darwin" ]; then
     EXT=dylib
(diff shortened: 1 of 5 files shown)


j-rivero:
Could you please merge the current jade-devel branch so we get some testing from the travis service? Thanks.

peci1:
Here we go...

gazebo_ros/scripts/debug
```diff
@@ -1,6 +1,10 @@
 #!/bin/sh
 final="$@"
 
+# remove ros remapping arguments like "__name:=node_name"; otherwise, they'd 
+# be interpreted as a world name by gzserver, and it'd crash while loading it
+final=`echo "$final" | sed 's/[^ ]*:=[^ ]* \?//g'`
```
j-rivero, line 6: we probably don't need the echo call, does `sed 's/[^ ]*:=[^ ]* \?//g' <<< $final` works for you?

peci1, line 6: Yes, it works. But I haven't known something like `<<<` exists. Isn't it more readable to do it with the `echo` call?

j-rivero, line 6: > Isn't it more readable to do it with the echo call?

I would say that is a subjective matter but we are saving one extra call and only run grep. minimal minimal peformance impact but better use of tools, IMHO. I leave it up to you.

peci1, line 6: Okay. Then you know my preference :)

j-rivero:
Thanks for the patch!

PR #391, opened 2016-02-16
[gazebo_plugins] Fix row_step of openni_kinect plugin

wkentaro:
@hsu
In addition to #373, this is necessary.
I didn't notice that, sorry about this bug.


v4hn:
:+1: please merge this request, I hit this bug too and spent a whole day debugging it...

PR #392, opened 2016-02-24
openni_kinect: set correct row_step in pointcloudmsg

v4hn:
I was wondering why our robot_self_filter kept segfaulting...
Until I figured out this gazebo plugin sends invalid PointCloud2 messages.


v4hn:
closed as duplicate of #391

PR #393, opened 2016-02-25
Enabling continuous integration with Travis (jade)

j-rivero:
First part of integration with Travis.

Tests are failing although the build is marked as successful. The patches for the testing suite will become in a new pull request. We probably need X support to run the testing suite but at least we can integrate in Travis some static/dynamic analysis tools and it check the compilation with the ros of gazebo and ROS ecosystem.

A new pull request should be directed to the `indigo-devel` branch and new support will be added to test different versions of gazebo.

The testing result of travis in my fork: https://travis-ci.org/j-rivero/gazebo_ros_pkgs/builds/111649828.

Mental note: enable travis integration after merging in the travis console for this repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,55 @@
+sudo: required
+dist: trusty
+language: generic
+compiler:
+  - gcc
+env:
+  global:
+    - CI_ROS_DISTRO=jade
+
+before_install:
+  - export CI_SOURCE_PATH=$(pwd)
+  - export REPOSITORY_NAME=${PWD##*/}
+  - echo "Testing branch $TRAVIS_BRANCH of $REPOSITORY_NAME"
+  # Add ROS repositories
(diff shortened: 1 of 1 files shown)


carlosjoserg:
> Tests are failing although the build is marked as successful. 

Sounds like a [volkswagen](https://github.com/auchenberg/volkswagen) thing, you better start greppin `require('volkswagen')` ;)

The successful build would mean that the command `catkin run_tests -p1 -j4 --summary` successfully ran the test, found that tests are failing and reported so, and finally exits with 0...

This issue might help ros/catkin#576, they suggest to use `catkin_test_results`

scpeters:
We have `catkin_test_results` in the robot_model travis file:
- https://github.com/ros/robot_model/blob/indigo-devel/.travis.yml

j-rivero:
Thanks for the comments, it is failing now as expected.

scpeters:
+1

PR #394, opened 2016-02-25
Enabling continuous integration with Travis (indigo)

j-rivero:
Same that pull request #393 but for Indigo.


scpeters:
~~Looks good.~~ We need more tests though!
~~+1~~

Retracted approval, since there are outstanding comments on #393

PR #400, opened 2016-03-01
Backporting #373 and #391 to indigo-devel

otamachan:
As I'm still using Indigo and #373 and #391 solved my problem, I would appreciate if these changes are backported to `indigo-devel`.
#211 seems to be the same problem on `hydro-devel`.


j-rivero:
Looks good to me, @scpeters ?

otamachan:
Thanks :)

PR #403, opened 2016-03-04
Include gazebo binary package as runtime dependency

j-rivero:
This has the nice effect of fixing the testing suite in current jade-devel.


scpeters:
+1

PR #405, opened 2016-03-10
gazebo_packages/gazebo_ros_joint_pose_trajectory.cpp BUG that caused simulated trajectories to run fast

BenBlumer:
In the Gazebo plugin to run a ros JointTrajectory, the points in the class.points_ member never had their time_from_start set. This caused all the time_from_starts to be set to 0. This in turn caused all the trajectory points to be played instantly one after the other in Gazebo.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_joint_pose_trajectory.cpp
+++ b/gazebo_plugins/src/gazebo_ros_joint_pose_trajectory.cpp
@@ -201,6 +201,8 @@ void GazeboRosJointPoseTrajectory::SetTrajectory(
   for (unsigned int i = 0; i < points_size; ++i)
   {
     this->points_[i].positions.resize(chain_size);
+    this->points_[i].time_from_start = trajectory->points[i].time_from_start;
+
     for (unsigned int j = 0; j < chain_size; ++j)
     {
       this->points_[i].positions[j] = trajectory->points[i].positions[j];
--- a/gazebo_plugins/src/gazebo_ros_joint_pose_trajectory.cpp
+++ b/gazebo_plugins/src/gazebo_ros_joint_pose_trajectory.cpp
@@ -202,7 +202,6 @@ void GazeboRosJointPoseTrajectory::SetTrajectory(
   {
     this->points_[i].positions.resize(chain_size);
     this->points_[i].time_from_start = trajectory->points[i].time_from_start;
-
     for (unsigned int j = 0; j < chain_size; ++j)
     {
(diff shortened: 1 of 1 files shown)


scpeters:
So this PR assumes that the `time_from_start` is in absolute simulation time, but currently it is interpreted as relative time. Does the documentation state one way or the other?

BenBlumer:
Actually -- this keeps it as relative time. `trajectory_start` is still initialized to the time of the message header. The condition for updating to the next joint pose is still `cur_time > trajectory_start` and then within that loop, `trajectory_start` is still updated to `trajectory_start += this->points_[this->trajectory_index].time_from_start`.

BUT: Without this PR, `time_from_start` is never given a value. 

`this->points_` is declared in [the header](https://github.com/BenBlumer/gazebo_ros_pkgs/blob/jade-devel/gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_pose_trajectory.h):
`private: std::vector<trajectory_msgs::JointTrajectoryPoint> points_;`

and then the different points attributes are set in lines 199-208 (of the .cpp file):

```
unsigned int points_size = trajectory->points.size();
  this->points_.resize(points_size);
  for (unsigned int i = 0; i < points_size; ++i)
  {
    this->points_[i].positions.resize(chain_size);
    for (unsigned int j = 0; j < chain_size; ++j)
    {
      this->points_[i].positions[j] = trajectory->points[i].positions[j];
    }
  }

```

But `time_from_start` is never set.

BenBlumer:
Why was this made untargeted?

jackieokay:
We currently have some time pressure on us to make an initial release for ROS Kinetic in 2 days We're sorting the backlog of PRs for this repository into changes that can be easily merged into the initial Kinetic release, and PRs that require a bit more review and attention and probably won't make it into the initial release. We decided this required some additional testing because it could change existing behavior in the package. But we'll be regularly checking the PR backlog and retargeting the existing PRs into subsequent releases.

BenBlumer:
The current plugin doesn't work -- one can quickly grep `this->points_` and see that `time_from_start` is never set (but is referenced). The existing behaviour is to ram through the joint trajectory as quickly as the processor/gpu allows -- this probably should be changed :)

scpeters:
Thanks for the explanation; yes the code does look to be incorrect as is. I think `gazebo_ros_joint_trajectory.cpp` has the same problem. I'll submit a PR to `indigo-devel` as well, since it looks like this is a problem there as well.

scpeters:
+1

Issue #348, opened 2015-08-12
Block laser plugins should produce PointCloud2 messages


bmagyar:
or at least it should be configurable which type to produce so that different sensors can be easily replicated (like velodye).

bmagyar:
I'd assign this to myself if I could :)

ironmig:
@j-rivero, this was fixed

kev-the-dev:
Duplicate of #50 (maintainers would appreciate this contribution so long as it maintains the current behavior as default)

PR #413, opened 2016-03-26
change the output type of  block laser from pointcloud to pointcloud2

libing64:
#348 Since the output of Velodyne 3d lidar is pointcloud2 msg, so I change  the block laser's output to pointcloud2, then it can be used for 3d lidar simulation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,1 @@
+/opt/ros/indigo/share/catkin/cmake/toplevel.cmake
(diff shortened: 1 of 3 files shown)


lucasw:
Maybe it would be good to have both types?  There could be a unit test that creates both and makes sure the points are the same.  

Is there a good standalone unit test possible?   Maybe scan a flat plane and make sure points are on that plane. 

A point_cloud_converter node can do the conversion, though at additional cost over generating the right type from the source.

libing64:
@lucasw 
For the unit test, I will add it later.
For the output type, I just want to keep the code simple. If someone wants to use pointcloud2 type, he could merge this pr to his branch.

scpeters:
Changing the message type is a big change. I agree with @lucasw that we should add a new topic with the pointcloud2 message type so we don't break existing code.

j-rivero:
Thanks for the PR @libing64. I agree with @lucasw and @scpeters, keep already existing code working is a major point for gazebo_ros_pkgs. I'm closing this PR by now, feel free to reopen it or open a new one implementing the idea of having two different topics one for each version of PointCloud.

Issue #409, opened 2016-03-22
[gazebo_ros] More than 1 gzserver rostest results in timeout


lucasw:
I can create rostests that use gzserver that run fine when run individually using rostest or catkin_make run_tests and the individual test name, but trying two tests results in a timeout.

It looks like gazebo isn't getting shut down properly.

Example in  https://github.com/lucasw/gazebo_ros_pkgs/tree/gazebo_ros_tests/gazebo_ros/test

```
$  catkin_make run_tests_gazebo_ros
####
#### Running command: "make run_tests_gazebo_ros -j4 -l4" in "/home/lucasw/catkin_ws/build"
####
-- run_tests.py: execute commands
-- run_tests.py: execute commands
  /opt/ros/jade/share/rostest/cmake/../../../bin/rostest --pkgdir=/home/lucasw/catkin_ws/src/gazebo_ros_pkgs/gazebo_ros --package=gazebo_ros --results-filename test_gazebo_ros.xml --results-base-dir /home/lucasw/catkin_ws/build/test_results /home/lucasw/catkin_ws/src/gazebo_ros_pkgs/gazebo_ros/test/gazebo_ros.test 
  /opt/ros/jade/share/rostest/cmake/../../../bin/rostest --pkgdir=/home/lucasw/catkin_ws/src/gazebo_ros_pkgs/gazebo_ros --package=gazebo_ros --results-filename test_gazebo_ros2.xml --results-base-dir /home/lucasw/catkin_ws/build/test_results /home/lucasw/catkin_ws/src/gazebo_ros_pkgs/gazebo_ros/test/gazebo_ros2.test 
INFO: cannot create a symlink to latest log directory: [Errno 2] No such file or directory: '/home/lucasw/.ros/latest'
... logging to /home/lucasw/.ros/log/rostest-glover-23010.log
... logging to /home/lucasw/.ros/log/rostest-glover-23011.log
[ROSUNIT] Outputting test results to /home/lucasw/catkin_ws/build/test_results/gazebo_ros/rostest-test_gazebo_ros2.xml
[ROSUNIT] Outputting test results to /home/lucasw/catkin_ws/build/test_results/gazebo_ros/rostest-test_gazebo_ros.xml
[ INFO] /gazebo /home/lucasw/catkin_ws/src/gazebo_ros_pkgs/gazebo_ros/src/gazebo_ros_api_plugin.cpp:155 : Finished loading Gazebo ROS API Plugin.
[...]

PR #419, opened 2016-04-09
[gazebo_plugins] Rapid projector toggle rostest

lucasw:
I originally was hoping to duplicate an issue where toggling a projector on and off would crash gazebo, but having made a rostest of it the test passes.  But I think it is a good test and could be developed into more elaborate tests of projectors (adding a camera and some image comparison tests to see that the projector actually projected and toggled on and off would be nice), and serves as an example of launching the projector plugin in a world file (I couldn't find a recent one).

The xacro file will be built into a similar test that uses the xacro -> urdf -> sdf pipeline, it could be left out of the merge until I finish it, or I could finish it first.


gazebo_plugins/CMakeLists.txt
```diff
@@ -339,8 +339,11 @@ install(DIRECTORY test
 # Tests
 if (CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
-  add_rostest_gtest(set_model_state-test
-                    test/set_model_state_test/set_model_state_test.test 
-                    test/set_model_state_test/set_model_state_test.cpp)
-  target_link_libraries(set_model_state-test ${catkin_LIBRARIES})
+  # Can't have two of these until #409 is resolved
```
j-rivero: I think that we are safe to enable all the tests again, #409 was closed by running tests one by one.

j-rivero:
Thanks for the contribution Lucas. Could you please fix the conflicts when you have some time?

ahcorde:
No activity since 2016 and targeting an [unsupported version](http://wiki.ros.org/Distributions). Closing this feel free to reopen again in a supported version if the PR is relevant.

PR #426, opened 2016-04-25
Upgrade to gazebo 7 and remove deprecated driver_base dependency

jackieokay:
the dependency on gazebo7 in package.xml should probably be gazebo instead, and I think there needs to be a rosdep rule for libgazebo7-dev. Or maybe we should make a rosdep rule for libgazebo-dev instead? @wjwwood let me know what you think

This still doesn't compile on Xenial because control_toolbox hasn't been released yet. I'm not sure why the dependency blocker script didn't catch that dependency, maybe the dependency didn't exist in Jade...? Anyway, @bmagyar, just a heads up that gazebo_ros_pkgs actually does depend on control_toolbox, sorry about the confusion. I'm considering the idea of removing the dependency because it only uses control_toolbox for PIDs, and the Gazebo API provides a PID class, but it could be risky.


jackieokay:
update, it turns out that the handy website tool does list control_toolbox as a dependency of gazebo_ros_control--I was just a bit mixed up when I wrote the first post

wjwwood:
If the package will only support Gazebo7 (fine by me), then I think it should use the explicit `libgazebo7-dev` key.

scpeters:
The `driver_base` change looks good, but we need the key name to match the rosdistro pull request https://github.com/ros/rosdistro/pull/11288

bmagyar:
+1 for gazebo 7.

About the PID class... I'm not sure. Naively I'd stick with the current one as that's the PID class that people implementing their own hardware interface will use too on non-simulated platforms. On the flip side, there is this tempting shortcut to just go with Gazebo's PIDs. Maybe it's worth comparing the 2 implementations?

jackieokay:
I've removed a stray reference to `gazebo7` in fc716a0. The rosdistro PR was also updated to have `libgazebo7-dev` key.

jackieokay:

jackieokay:
I cherry picked @wjwwood's commit which ignores gazebo_ros_control into this branch so that we can test with PR testing on the farm until control_toolbox is released.

jackieokay:
Build fails because the rosdistro PR isn't merged yet, d'oh

jackieokay:

jackieokay:
Alrighty, compiling the code succeeded with 66 warnings and 1 failing test:

http://build.ros.org/job/Kpr__gazebo_ros_pkgs__ubuntu_xenial_amd64/9/

@wjwwood thanks for setting up the PR testing!

wjwwood:
Looks like a test failed:

http://build.ros.org/job/Kpr__gazebo_ros_pkgs__ubuntu_xenial_amd64/9/testReport/junit/rostest.runner/RosTest/testset_model_state_test/

And there were warnings: 

http://build.ros.org/job/Kpr__gazebo_ros_pkgs__ubuntu_xenial_amd64/9/warnings17Result/new/

but I don't know if that causes the CI to fail or not...

wjwwood:
Ah, beat me to it 😄, no problem.

wjwwood:
This looks to be the culprit:

From http://build.ros.org/job/Kpr__gazebo_ros_pkgs__ubuntu_xenial_amd64/9/console#console-section-2:

> 00:44:02.413 /tmp/catkin_workspace/src/gazebo_ros_pkgs/gazebo_ros/scripts/gzserver: 26: .: Can't open /usr/share/gazebo//setup.sh

wjwwood:
Is it installing the `gazebo7` deb? On my Xenial machine:

```
william@ubuntu:~$ apt-file search setup.sh | grep gazebo7
gazebo7: /usr/share/gazebo-7/setup.sh
gazebo7: /usr/share/gazebo/setup.sh
```

wjwwood:
I just installed `libgazebo7-dev` and I did not have the `/usr/share/gazebo/setup.sh` file. I do after installing `gazebo7`.

jackieokay:
ok, I mistakenly thought libgazebo7-dev would pull in gazebo7. I'm adding gazebo as a run dependency (rosdep should recognize gazebo as gazebo7 now).

scpeters:
I thought `libgazebo7-dev` would pull in `gazebo7` also

now we know

jackieokay:
The test passes on Jenkins now. Shall I squash and merge this or are there other issues we want to address?

gazebo_ros/scripts/gzserver
```diff
@@ -22,5 +22,5 @@ then
     final="$final -s `catkin_find --first-only libgazebo_ros_api_plugin.$EXT`"
 fi
 
-setup_path=$(pkg-config --variable=prefix gazebo)/share/gazebo/
+setup_path=$(pkg-config --variable=prefix gazebo)/share/gazebo
```
scpeters, line 25: the other scripts have this extra slash, but it's not urgent to change

scpeters:
+1

PR #428, opened 2016-04-25
temporary hack to **fix** the -J joint position option (issue #93), s…

scpeters:
…leeping for 1 second to avoid race condition. this branch should only be used for debugging, merge only as a last resort.


hsu:
closing for now. will try to brainstorm a better solution.

PR #404, opened 2016-03-04
Remove cmake warnings

j-rivero:
remove useless cmake code

PR #436, opened 2016-04-26
Remove useless cmake code in tests

j-rivero:


scpeters:
Why is that cmake code useless? Just curious

j-rivero:
Sorry for the empty the description. Looks to me like the gazebo_msgs_gencpp dependency does not currently exists.  `The dependency target "gazebo_msgs_gencpp" of target "gazebo_ros_bumper" does not exist.` in the CI log.

scpeters:
This looks like a different fix than #404. Should this match that PR?

j-rivero:
let's merge forward indigo and get #404

PR #437, opened 2016-04-26
issue #324 Also accept /world for the frameName parameter in gazebo_r…

hugomatic:


scpeters:
Is this going to be merged to `indigo-devel` as well?

j-rivero:
> Is this going to be merged to indigo-devel as well?

yes. done.

PR #439, opened 2016-04-26
Fix unset variable in gazebo_ros_joint_pose_trajectory

scpeters:
This is a duplicate of #405 that is back-ported to `indigo-devel`. The problem is that the `SetTrajectory(const trajectory_msgs::JointTrajectory)` function didn't copy the `time_from_start` field into the `points_` member variable.


j-rivero:
+1 feel free to merge

PR #441, opened 2016-04-27
Add note about where to write simulation code

j-rivero:
Add text explaining that real simulation code should go into gazebo upstream code.

@scpeters @nkoenig


scpeters:
+1

PR #292, opened 2015-01-29
tricycle drive motorController debugged

todorangrg:
Now the tricycle motor controller behaves as expected, even when acceleration limits are low.

PR #444, opened 2016-04-27
Maxbader indigo devel tricycledrive enchanted

hsu:
attempt at fixing conflicts of #292


gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -227,17 +227,15 @@ void GazeboRosTricycleDrive::motorController ( double target_speed, double targe
...
         if ( fabs ( diff_speed ) < wheel_speed_tolerance_ ) {
-            applied_speed = target_speed;
-        } else if ( diff_speed < target_speed ) {
-            applied_speed = current_speed + wheel_acceleration_ * dt;
-        } else {
-            applied_speed = current_speed - wheel_deceleration_ * dt;
+            applied_speed = current_speed;
+        } else if ( fabs(diff_speed) > wheel_acceleration_ * dt ) {
+            if(diff_speed > 0){ applied_speed = current_speed - wheel_acceleration_ * dt;}
+            else              { applied_speed = current_speed + wheel_deceleration_ * dt;}
```
scpeters: did you want to fix the indentation you commented about in #292 ?

hsu:
I don't quite like using SetVelocity because of unrealistic impulses, or SetPosition because it skips dynamics. But it's an improvement of the state of the plugin.

Ideally, this should be rewritten using SetForce to control steering / drive if the need arises.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
hsu: I fixed the indents, and I reverted to Gazebo style in some places due to lack of self restraint :)
but given things are not formatted uniformly in gazebo_ros_pkgs right now, we should probably deal with stylistic issues in a separate global effort.

scpeters: I was just reminding you of a comment you made. I don't think we should make huge style changes until we declare what the style should be for these packages

scpeters:
+1

PR #445, opened 2016-04-27
fix string replacement to look for mesh filename surrounded by single…

hsu:
… or double quotes. Also do not replace unless there are two slashes after package:

### DIFF
--- a/gazebo_ros/scripts/spawn_model
+++ b/gazebo_ros/scripts/spawn_model
@@ -249,7 +249,7 @@ class SpawnModel():
           sys.exit(0)
 
         if self.package_to_model:
-            model_xml = re.sub("<\s*mesh\s+filename\s*=\s*\"package:/?/?","<mesh filename=\"model://", model_xml)
+            model_xml = re.sub("<\s*mesh\s+filename\s*=\s*([\"|'])package://","<mesh filename=\g<1>model://", model_xml)
 
         # setting initial pose
         initial_pose = Pose()


scpeters:
This is better, but I also realized that the `<mesh>` element can also have a `scale` attribute (see [atlas.urdf](https://bitbucket.org/osrf/drcsim/src/583228eb/atlas_description/urdf/atlas.urdf#atlas.urdf-13) for an example), and I believe this regular expression wouldn't work if the attributes are rearranged.

Here's the different cases I can think of:
- Combinations of quotes with filename first:

```
<mesh filename="package://description/meshes/mesh.dae" scale="1 1 1" />
<mesh filename='package://description/meshes/mesh.dae' scale="1 1 1" />
<mesh filename="package://description/meshes/mesh.dae" scale='1 1 1' />
<mesh filename='package://description/meshes/mesh.dae' scale='1 1 1' />
```
- Combinations of quotes with scale first:

```
<mesh scale="1 1 1" filename="package://description/meshes/mesh.dae" />
<mesh scale="1 1 1" filename='package://description/meshes/mesh.dae' />
<mesh scale='1 1 1' filename="package://description/meshes/mesh.dae" />
<mesh scale='1 1 1' filename='package://description/meshes/mesh.dae' />
```

I can make an issue for this in the interest of making the kinetic release.

scpeters:
I created issue #449 to track the issues and will approve this change since it is an improvement.

scpeters:
+1

Issue #212, opened 2014-06-26
itnode_->advertise called with wrong "latched" argument


dpinol:
GazeboRosCameraUtils::LoadThread makes this call. Last argument should be the latched boolean, and instead a CallbackQueue is passed

```
  this->image_pub_ = this->itnode_->advertise(
    this->image_topic_name_, 2,
    boost::bind(&GazeboRosCameraUtils::ImageConnect, this),
    boost::bind(&GazeboRosCameraUtils::ImageDisconnect, this),
    ros::VoidPtr(), &this->camera_queue_);
```

This is easily spotted compiling with clang:

```
/home/dani/svn/catkin_ws-clang3.5-analyzer/src/gazebo_ros_pkgs/gazebo_plugins/src/gazebo_ros_camera_utils.cpp:307:28: error: 
      address of 'this->camera_queue_' will always evaluate to 'true' [-Werror,-Wpointer-bool-conversion]
    ros::VoidPtr(), &this->camera_queue_);
                     ~~~~~~^~~~~~~~~~~~~
```

jonbinney:
Interesting - so it is also not actually using the queue, right?

dpinol:
I don't have the code in front of me, but I think that for that you have to associate the queue to the node.

j-rivero:
Looking into the API, I would say there is in fact a bug.

What I think that is happening is: the reference to a class member is passed and evaluated as a boolean. Since the reference is not null, the evaluation results into true.

scpeters:
code in [gazebo_plugins/src/gazebo_ros_camera_utils.cpp:319-323](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/c190bf69aa5994351114430c529c5c1dd892b8dc/gazebo_plugins/src/gazebo_ros_camera_utils.cpp#L319-L323)

scpeters:
Hm...the online docs don't show a bool in the API:
- http://docs.ros.org/indigo/api/roscpp/html/structros_1_1AdvertiseOptions.html

scpeters:
https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/include/ros/advertise_options.h#L134-L159

j-rivero:
fixed.

PR #451, opened 2016-04-27
Use proper boolean value for latched in CameraUtils

j-rivero:
Fix for #212. Thanks to Daniel Pinyol.


scpeters:
So you set the value to `true`, which means latching is enabled? Is that preferred for this topic?

scpeters:
I'm not sure this issue is valid (see https://github.com/ros-simulation/gazebo_ros_pkgs/issues/212#issuecomment-215195235 and https://github.com/ros-simulation/gazebo_ros_pkgs/issues/212#issuecomment-215196026 )

j-rivero:
> So you set the value to true, which means latching is enabled? Is that preferred for this topic?

I was trying to preserve the current behaviour but fix the problem with the API call which evaluates to true.

> I'm not sure this issue is valid (see #212 (comment) and #212 (comment) )

Clearly not valid, thanks.

scpeters:
+1

scpeters:
thanks for the explanation

PR #454, opened 2016-04-27
merge indigo, jade to kinetic-devel

scpeters:


scpeters:
good call

j-rivero:
I've been giving a try to the release on my Wily machines and things seems to work or at least it is not completely broken. I think it is safe to merge at this point.

PR #458, opened 2016-05-16
Gazebo_ros_packages - Block Laser Plugin : Fixed negative z-coordinate in point cloud generation

ozaslan:
Hi,

Your code looks like a copy of https://goo.gl/bwtqJk . At line 333 is a negation which used to be valid older Gazebo version. The fix just removes the minus sign.

Thanks

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_block_laser.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_block_laser.h
@@ -47,86 +47,86 @@
 namespace gazebo
 {
 
-	class GazeboRosBlockLaser : public RayPlugin
-	{
-		/// \brief Constructor
-		/// \param parent The parent entity, must be a Model or a Sensor
-		public: GazeboRosBlockLaser();
+  class GazeboRosBlockLaser : public RayPlugin
+  {
+    /// \brief Constructor
+    /// \param parent The parent entity, must be a Model or a Sensor
+    public: GazeboRosBlockLaser();
 
-				/// \brief Destructor
-		public: ~GazeboRosBlockLaser();
+    /// \brief Destructor
+    public: ~GazeboRosBlockLaser();
 
-				/// \brief Load the plugin
-				/// \param take in SDF root element
-		public: void Load(sensors::SensorPtr _parent, sdf::ElementPtr _sdf);
+    /// \brief Load the plugin
+    /// \param take in SDF root element
+    public: void Load(sensors::SensorPtr _parent, sdf::ElementPtr _sdf);
 
-				/// \brief Update the controller
-		protected: virtual void OnNewLaserScans();
+    /// \brief Update the controller
+    protected: virtual void OnNewLaserScans();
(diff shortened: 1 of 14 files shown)


ozaslan:
Hi

I also added a second channel to the Pointcloud message that encodes index of the laser beam with which the corresponding point is captured. This is useful for some of the estimation/mapping algorithms performance of which depend/increase on/with the proposed data structure.  Also, now the coder can modify the update rate of the sensor from within the plugin. This was actually a to-do task that one of the comments in the code writes. Lastly, I added conditionals that choose Gazebo version specific functions in order to silence compiler warnings for deprecated functions.

ozaslan:
Hi again

I also noticed that when multiple sensors are spawned with the plugin, the topic names clash. Using a static variable, the code now keeps track of the number of instances of the class and appends the topic name appropriately to get unique names for each topic.

scpeters:
There are many changes to the whitespace of this file (spaces converted to tabs). Can you minimize these changes so we can see the actual changed code?

ozaslan:
I assume you are using the web interface to see the changes. Can you please try the first trick @ https://github.com/blog/967-github-secrets . Appending the link with "?w=1" discards whitespace changes. From the command line "git diff -w" does the same job.

Thanks

scpeters:
That `?w=1` is a helpful trick.

Large unrelated whitespace changes can cause conflicts with other pull requests and also pollute the line-by-line history in `git blame`, so I request that they are removed before merging pull requests. It's also easier to review after the whitespace has been fixed in case any changes are introduced, so I usually wait the whitespace to be fixed before conducting an extensive review.

Thanks

ozaslan:
I agree with your arguments. I tried to clean out the whitespace-tab problem. 

Thanks

j-rivero:
Thanks for the contribution @ozaslan.

Looks to me like your pull request cover three different things:
- Fixes the problem with the Z axis
- Fixes the problem of having more than one plugin
- Add the ability to change the update rate of the sensor from within the plugin.

Amazing! It would be possible to split the modifications in three different pull requests? It will be easier to review, proper changelogs for bloom, etc.

j-rivero:
> Amazing! It would be possible to split the modifications in three different pull requests? It will be easier to review, proper changelogs for bloom, etc.

@ozaslan if you don't have time for this, it is fine. I will take care of the changelogs. Could you please resolve the conflicts? Thanks.

ozaslan:
@j-rivero , sorry for the delay. These days I am a bit busy. But I will resolve the conflicts the soonest.

j-rivero:
> @j-rivero , sorry for the delay. These days I am a bit busy. But I will resolve the conflicts the soonest.

No rush, take all the time you need. I just wanted to be sure that my recommendations of splitting the pull request are not a barrier for you.
Thanks.

kev-the-dev:
I'm closing this as jade is EOL. Also, the original issue this PR solves was fixed in #582. Feel free to PR to a supported distro if you think the other features would still be useful

PR #459, opened 2016-06-01
launch scripts override GAZEBO_MASTER_URI

peci1:
GAZEBO_MASTER_URI is always the one written in `setup.sh`, even if a different value is set in the user's environment. 

I consider it counter-intuitive, if you e.g. run `GAZEBO_MASTER_URI=http://myserver:11345 rosrun gazebo_ros gzserver`, that the server is still started with the default URI.

### DIFF
--- a/gazebo_ros/scripts/gazebo
+++ b/gazebo_ros/scripts/gazebo
@@ -21,11 +21,17 @@ then
 fi
 
 client_final="-g `catkin_find --first-only libgazebo_ros_paths_plugin.$EXT`"
+
+setup_path=$(pkg-config --variable=prefix gazebo)/share/gazebo/
+
+# source setup.sh, but keep local modifications to GAZEBO_MASTER_URI
 desired_master_uri="$GAZEBO_MASTER_URI"
+. $setup_path/setup.sh
+if [ "$desired_master_uri" = "" ]; then
+	desired_master_uri="$GAZEBO_MASTER_URI"
+fi
 
 # Combine the commands
-setup_path=$(pkg-config --variable=prefix gazebo)/share/gazebo/
-. $setup_path/setup.sh 
 GAZEBO_MASTER_URI="$desired_master_uri" gzserver $final & 
 GAZEBO_MASTER_URI="$desired_master_uri" gzclient $client_final
 
--- a/gazebo_ros/scripts/gazebo
+++ b/gazebo_ros/scripts/gazebo
@@ -21,12 +21,13 @@ then
 fi
 
 client_final="-g `catkin_find --first-only libgazebo_ros_paths_plugin.$EXT`"
+desired_master_uri="$GAZEBO_MASTER_URI"
 
 # Combine the commands
 setup_path=$(pkg-config --variable=prefix gazebo)/share/gazebo/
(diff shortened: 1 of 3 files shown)


j-rivero:
I agree with the goal.

I wonder if we can remove the source of setup.sh completely. There are [default values at compile time](https://bitbucket.org/osrf/gazebo/pull-requests/620/add-compile-time-defaults-for-env-vars/diff) since gazebo 1.9. The files affected in this PR are probably older.

@scpeters @hsu @nkoenig what do you think?

scpeters:
I think we still need to source setup.sh since the default values are only used if the `GAZEBO_*` environment variables are unset.

Example, we have a model that uses a gazebo plugin (say `libLiftDragPlugin.so`) provided by upstream gazebo and also a gazebo ros plugin from the `gazebo_plugins` package (say `libgazebo_ros_imu.so`).

If `GAZEBO_PLUGIN_PATH` is unset, then it will default to `/usr/lib/arch/gazebo-7/plugins/` and `libLiftDragPlugin.so` will be findable, but not `libgazebo_ros_imu.so`.

If `GAZEBO_PLUGIN_PATH` is set to `/opt/ros/distro/lib`, it will be able to find `libgazebo_ros_imu.so` but not `libLiftDragPlugin.so`.

If `GAZEBO_PLUGIN_PATH` is set to `/opt/ros/distro/lib` and then the setup.sh is sourced, it will prepend `/usr/lib/arch/gazebo-7/plugins/`, so that both plugins are findable.

It would be nice to get rid of the setup.sh, but I think it's still necessary.

scpeters:
Looks good, though I haven't tested it

peci1:
Hmm... now I see a possible problem: if the user runs these scripts without GAZEBO_MASTER_URI set beforehand, it might cause a situation when the server/client is run with an empty GAZEBO_MASTER_URI... Am I right? So should we also add a check and only overwrite the URI when it is set and nonempty? What do you think?

peci1:
What about this?

I'd also like to prefix the last line calling `gazebo`/`gzserver`/`gzclient` with `LC_ALL=C`, since if you use a locale, where `,` is the decimal separator, SDF parsing fails. Could I add it here, or should I open a new PR for that?

And last thing: looking into `setup.sh`, I see `GAZEBO_MODEL_DATABASE_URI` suffers the same problem as `GAZEBO_MASTER_URI`. Should I take care of it the same way here?

j-rivero:
> What about this?

Looks good to me, now is working.

> I'd also like to prefix the last line calling gazebo/gzserver/gzclient with LC_ALL=C, since if you use a locale, where , is the decimal separator, SDF parsing fails. Could I add it here, or should I open a new PR for that?

That [bug](https://bitbucket.org/osrf/sdformat/issues/60/error-when-starting-gazebo-if-lc_numeric) should be fixed since sdformat 2.3.2 so I prefer not mess with the LC_ALL variable if possible. Which version are you using?

> And last thing: looking into setup.sh, I see GAZEBO_MODEL_DATABASE_URI suffers the same problem as GAZEBO_MASTER_URI. Should I take care of it the same way here?

I would not change it unless we found a real use case for it.

Thanks!

PR #460, opened 2016-06-06
[gazebo_plugins] bugfix: duplicated tf prefix resolution in gazebo_ros_camera plugin

furushchev:
In line 280-287 in `gazebo_ros_camera_utils.cpp`, there are codes that resolve tf prefix which I think is duplicated code of line 290-296.
When the spawning node is launched at `group` tag (for example `group ns="robot1"`) in launch file, and `tf_prefix` ros param is set, (for example `tf_prefix` is also set as "robot1"), the frame of `camera_info`'s header is published as `/robot1/robot1/camera_rgb_frame` which is resolved tf twice, though true tf frame is `/robot1/camera_rgb_frame`)

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_camera_utils.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera_utils.cpp
@@ -277,14 +277,6 @@ void GazeboRosCameraUtils::LoadThread()
   this->camera_info_manager_.reset(new camera_info_manager::CameraInfoManager(
           *this->rosnode_, this->camera_name_));
 
-  // resolve tf prefix
-  std::string key;
-  if(this->rosnode_->searchParam("tf_prefix", key)){
-    std::string prefix;
-    this->rosnode_->getParam(key, prefix);
-    this->frame_name_ = tf::resolve(prefix, this->frame_name_);
-  }
-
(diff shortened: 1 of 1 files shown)


j-rivero:
Looks good to me, both pieces of code (using different functions) are doing the same. My only concern would be to break current working code out there.

Options: 
- Go for the change as it is now and write the change clearly in the Changelog (or add a new Migration.md file)
- Create a new class parameter (`_remove_duplicate_tf_namespace`)  to enable this fix, off by default. Migrations are sometimes easier this way (we can enable it by default in the next ROS releases).

furushchev:
@j-rivero Thank you for reviewing. I'd like to add this change to CHANGELOG, and go for this change as it is now, because current code is not convenient on most situation.
Can I append to CHANGELOG.rst directly about this change or only comment to this page (because changelog includes versioning and I dont know about versioning of this repository)

furushchev:
@j-rivero How about my previous comment?

furushchev:
@j-rivero ping

j-rivero:
Thanks for the interest.

I've gave a run with the turtlebot all seems to keep working.

furushchev:
@j-rivero Thank you very much for merging!

PR #461, opened 2016-06-13
Fix gazebo7 deprecation warnings

nkoenig:


hsu:
+1

scpeters:
i'm testing it now

PeterMitrano:
+1
Is the `config.h` needed in gazebo_plugins/src/gazebo_ros_gpu_laser.cpp? Looks like you made a commit to remove some but added it there

gazebo_plugins/src/gazebo_ros_gpu_laser.cpp
```diff
@@ -34,6 +34,7 @@
 #include <gazebo/sensors/GpuRaySensor.hh>
 #include <gazebo/sensors/SensorTypes.hh>
 #include <gazebo/transport/transport.hh>
+#include <gazebo/gazebo_config.h>
```
PeterMitrano, line 37: this is the one I'm talking about

scpeters, line 37: it might not need it because it should also be included by `gazebo_ros_utils.h`, but I think it doesn't hurt to have it there

scpeters:
+1

gazebo_plugins/src/gazebo_ros_gpu_laser.cpp
PeterMitrano, line 37: yea I just noticed this one was explicitly added so I thought I'd mention it

chapulina:
~~Why is this targeted at indigo if that's supposed to match gazebo 2?~~ -- Nevermind, just saw the ifs

PeterMitrano:
could be because people like me build from source and want to use 7? Idk, I assumed that's why...

chapulina:
+1

nkoenig:
It's targeted to indigo-devel so that people can build from source, and also because we release a debian for indigo and gazebo7.

PR #464, opened 2016-06-16
[gazebo_plugins] bugfix: duplicated tf prefix resolution in gazebo_ros_camera plugin (kinetic-devel)

j-rivero:
{ port of pull request #460 }
In line 280-287 in `gazebo_ros_camera_utils.cpp`, there are codes that resolve tf prefix which I think is duplicated code of line 290-296.
When the spawning node is launched at `group` tag (for example `group ns="robot1"`) in launch file, and `tf_prefix` ros param is set, (for example `tf_prefix` is also set as "robot1"), the frame of `camera_info`'s header is published as `/robot1/robot1/camera_rgb_frame` which is resolved tf twice, though true tf frame is `/robot1/camera_rgb_frame`)


j-rivero:
ouch, erroneous branch.

PR #435, opened 2016-04-25
Replacement for #303

caguero:
#303

PR #480, opened 2016-08-05
Delete spawned model when rospy shutdown

lgeo3:
Usefull when spawn model script is used in a launch file. Each time the
launch is stoped (i.e ctrl-c) the spawned model is removed from gazebo.
It allows to relaunch multiple time the same launch file which occurs
frequently when updating urdf model.

### DIFF
--- a/gazebo_ros/scripts/spawn_model
+++ b/gazebo_ros/scripts/spawn_model
@@ -199,7 +199,6 @@ class SpawnModel():
     def callSpawnService(self):
 
         # wait for model to exist
-        rospy.init_node('spawn_model')
 
         if not self.wait_for_model == "":
           rospy.Subscriber("%s/model_states"%(self.gazebo_namespace), ModelStates, self.checkForModel)
@@ -291,6 +290,15 @@ class SpawnModel():
 
         return
 
+    def deleteSpawnedModel(self):
+        """
+        Remove the model spawned
+        """
+        delete_model_service_name = '/'.join([self.gazebo_namespace, 'delete_model'])
+        delete_model = rospy.ServiceProxy(delete_model_service_name, gazebo_interface.DeleteModel)
+        rospy.loginfo("Calling service %s" % delete_model_service_name)
+        resp = delete_model(model_name=self.model_name)
+
 if __name__ == "__main__":
     if len(sys.argv) < 2:
         print usage()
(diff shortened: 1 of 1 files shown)


j-rivero:
Hello Laurent, thanks for the PR. Let me fully understand which is the problem, could you please describe which is the use case that this PR will fix? Or in other words, what are you trying to do that you can not do without this PR.

Thanks!

lgeo3:
Sorry I send the pull request in a hurry and didn't explain the use case.

During development of a new robot, I follow an iterative approach to configure the robot URDF file for simulation. It consists mainly in setting some gazebo parameters in the URDF (e.g. configuring frictions, max effort etc) and testing if the simulation seems correct.  I usually modify the URDF, stop simulation, start it again and re-spawn the robot in gazebo. This work-flow is a bit heavy.

One way to reduced this work-flow, is to keep the same gazebo simulation running and only "removing" the robot before re-spawning it with the new URDF parameters.

My first try was to manually remove the robot using: `rosservice call /gazebo/delete_model "model_name: "'my_robot_name'"`. But as I usually use a launch file to spawn the robot in gazebo, a more automatic way seems possible: the spawned robot model could be removed if the launch file receive a SIGINT  (e.g with a ctrl-c). 

Thus, my second try was to make the spawn_model node to stay active in background until a SIGINT is received, and if a SIGINT is received remove the robot model automatically. That is the purpose of this pull request. The workflow is less heavy and becomes:

```
launch gazebo
while (not(is_simulation_ok())):
    edit the robot urdf
    launch roslaunch file to spawn the robot # this send a ctrl-c to previously launch node)
```

The launch file that I use is like this:

```
<param name="robot_description" command="$(find xacro)/xacro.py '$(find stan_description)/urdf/stan_v2.xacro' --inorder in_simulation:=true" />
<node name="mybot_spawn" pkg="gazebo_ros" type="spawn_model" args="-urdf -param robot_description -model my_robot_name -z 0.0001 -x 37.90 -y 27.26 -Y -0.075" />
<node .. others nodes ... />
```

kev-the-dev (changes requested):
Thanks for the contribution! This changes the default behavior of the script and would be disruptive to current users. I think this would be more appropriate for a new script, perhaps ```delete_model```, which could have a flag ```-on-exit``` to have this delete when shutdown behavior.

kev-the-dev:
Actually this was implemented in #435 (this is a duplicate). @scpeters can we close this?

PR #482, opened 2016-08-17
[gazebo_plugins] Fix vacuum gripper plugin to work in stable

wkentaro:
In order to make the vacuum gripper plugin work in stable.
Without last commit, 4c13f24, the object will have too fast angular velocity unexpectedly.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
+++ b/gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
@@ -217,18 +217,19 @@ void GazeboRosVacuumGripper::UpdateChild()
         links[j]->SetAngularAccel(link_->GetWorldAngularAccel());
         links[j]->SetLinearVel(link_->GetWorldLinearVel());
         links[j]->SetAngularVel(link_->GetWorldAngularVel());
-        double norm_force = 1 / norm;
         if (norm <= min_distance_) {
           // apply friction like force
           // TODO(unknown): should apply friction actually
           link_pose.Set(parent_pose.pos, link_pose.rot);
           links[j]->SetWorldPose(link_pose);
+        } else {
+          double norm_force = 1 / norm;
+          if (norm_force > max_force_) {
+            norm_force = max_force_;
+          }
+          math::Vector3 force = norm_force * diff.pos.Normalize();
+          links[j]->AddForce(force);
         }
-        if (norm_force > max_force_) {
-          norm_force = max_force_;
-        }
-        math::Vector3 force = norm_force * diff.pos.Normalize();
-        links[j]->AddForce(force);
         grasping_msg.data = true;
       }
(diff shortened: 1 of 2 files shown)


j-rivero:
Thanks for the PR Kentaro. I think that we don't have an example of world in the repo which uses the vacuum_gripper. Would you mind to attach one to this PR or write some instructions for setting up a world with one?

gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
```diff
@@ -192,23 +212,24 @@ void GazeboRosVacuumGripper::UpdateChild()
       math::Pose link_pose = links[j]->GetWorldPose();
       math::Pose diff = parent_pose - link_pose;
       double norm = diff.pos.GetLength();
-      if (norm < 0.05) {
+      if (norm <= max_distance_) {
         links[j]->SetLinearAccel(link_->GetWorldLinearAccel());
-        links[j]->SetAngularAccel(link_->GetWorldAngularAccel());
+        links[j]->SetAngularAccel(math::Vector3(0, 0, 0));
```
j-rivero, line 217: I'm not sure if ignoring totally the angular acceleration and set it to 0 is the best option here. @scpeters , @hsu could you comment on this change please?

oroulet:
Why is this not merged? it seems rather obvious? hardcoding the search distance was a bad idea.

j-rivero:
> Why is this not merged? it seems rather obvious? hardcoding the search distance was a bad idea.

I agree. My main concern is the change in [the angular acceleration](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/482#discussion_r78857804) that could affect current simulations.

varunvp:
Has this been merged? Any help required?

gazebo_plugins/src/gazebo_ros_vacuum_gripper.cpp
```diff
@@ -110,6 +110,26 @@ void GazeboRosVacuumGripper::Load(physics::ModelPtr _model, sdf::ElementPtr _sdf
...
+    max_force_ = 20;
+  }
+  // max distance to apply force: [m]
+  if (_sdf->HasElement("maxDistance")) {
+    max_distance_ = _sdf->GetElement("maxDistance")->Get<double>();
+  } else {
+    max_distance_ = 0.05;
+  }
+  // distance to apply friction: [m]
+  if (_sdf->HasElement("minDistance")) {
```
chapulina, line 127: I think `minDistance` is a misleading name for this parameter, because it sounds like anything below this distance will not be affected, which is not true.

ahcorde:
targeting an [unsupported version](http://wiki.ros.org/Distributions). Closing this, feel free to reopen again in a supported version if the PR is still relevant.

PR #493, opened 2016-09-19
Add an argument to enable required flag on gazebo gui

vincentrou:
This change will enable the ability to close all the nodes launched with gazebo when you close the gazebo gui window.
This is  a more convenient way to close the simulation, in comparison to ctrl+c in the right terminal.

### DIFF
--- a/gazebo_ros/launch/empty_world.launch
+++ b/gazebo_ros/launch/empty_world.launch
@@ -13,6 +13,7 @@
   <arg name="respawn_gazebo" default="false"/>
   <arg name="use_clock_frequency" default="false"/>
   <arg name="pub_clock_frequency" default="100"/>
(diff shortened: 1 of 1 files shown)


j-rivero:
Thanks for the contribution Vincent.

I'm not entirely sure that `required_gui`is the best name that we can give to the option. When I read this PR for the first time I thought that something related to the gazebo GUI was involved in the change when it really affects the way of ROS launch is working, if I'm not wrong. How about leave it just as `required` to match the same concept that in the rest of ROS?

If you have time, please add to the PR description the use case that you found that requires this change.

vincentrou:
You are right about the name, I have changed it.

j-rivero:
> You are right about the name, I have changed it.

Thanks. Looks good to me.

PR #496, opened 2016-10-06
Fixed getLinkState service's angular velocity return - indigo

shamwell:
A typo had the getLinkState service returning angular Vx for both Vx and Vz.


j-rivero:
Thanks for the contribution, looks like the proper bugfix to me. Accepted  the change in Indigo.

PR #497, opened 2016-10-15
gazebo_ros_camera_utils: add try/catch

scpeters:
The image_transport advertise call for image_raw was throwing exceptions quite often. This catches those exceptions and gives a `ROS_WARN`.

Room for improvement:
- catch exceptions from other `advertise` calls
- should this be `ROS_ERROR`?

This may be improved by https://github.com/ros/class_loader/pull/40


j-rivero:
> The image_transport advertise call for image_raw was throwing exceptions quite often.

Let me ask: how is that affecting the user experience? how critical are the exceptions?

scpeters:
this is very old, and I'm not sure if it's still an issue, so I'm going to close it

PR #499, opened 2016-10-25
Added a more proper way of interpreting the "package://" URI scheme in SDF files.

peci1:
Requires https://bitbucket.org/osrf/gazebo/pull-requests/2479 .

### DIFF
--- a/gazebo_ros/src/gazebo_ros_paths_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_paths_plugin.cpp
@@ -27,11 +27,59 @@
 
 #include <map>
 
-namespace gazebo
-{
+namespace gazebo {
 
 typedef std::vector<std::string> V_string;
 typedef std::map<std::string, std::string> M_string;
+
+class ROSPackageURISchemeResolver : public common::URISchemeResolver
+{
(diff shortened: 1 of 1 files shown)


j-rivero:
Thanks Martin for the bridge to the change upstream. I'm afraid that the new change in Gazebo upstream would be released in gazebo8 which will be supported by ROS Lunar or eventually by a `ros-kinetic-gazebo8-*` wrappers.

We would need ifdef checks to be sure that changes affects only the gazebo8 versions. I'm going to decline the PR, feel free to open a new one with the ifdefs against `kinetic-devel`. I set the `ros-lunar` tag so we don't forget to review it when the new release is out.

peci1:
For future review: Things got more complicated on the way and the PR was in the end implemented in ign-common package (https://bitbucket.org/ignitionrobotics/ign-common/pull-requests/129/add-uri-resolvers2/). The changes got merged into that package, but gazebo itself will probably start using it in version 10 or 11.

chapulina:
Note that Gazebo 10 has already been released, and it has an optional dependency on `ign-common1`, which doesn't have the API added in that PR.

PR #503, opened 2016-11-03
Fix distortion coefficients order

efernandez:
It should be D = {k1, k2, p1, p2, k3}, according to:
- sensor_msgs/CameraInfo:
  http://docs.ros.org/api/sensor_msgs/html/msg/CameraInfo.html
- OpenCV:
  http://docs.opencv.org/2.4/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_camera_utils.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera_utils.cpp
@@ -524,11 +524,14 @@ void GazeboRosCameraUtils::Init()
   camera_info_msg.distortion_model = "plumb_bob";
   camera_info_msg.D.resize(5);
 #endif
+  // D = {k1, k2, t1, t2, k3}, as specified in:
+  // - sensor_msgs/CameraInfo: http://docs.ros.org/api/sensor_msgs/html/msg/CameraInfo.html
+  // - OpenCV: http://docs.opencv.org/2.4/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html
   camera_info_msg.D[0] = this->distortion_k1_;
   camera_info_msg.D[1] = this->distortion_k2_;
-  camera_info_msg.D[2] = this->distortion_k3_;
-  camera_info_msg.D[3] = this->distortion_t1_;
-  camera_info_msg.D[4] = this->distortion_t2_;
+  camera_info_msg.D[2] = this->distortion_t1_;
+  camera_info_msg.D[3] = this->distortion_t2_;
+  camera_info_msg.D[4] = this->distortion_k3_;
   // original camera_ matrix
   camera_info_msg.K[0] = this->focal_length_;
   camera_info_msg.K[1] = 0.0;


efernandez:
ping @j-rivero 

Let me know if I should target `jade-devel` instead, or `kinetic-devel`, but I think the fix should be cherry-picked to all branches/distros.

efernandez:
RAW image (image_raw) to test `indigo-devel` (note that I need to stop/start gazebo to test with the fix here later):

camera_distortion_image_raw

Undistorted image (image_rect) on `indigo-devel`:

camera_distortion_image_rect

RAW image (image_raw) to test this fix:

camera_distortion_raw_fix

Undistorted image (image_rect) with this fix:

camera_distortion_image_rect_fix

Although it's clear to me that the assignment of the distortion coefficients to `D` is wrong, the result with the fix still doesn't provide a perfect undistorted image. Maybe gazebo is using a different model for the distortion ?

I can also try with greater values for the misplaced values: `k3`, `p1` and `p2`, to see if it's more evident the fix is doing something.

efernandez:
With these values the fix is clearly doing better:
- k3 = 0.1
- p1 = 0.00691
- p2 = -0.00471

With `indigo-devel`:

before_fix

With this fix:

after_fix

lucasw:
Is this no longer true?

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_plugins/src/gazebo_ros_camera_utils.cpp#L242 says:

```
ROS_WARN("gazebo_ros_camera_ simulation does not support non-zero"
         " distortion parameters right now, your simulation maybe wrong.");
```

I think indigo is packaged by default with a version of gazebo that doesn't do distortion, the gazebo changelog says it was added in 5.0.0.

You could progressively zero out the higher order parameters and see if the lower order ones are being handled properly- the rectified output should look straighter than in your example images if everything is being done correctly.

Also, it's worth finding out if https://bitbucket.org/osrf/gazebo/src/ec2881681a3e1b075cf512f83ea39ad2c356cdce/gazebo/rendering/Distortion.cc?at=default&fileviewer=file-view-default is actually implementing the inverse of the image_proc undistort.

(I wrote an opencv distort a while ago, and I've tested it to show that it does do the inverse of the the image_proc undistort: https://github.com/lucasw/vimjay/blob/master/src/standalone/cv_distort_image.cpp)

efernandez:
Thanks for your comments @lucasw . I'd look into this during the weekend.

I've been using Gazebo 7.4.0 BTW.

efernandez:
BTW, I have the hypothesis that Gazebo is doing the horizontal distortion well, but not the vertical one. I want to check with a square camera image to see how the results are.

Will also look image_proc/OpenCV is doing the inverse of Gazebo.

iche033:
The distortion coefficient ordering fix looks good to me

gazebo's [distortion model](https://bitbucket.org/osrf/gazebo/src/default/gazebo/rendering/Distortion.cc?fileviewer=file-view-default#Distortion.cc-261) is based on the [equations here](http://en.wikipedia.org/wiki/Distortion_%28optics%29#Software_correction) and I believe it should be the same as the ones used in opencv. 

One thing that gazebo does is it tries to crop the black borders for the distorted images and (thus changing the resulting image a little) I wonder if that's the problem. You can replace `/usr/share/gazebo-7/media/materials/programs/camera_distortion_map_fs.glsl` with [this one](https://gist.github.com/iche033/55bc0b64a5ab01baae26fb2e6fb78b0e) to output the original distorted image with borders and see if undistortion works better.

Testing with a square camera image is a good idea. Another test would be to just directly use opencv (bypassing gazebo_ros_pkgs) to undistort a still image generated by gazebo and see if the results are any different.

j-rivero:
Interesting discussion, thanks for the patch and the views on the topic. In my understanding of the topic with the help of Ian, we have two different things:
- The patch seems correct and can be merged.
- A discussion about  how gazebo is implementing the distortion and if it is correctly implemented or not.

For the first point I would like to consider what would be the impact for Indigo users. I try to maintain the changes affecting ROS Indigo (even if sometimes if clearly not ideal) at a minimun level so we gain stability there. Do you think that the change could impact highly in code already written consuming the image distortion?. I'm going to open the PRs for Jade and Kinetic now and merge them.

iche033:
I think it would be nice to get in. It'll affect indigo users relying on these distortion coefficients to undistort camera images. It's possible that people already have a workaround in place to swap these coefficients so they'll need to remove the workaround.

efernandez:
Sorry I haven't had time to look further into this.

I remember I tried the square camera, but it didn't work as I expected, so I think there's something else going on. This PR as is won't be enough. I mean, even if we set the distortion coefficients in order, the distortion applied in Gazebo won't be undone properly from OpenCV.

I'd try to go back into this in the future and update this with my actual findings, which I don't have at hand right now.

Kukanani:
I can confirm that the bad undistortion @efernandez is seeing is due to the border crop on barrel distortion cameras. In #572 I have added the ability to disable this border cropping, making the distorted images correctly match the parameters passed to the camera plugin. 

I'd like to request that this PR be merged, and the relevant discussion on distortion be moved to PR #572 .

efernandez:
Yes, I also think this should be merged.

Rest of the discussion can continue there in #572 and hopefully someone makes the distortion work well. Unfortunately, I'm busy now and I can help with that.

I wonder why the indigo and jade tests had failed ?!
Are they legitimate failures?

At least this one is not: https://build.osrfoundation.org/job/ros_gazebo7_pkgs-ci-pr_any_indigo-trusty-amd64/1/consoleFull#815664435e1461dac-930a-4f4d-96c7-1c0beb22914c

It's failing with apt-get dependencies regarding gazebo2, when the test name says gazebo7:

``` bash
The following packages have unmet dependencies:
 gazebo2 : Depends: libsdformat-dev (>= 1.4.11-1osrf1) but it is not going to be installed
           Depends: libsdformat-dev (< 2.0.0) but it is not going to be installed
W: Duplicate sources.list entry http://archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages (/var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_trusty-updates_main_binary-amd64_Packages.gz)
W: Duplicate sources.list entry http://archive.ubuntu.com/ubuntu/ trusty-updates/restricted amd64 Packages (/var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_trusty-updates_restricted_binary-amd64_Packages.gz)
W: Duplicate sources.list entry http://archive.ubuntu.com/ubuntu/ trusty-updates/universe amd64 Packages (/var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_trusty-updates_universe_binary-amd64_Packages.gz)
W: You may want to run apt-get update to correct these problems
E: Unable to correct problems, you have held broken packages.
ERROR: the following rosdeps failed to install
  apt: command [apt-get install -y gazebo2] failed
Build step 'Execute shell' marked build as failure
Recording test results
```

Can someone fix it and re-run the tests, please?

j-rivero:
> I wonder why the indigo and jade tests had failed ?!
> Are they legitimate failures?
> 
> At least this one is not: https://build.osrfoundation.org/job/ros_gazebo7_pkgs-ci-pr_any_indigo-trusty-amd64/1/consoleFull#815664435e1461dac-930a-4f4d-96c7-1c0beb22914c
> 
> It's failing with apt-get dependencies regarding gazebo2, when the test name says gazebo7:

Ouch, that is a problem in the testing infrastructure, I will fix it. I think that we are safe to merge this now.

130s:
@efernandez Off-topic though, how do you spawn AR tags on Gazebo?
I'm preparing a PR against https://github.com/mikaelarguedas/gazebo_models where some tags are available.

efernandez:
@130s I wasn't satisfied with anything available publicly, so I did my own gazebo models. It's a private package, but feel free to DM me and maybe we can open source it.

It spawns any tag from all the pre-generated tag families in https://april.eecs.umich.edu/software/apriltag.html .

I use textured planes, but I'd like to switched to textured boxes, because planes can't be moved in Gazebo :disappointed: 
I use SDF, but I'd also like to have URDF versions I can attach to a robot model. I can't do that with an SDF.

If you support any of these two things, I'd like to talk with you. Same if you'd like to look into them, and that would make it easier to open source my package.

nicolaerosia:
@efernandez try https://github.com/andreasBihlmaier/gazebo2rviz and https://github.com/andreasBihlmaier/pysdf, I've used it to generate urdf for rviz

PR #516, opened 2016-12-01
Workaround to support gazebo and ROS arguments in the command line

### LINKED ISSUE #387: Support to use gazebo and ROS parameters when using rosrun

j-rivero:
Reorder command line arguments to place ROS remappings at the end so gazebo passed them to be handle by gazebo ROS plugins. While this is not the recommended way of using rosrun, it could be useful for
some use cases.

Should fix #387

### DIFF
--- a/gazebo_ros/scripts/libcommon.sh
+++ b/gazebo_ros/scripts/libcommon.sh
@@ -1,14 +1,14 @@
 #!/bin/sh
 
-# the function relocate all the remappings in the command at the end of the string
-# this allows some punky uses of rosrun, for more information see:
+# the function relocates all the ROS remappings in the command at the end of the
+# string this allows some punky uses of rosrun, for more information see:
 # https://github.com/ros-simulation/gazebo_ros_pkgs/issues/387
 relocate_remappings()
 {
   command_line=${1}
   
   for w in $command_line; do
-    if $(echo $w | grep ':='> /dev/null); then
+    if $(echo $w | grep -q ':='); then
       ros_remaps="$ros_remaps $w"
     else
       gazebo_args="$gazebo_args $w"
--- a/gazebo_ros/scripts/libcommon.sh
+++ b/gazebo_ros/scripts/libcommon.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# the function relocate all the remappings in the command at the end of the string
+# this allows some punky uses of rosrun, for more information see:
+# https://github.com/ros-simulation/gazebo_ros_pkgs/issues/387
+relocate_remappings()
+{
+  command_line=${1}
+  
+  for w in $command_line; do
+    if $(echo $w | grep ':='> /dev/null); then
+      ros_remaps="$ros_remaps $w"
+    else
+      gazebo_args="$gazebo_args $w"
+    fi
+  done
+
+  echo $gazebo_args$ros_remaps | cut -c 1-
+}
--- a/gazebo_ros/CMakeLists.txt
+++ b/gazebo_ros/CMakeLists.txt
@@ -51,6 +51,7 @@ catkin_package(
   std_msgs
   gazebo_msgs
 
+
   DEPENDS
     TinyXML
 )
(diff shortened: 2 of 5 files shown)


peci1:
The changes in CMakeLists seem to contain lots of whitespace changes. Are these intentional?

gazebo_ros/scripts/libcommon.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# the function relocate all the remappings in the command at the end of the string
+# this allows some punky uses of rosrun, for more information see:
+# https://github.com/ros-simulation/gazebo_ros_pkgs/issues/387
+relocate_remappings()
+{
+  command_line=${1}
+  
+  for w in $command_line; do
+    if $(echo $w | grep ':='> /dev/null); then
```
peci1: What about `grep -q ':='` instead of `grep ':=' > /dev/null`?

peci1:
I've tested my use case with remapping before the world file and it works correctly: Gazebo loads the intended world file and ROS receives the remappings.

gazebo_ros/scripts/libcommon.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+# the function relocate all the remappings in the command at the end of the string
```
peci1: "the function relocate" => "the function relocates"
"all the remappings" => "all the ROS remappings"

gazebo_ros/scripts/libcommon.sh
j-rivero: better, c1681bc009d94cec96bf173adfd2a3d2510c7e5e

gazebo_ros/scripts/libcommon.sh
j-rivero: fixed c1681bc009d94cec96bf173adfd2a3d2510c7e5e

jonbinney:
I tested on the simulated launchfile that I use for my robot, and it works for me!

j-rivero:
Perfect. Thanks very much for all the ideas and feedback. Let's merge and release. It will take until the next sync to bump to the stable repository.

PR #517, opened 2016-12-04
Change boost::shared_ptr to urdf::JointConstSharedPtr

jspricke:


j-rivero:
Hello Jochen:

The change break current Ubuntu builds, we probably need to check for the urdfdom version in order to change the type of pointers. Which platform are you using to test the `gazebo_ros_pkgs` build? Which version of  `urdfdom-headers`?

Thanks.

jspricke:
Sorry, wrong branch.

PR #519, opened 2016-12-06
Change boost::shared_ptr to urdf::JointConstSharedPtr

jspricke:


j-rivero:
Thanks Jochen, looks good to me.

PR #532, opened 2017-02-06
[gazebo_plugins] Ensure gazebo_ros_camera Publishes After Reset

kennysharma:
After a simulation reset, gazebo_ros_camera does not publish images because it is using its internal (and not reset by the world event) last_update_time_ rather than it's parent sensor's update time.

I think all of these issues were fixed in the main gazebo repo in the last major release or two, but it looks like the plugins were not fully updated. There are likely similar issues in the other plugins that should be looked into (by someone who is more familiar with them).


iche033:
I think this is covered in pull request #540 which also fixes the problem for other plugins

j-rivero:
> I think this is covered in pull request #540 which also fixes the problem for other plugins

@kennysharma could you please verify if the problem is already fixed after the latest changes in `kinetic-devel` branch.

j-rivero:
Follow in pr #683

PR #544, opened 2017-02-16
Fix timestamp issues for rendering sensors (indigo-devel)

j-rivero:
{ port of pull request #538 }
This PR builds on top of pull request #410 and applies the timestamp fix to kinect_openni and prosilica sensors


j-rivero:
Ouch, bad use of the clone script, sorry.

PR #556, opened 2017-03-23
Restore catkin fix

davetcoleman:
Was held back for Kinetic release due to regressions, see https://github.com/ros-simulation/gazebo_ros_pkgs/pull/552

### DIFF
--- a/gazebo_plugins/CMakeLists.txt
+++ b/gazebo_plugins/CMakeLists.txt
@@ -1,24 +1,26 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(gazebo_plugins)
 
+option(ENABLE_DISPLAY_TESTS "Enable the building of tests that requires a display" OFF)
+
 find_package(catkin REQUIRED COMPONENTS 
-  message_generation 
-  gazebo_msgs 
-  roscpp 
-  rospy 
-  nodelet 
-  angles 
-  std_srvs 
-  geometry_msgs 
-  sensor_msgs 
-  nav_msgs 
-  urdf 
-  tf 
-  tf2_ros 
-  dynamic_reconfigure 
-  rosgraph_msgs 
-  trajectory_msgs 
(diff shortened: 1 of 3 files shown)


j-rivero:
The problem (opencv related) seems unrelated to these changes. I'm still a bit afraid of breaking people's code with this change but I also think that is the right fix. I will try to write a good changelog and hopefully google is already aware of our big number of issues in different trackers so people can find quickly the way to workaround in their software.

davetcoleman:
At least some of the plugins [have already been fixed](https://github.com/tu-darmstadt-ros-pkg/hector_gazebo/commit/271695d90e9115d5ab53ecb5446525790f10c313) after you released this change - it just needs soak time. I'm happy to contact package maintainers of broken plugins next time you release gazebo_ros_pkgs - which hopefully will be soon so downstream packages have time to fix their plugins.

davetcoleman:
This PR is failing because of 

```
CMake Error at /opt/ros/kinetic/share/cv_bridge/cmake/cv_bridgeConfig.cmake:106 (message):

  Project 'cv_bridge' specifies '/opt/ros/kinetic/include/opencv-3.2.0-dev'

  as an include dir, which is not found.  It does neither exist as an

  absolute directory nor in

  '/opt/ros/kinetic//opt/ros/kinetic/include/opencv-3.2.0-dev'.  Ask the

  maintainer 'Vincent Rabaud <vincent.rabaud@@gmail.com>' to fix it.

Call Stack (most recent call first):

  /opt/ros/kinetic/share/catkin/cmake/catkinConfig.cmake:76 (find_package)

  CMakeLists.txt:6 (find_package)
```

Seems very unrelated

davetcoleman:
@j-rivero is there a way to re-trigger the build without pushing a commit?

j-rivero:
> Seems very unrelated

It is, we can safely merge. Thanks Dave.

PR #567, opened 2017-04-24
Revert catkin warning fix

j-rivero:
See https://github.com/yujinrobot/kobuki_desktop/issues/50 . @davetcoleman the commit causes regressions in different points that would be better resolved by the inclusion of the gazebo_dev (pull request #473). I would prefer not to revert again the commit so let's role a new fast release to solve third party regressions and afterwards merge #473 into the kinetic-devel and lunar-devel branches.


j-rivero:
@osrf-jenkins run test please

PR #568, opened 2017-04-24
Add catkin package(s) to provide the default version of Gazebo (kinetic-devel)

j-rivero:
{ port of pull request #473 }

As a follow-up of the discussion in https://github.com/tu-darmstadt-ros-pkg/hector_gazebo/pull/30 (@mikepurvis, @tfoote, @meyerj), I created a proof-of-concept catkin package `gazebo_dev`.

The new package contains no code and its main purpose is to depend on the respective ROS distro's default version of the Gazebo development library package. Other packages that want to link to the Gazebo libraries can in turn depend on `gazebo_dev`, removing the need to maintain separate branches for different ROS and Gazebo versions as long as the source code is compatible. The current practice of depending on `gazebo_ros` pulls in some unnecessary dependencies (see [original discussion](https://github.com/tu-darmstadt-ros-pkg/hector_gazebo/pull/30#issuecomment-227872716)).
Secondary, the package provides a cmake script exposed via catkin's [CFG_EXTRAS mechanism](http://docs.ros.org/indigo/api/catkin/html/dev_guide/generated_cmake_api.html#catkin-package) so that the `gazebo_dev` package can be used in the same way as any other catkin package without the need to find the Gazebo libraries or apply C++ flags explicitly. I modified the other packages in this repository to only depend on `gazebo_dev` and removed the obsolete cmake code. It should be noted that all packages that directly depend on `gazebo_dev` will compile in C++11 mode because of the addition of `GAZEBO_CXX_FLAGS` to `CMAKE_CXX_FLAGS`. That's why this dependency has not been exported via the `CATKIN_DEPENDS` key of `catkin_package()` in the `gazebo_plugins` package. Derived packages have to depend on `gazebo_dev` explicitly.

Possible extensions would be a check in [gazebo_dev-extras.cmake](https://github.com/ros-simulation/gazebo_ros_pkgs/compare/indigo-devel...meyerj:indigo-gazebo-dev?expand=1#diff-3deb71e85d4c63095f615914b7270470) whether the found Gazebo version is indeed the expected one or even to provide a way to override the required version by means of cmake variables.

The `gazebo_dev` package does not declare a run-time dependency for the same Gazebo version at the moment. I am not sure if this is required. A rosdep key `gazebo_default` could be established that points to the correct gazebo binary package, so that derived packages like gazebo_plugins can build_depend on `gazebo_dev` and exec_depend on `gazebo_default`. Or is there a need for a separate run-time only package with that name, which would provide the per-distro `gazebo_default` key? Like

```
gazebo <----------\
libgazebo5-dev <---+--(build_export_depend)-- gazebo_dev (ROS) <------\
libgazebo7-dev <--/                                                 (build_depend)
                                                                        +---------- gazebo_plugins (ROS)
gazebo <---\                                                        (exec_depend)
gazebo5 <---+----------(exec_depend)--------- gazebo_default (ROS) <--/
gazebo7 <--/
```

This patch also converts all package manifest to format 2 according to [REP 140](http://www.ros.org/reps/rep-0140.html) to be able to use the new `<build_export_depend>` tag. Furthermore it removes the unnecessary dependency `message_generation` from package `gazebo_ros`.

I also created a [branch indigo-gazebo-dev](https://github.com/meyerj/hector_gazebo/tree/indigo-gazebo-dev) in [hector_gazebo](http://wiki.ros.org/hector_gazebo) ([diff](https://github.com/tu-darmstadt-ros-pkg/hector_gazebo/compare/indigo-devel...meyerj:indigo-gazebo-dev)) which applies a patch to use the new `gazebo_dev` package and which compiles successfully in ROS indigo, jade and kinetic with the respective default Gazebo versions.

### DIFF
--- a/gazebo_dev/CMakeLists.txt
+++ b/gazebo_dev/CMakeLists.txt
@@ -1,0 +1,8 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(gazebo_dev)
+
+find_package(catkin REQUIRED)
+
+catkin_package(
+  CFG_EXTRAS gazebo_dev-extras.cmake
+)
--- a/gazebo_dev/cmake/gazebo_dev-extras.cmake
+++ b/gazebo_dev/cmake/gazebo_dev-extras.cmake
@@ -1,0 +1,13 @@
+# Depend on system install of Gazebo
+find_package(gazebo REQUIRED)
+
+message(STATUS "Gazebo version: ${GAZEBO_VERSION}")
+
+# The following lines will tell catkin to add the Gazebo directories and libraries to the
+# respective catkin_* cmake variables.
+set(gazebo_dev_INCLUDE_DIRS ${GAZEBO_INCLUDE_DIRS})
+set(gazebo_dev_LIBRARY_DIRS ${GAZEBO_LIBRARY_DIRS})
+set(gazebo_dev_LIBRARIES ${GAZEBO_LIBRARIES})
+
+# Append gazebo CXX_FLAGS to CMAKE_CXX_FLAGS (c++11)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${GAZEBO_CXX_FLAGS}")
--- a/gazebo_dev/package.xml
+++ b/gazebo_dev/package.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>gazebo_dev</name>
+  <version>0.0.0</version>
+  <description>
+    Provides a cmake config for the default version of Gazebo for the ROS distribution.
+  </description>
+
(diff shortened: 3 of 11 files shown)


meyerj:
I already had a branch based on kinetic-devel linked from https://github.com/ros-simulation/gazebo_ros_pkgs/pull/473#issuecomment-266531414: https://github.com/ros-simulation/gazebo_ros_pkgs/compare/kinetic-devel...meyerj:kinetic-gazebo-dev

j-rivero:
> I already had a branch based on kinetic-devel linked from #473 (comment): kinetic-devel...meyerj:kinetic-gazebo-dev

Arg, sorry I missed it. Would you mind creating a pull request so I can close this one?

j-rivero:
> Arg, sorry I missed it. Would you mind creating a pull request so I can close this one?

I've forked your branch, solve the conflicts and create a new pr #571

Issue #569, opened 2017-04-25
[gazebo_plugins] Depth info not published in ros


ksatyaki:
For the Depth Camera plugin, the depth info is not published. I'm now working on cleaning up some code and making the plugin publish the right camera info.

PR #570, opened 2017-04-25
[gazebo_plugins] Publish camera info properly for depth cameras.

ksatyaki:
Aims to address #569


ksatyaki:
Has this been addressed already? Any info?

j-rivero:
> Has this been addressed already? Any info?

I don't think so. Some comments after giving a quick look into the PR: 
 * You are removing some protected functions so we are altering the API in the PR. Since `kinetic` is a mature release I'm not in favour of ABI changes to it. If you are interested into getting the changes in `kinetic` we should come up with changes that respect the ABI. API changes can go into the next Melodic release.
* I would like to include a test or at least an example that somehow can be run to demonstrate that we are not breaking camera support and that helps to verify that the new information is indeed being published.

Thanks!

PR #572, opened 2017-04-25
Allow disabling distorted camera border crop (and associated tests)

Kukanani:
Note that this PR is dependent on the merge of [this PR](https://bitbucket.org/osrf/gazebo/pull-requests/2678/pincushion-distortion-model/diff), which implements pincushion distortion in Gazebo. It also depends on the checkboard_plane from [this PR on gazebo_models](https://bitbucket.org/osrf/gazebo_models/pull-requests/253/textured-camera-calibration-validation/diff).

Adds a `<borderCrop>` tag to the sensor plugin XML. The default behavior is `border_crop_ = true`, which retains the same functionality by default.

Disabling border crop makes the distorted camera image match the expressed distortion parameters. This allows the creation of tests to ensure that the distortion is acting as expected. The tests use OpenCV's undistort() with varying camera parameters, and checks that the provided distortion parameters result in the best possible undistortion.

@iche033


j-rivero:
@osrf-jenkins run test please

Kukanani:
The tests will not pass until [this PR](https://bitbucket.org/osrf/gazebo/pull-requests/2678/pincushion-distortion-model/diff) is merged, which will add pincushion distortion to Gazebo.

j-rivero:
> The tests will not pass until this PR is merged, which will add pincushion distortion to Gazebo.

Thanks Adam. We will probably need to release a new version of gazebo8 to propagate gazebo changes to this CI scripts. With the comment I was testing the infrastructure to check if some problems are solved, please ignore it.

Kukanani:
Both prerequisite PRs have been merged into gazebo7 and gazebo_models.

j-rivero:
@osrf-jenkins run test please

Kukanani:
The tests pass on my local Xenial machine. Is there anything I should keep in mind while trying to replicated the failure locally? I'm running the tests with 1 active thread only.

Kukanani:
I'm not sure why the tests are timing out on Jenkins, because my local machine completes them in less than 5 seconds. But just to be safe, I increased the test execution time. If these tests still timeout, I'm not sure what else to try next.

gazebo_plugins/CMakeLists.txt
```diff
@@ -366,12 +366,24 @@ install(DIRECTORY test
...
 if (CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
+
   add_rostest_gtest(set_model_state-test
                     test/set_model_state_test/set_model_state_test.test
                     test/set_model_state_test/set_model_state_test.cpp)
-  add_rostest(test/range/range_plugin.test)
   target_link_libraries(set_model_state-test ${catkin_LIBRARIES})
 
+  add_rostest(test/range/range_plugin.test)
```
j-rivero, line 375: Looks to me that we would need a valid DISPLAY to run them, right? They are failing in the ROS buildfarm (no DISPLAY) but are passing in the Gazebo buildfarm.

If this is the case, look at line 387 and place the test under the ENABLE_DISPLAY_TESTS conditional please.

Kukanani: The cameras render images, but no GUI is involved. Does that require a DISPLAY?

j-rivero: AFAIK they need a DISPLAY for creating the rendering. @iche033 could you please confirm it?

Kukanani:
Looks like the tests are passing now, thanks for the tip! I'm not sure what's causing the single failure case.

j-rivero:
> Looks like the tests are passing now, thanks for the tip! I'm not sure what's causing the single failure case.

I think that we are good to merge. The problem is with `gazebo8` (note that officially kinetic is supporting `gazebo7`) that deprecates a good bunch of functions. It is not a big deal since: 1. it is an unofficial version 2. the problem are only warnings. Merging it now, thanks you very much for the contribution Adam.

gazebo_plugins/CMakeLists.txt
iche033: a bit too late, but yes, it requires a DISPLAY

gazebo_plugins/src/gazebo_ros_camera_utils.cpp
```diff
@@ -231,13 +232,13 @@ void GazeboRosCameraUtils::Load(sensors::SensorPtr _parent,
   else
     this->distortion_t2_ = this->sdf->Get<double>("distortionT2");
 
-  if ((this->distortion_k1_ != 0.0) || (this->distortion_k2_ != 0.0) ||
-      (this->distortion_k3_ != 0.0) || (this->distortion_t1_ != 0.0) ||
-      (this->distortion_t2_ != 0.0))
+  if (!this->sdf->HasElement("borderCrop"))
```
iche033, line 235: it would be nice to document the new `borderCrop` parameter somewhere, perhaps in your new [ROS depth camera](http://gazebosim.org/tutorials?tut=ros_depth_camera&branch=ros_depth_camera_preview) tutorial or the [Gazebo plugins in ROS tutorial ](http://gazebosim.org/tutorials?tut=ros_gzplugins&cat=connect_ros)

PR #584, opened 2017-05-23
gazebo_ros_api_plugin: improve plugin xml parsing when spawning model

scpeters:
Spawning a model with an xml comment that starts with plugin causes a seg-fault:
`<!--plugin-->`
or
`<!--plugin filename="lib.so"/-->`
Checking the `Type()` prevents trying to add child elements to xml comments, which is the source of the seg-fault. I also switched from `find` to `compare` to make sure that only `<plugin>` blocks are affected instead of elements that start with `plugin` like `<plugin1>`, `<pluginCustom>`, etc.

It should be easy to add a test for this that just spawns some simple models with these xml tags added.

Here's the backtrace I was seeing:

* https://gist.github.com/scpeters/87c9a8c81e0a279ffe7747727a51e9c9

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -2207,7 +2207,8 @@ void GazeboRosApiPlugin::walkChildAddRobotNamespace(TiXmlNode* robot_xml)
   child = robot_xml->IterateChildren(child);
   while (child != NULL)
   {
-    if (child->ValueStr().find(std::string("plugin")) == 0)
+    if (child->Type() == TiXmlNode::TINYXML_ELEMENT &&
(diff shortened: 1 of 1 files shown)


j-rivero:
The PR looks good to me. Before touching Indigo, let's merge and release Lunar and Kinetic.

scpeters:
Sounds good. Should I make separate PR's for K/L?

j-rivero:
> Sounds good. Should I make separate PR's for K/L?

not needed, already done and merged in pull request #627 and #625 . I will launch releases for both and merge this one in the following days.

scpeters:
@j-rivero should we try to merge this to indigo or just close it?

j-rivero:
> @j-rivero should we try to merge this to indigo or just close it?

Seems like a bug fix with low risk so fine for Indigo.

scpeters:
replaced by #731

PR #585, opened 2017-05-27
Parameter to disable ROS network interaction from/to Gazebo

j-rivero:
The PR implements a ROS parameter named `enable_ros_network` that allows to disable all the gazebo topics (except /clock) and services that are created from the `gazebo_ros` package. This is useful in situation where the interaction from user with the gazebo simulator should be limited or blocked.

To keep backwards compatibility, by default, the behaviour is to enable all topics and services if the parameter is not present so current code should not experiment any difference at all.

I've added a couple of tests based on the `ros_api_checker` that the OSRF developed during the times of the Virtual Robotics Challenge. It checks all the ROS topics and services in runtime to check if they are exactly the same that is described in the .yaml files available in this PR.


gazebo_ros/test/ros_network/ros_api_checker
```diff
@@ -0,0 +1,196 @@
...
+    services = []
+    strict = False
+    for f in files:
+        # Ignore args passed in by rostest
+        if f[:2] == '--' or f[:2] == '__':
+            continue
+        # Let parsing exceptions leak out; we'll catch them by noticing the
+        # absence of a test result file.
+        y = yaml.load(open(f))
+        for t in y['topics']:
```
ironmig: ```topics += y['topics']```

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -193,7 +196,20 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
   pub_model_states_connection_count_ = 0;
 
   /// \brief advertise all services
-  advertiseServices();
+  if (enable_ros_network_)
+    advertiseServices();
+
+  // Manage clock for simulated ros time
+  pub_clock_ = nh_->advertise<rosgraph_msgs::Clock>("/clock",10);
```
kev-the-dev, line 203: /clock is already advertised within advertiseServices(), perhaps it should be removed there.

gazebo_ros/test/ros_network/ros_network_default.test
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<launch>
+  <include file="$(find gazebo_ros)/launch/empty_world.launch">
+    <arg name="paused"   value="false"/>
+    <arg name="gui"      value="false"/>
+    <arg name="headless" value="false"/>
+    <arg name="debug"    value="false"/>
+    <arg name="enable_ros_network" value="true" />
```
kev-the-dev, line 8: Perhaps this argument should be excluded entirely in this test to ensure the default behavior is enabled.

kev-the-dev:
Bump to trigger CI, good to merge if no regressions

j-rivero:
out of date, probably can be left out

PR #586, opened 2017-05-29
Parameter to disable ROS network interaction from/to Gazebo (indigo-devel)

j-rivero:
{ port of pull request #585 }
The PR implements a ROS parameter named `enable_ros_network` that allows to disable all the gazebo topics (except /clock) and services that are created from the `gazebo_ros` package. This is useful in situation where the interaction from user with the gazebo simulator should be limited or blocked.

To keep backwards compatibility, by default, the behaviour is to enable all topics and services if the parameter is not present so current code should not experiment any difference at all.

I've added a couple of tests based on the `ros_api_checker` that the OSRF developed during the times of the Virtual Robotics Challenge. It checks all the ROS topics and services in runtime to check if they are exactly the same that is described in the .yaml files available in this PR.


j-rivero:

PR #588, opened 2017-06-01
[gazebo_ros_control] fix initialization issue in position controller

cyborg-x1:
The new e_stop feature seems to be missing an initialization of last_joint_position_command_ array. Which leads to the fact that the array, which will later set the position in the simulation (joint_position_command_), will be a zero length array, cause it is overwritten with the last_joint_position_ command in line 278. That leads to unpredicted behavior for the joint -> mostly in the joint acting like a helicopter rotor for me. My suggested changes seem to fix this for me.


kev-the-dev:
Looking at the code flow, this wouldn't be a problem. Line 292, which sets ```last_joint_position_command = joint_position_``` will always be run before line 295 where they are set the other way.

ahcorde:

PR #590, opened 2017-06-11
[gazebo_ros] Quote arguments to echo in libcommon.sh

daewok:
If /bin/sh is provided by bash, echo will consume arguments such as `-e`. On such a system, running `rosrun gazebo_ros gzserver -e ode empty_world.world` will execute `gzserver` with the `-e` missing (meaning the world file is ignored).


j-rivero:
Looks good to me. Thanks!

PR #601, opened 2017-07-20
Add RTF republisher

jliviero:
Added subscriber/callback to gz topic and ROS topic publisher to republish the current real-time factor.

Note that this is a re-implementation of the Gazebo method found [here](https://bitbucket.org/osrf/gazebo/src/a24b331f8ebfd712a226ba73b7f43d2d4c14fe15/tools/gz.cc?at=gazebo7&fileviewer=file-view-default#gz.cc-854).  The goal is to make the real-time factor available in ROS bags, for use as an easy "system performance" metric with post-run analysis.

### DIFF
--- a/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
+++ b/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
@@ -100,6 +100,9 @@
 
 #include <boost/algorithm/string.hpp>
 
+// For real-time factor calculations
+#include <boost/circular_buffer.hpp>
+
 namespace gazebo
 {
 
@@ -246,7 +249,7 @@ private:
   void publishSimTime();
 
   /// \brief Republish the Gazebo real-time factor
-  void publishRealTimeFactor(ConstWorldStatisticsPtr &_msg);
+  void publishRealTimeFactor();
 
   /// \brief
   void publishLinkStates();
@@ -321,7 +324,6 @@ private:
   gazebo::transport::PublisherPtr light_modify_pub_;
   gazebo::transport::PublisherPtr request_pub_;
   gazebo::transport::SubscriberPtr response_sub_;
-  gazebo::transport::SubscriberPtr rtf_sub_;    // Real-time factor republisher
 
   boost::shared_ptr<ros::NodeHandle> nh_;
   ros::CallbackQueue gazebo_queue_;
@@ -330,6 +332,7 @@ private:
   gazebo::physics::WorldPtr world_;
   gazebo::event::ConnectionPtr wrench_update_event_;
   gazebo::event::ConnectionPtr force_update_event_;
+  gazebo::event::ConnectionPtr rtf_update_event_;
   gazebo::event::ConnectionPtr time_update_event_;
   gazebo::event::ConnectionPtr pub_link_states_event_;
   gazebo::event::ConnectionPtr pub_model_states_event_;
@@ -386,10 +389,10 @@ private:
   gazebo::common::Time last_pub_clock_time_;
 
   /// \brief Sim time buffer for real-time factor calculation
-  std::list<common::Time> simTimes;
+  boost::circular_buffer<common::Time> sim_times_;
 
   /// \brief Real time buffer for real-time factor calculation
-  std::list<common::Time> realTimes;
+  boost::circular_buffer<common::Time> real_times_;
 
   /// \brief A mutex to lock access to fields that are used in ROS message callbacks
   boost::mutex lock_;
--- a/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
+++ b/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
@@ -245,6 +245,9 @@ private:
   void publishSimTime(const boost::shared_ptr<gazebo::msgs::WorldStatistics const> &msg);
   void publishSimTime();
 
+  /// \brief Republish the Gazebo real-time factor
+  void publishRealTimeFactor(ConstWorldStatisticsPtr &_msg);
+
   /// \brief
   void publishLinkStates();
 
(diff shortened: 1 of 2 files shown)


jliviero:
Despite the comment [found here](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/601/files#diff-49e473d5079f9419b8c83bba16e6e74eR185), this actually seems to work fine with `gzserver`.  I might be misunderstanding the intent of that message, though.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -187,6 +188,7 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
   light_modify_pub_ = gazebonode_->Advertise<gazebo::msgs::Light>("~/light/modify");
   request_pub_ = gazebonode_->Advertise<gazebo::msgs::Request>("~/request");
   response_sub_ = gazebonode_->Subscribe("~/response",&GazeboRosApiPlugin::onResponse, this);
+  rtf_sub_ = gazebonode_->Subscribe("~/world_stats", &GazeboRosApiPlugin::publishRealTimeFactor, this);
```
kev-the-dev: I believe this can be done without subscribing to world_stats (just use ```world_->GetSimTime()``` and ```world->GetRealTime()``` like in the publishSimTime function.

gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
```diff
@@ -380,6 +385,12 @@ class GazeboRosApiPlugin : public SystemPlugin
   int pub_clock_frequency_;
   gazebo::common::Time last_pub_clock_time_;
 
+  /// \brief Sim time buffer for real-time factor calculation
+  std::list<common::Time> simTimes;
+
+  /// \brief Real time buffer for real-time factor calculation
+  std::list<common::Time> realTimes;
```
kev-the-dev: [boost::circular_buffer](https://www.boost.org/doc/libs/1_65_1/doc/html/circular_buffer.html) would save some hassle with these

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1940,6 +1945,50 @@ void GazeboRosApiPlugin::publishSimTime()
...
+
+  this->realTimes.push_back(msgs::Convert(_msg->real_time()));
+  if (this->realTimes.size() > 20)
+    this->realTimes.pop_front();
+
+  common::Time simAvg, realAvg;
+  std::list<common::Time>::iterator simIter, realIter;
+  simIter = ++(this->simTimes.begin());
+  realIter = ++(this->realTimes.begin());
+  while (simIter != this->simTimes.end() && realIter != this->realTimes.end())
```
kev-the-dev: Rather than recalculating the average every loop, you could store a running sum which you add the new time to and (if the buffer is full) pop and subtract the oldest time from.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1940,6 +1945,50 @@ void GazeboRosApiPlugin::publishSimTime()
   pub_clock_.publish(ros_time_);
 }
 
+void GazeboRosApiPlugin::publishRealTimeFactor(ConstWorldStatisticsPtr &_msg)
```
kev-the-dev: This could just be moved to publishSimTime, ensuring that the /sim_real_time_factor also publishes at the specified pub_clock_frequency

kev-the-dev (changes requested):
Cool feature!

jliviero:
Coming back to this after letting it sit stale for a while.  Sorry for the delay!  @ironmig, I'm implementing your requested changes now.

I think it's probably still good to target `kinetic-devel`, and then cherry-pick to lunar/melodic.

gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
jliviero: I think that's a good idea.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
jliviero: That gets a little redundant when using `boost::circular_buffer` - if we're tracking how full the buffer is anyway, we might as well stick with the `std::list`.

What I'm looking at doing instead is using `std::accumulate` to produce a sum every pass.  Looking at the [original implementation](https://bitbucket.org/osrf/gazebo/src/a24b331f8ebfd712a226ba73b7f43d2d4c14fe15/tools/gz.cc#lines-854:892), it's not _actually_ calculating an average anyway.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
jliviero: I thought about that, but I wanted to keep RTF and sim time logically distinct.  What I've done instead is eliminated the `rtf_sub_` subscriber, and refactored this callback to be triggered on the world updated event (similar to `publishSimTime`)

gazebo_ros/src/gazebo_ros_api_plugin.cpp
jliviero: :+1:

jliviero:
@ironmig are there any other changes here you'd like to see?

jliviero:
> @ironmig are there any other changes here you'd like to see?

Gentle ping on this.  @chapulina, @j-rivero?

gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
chapulina: We've been trying to reduce our boost dependencies, especially when it's straight forward to implement the same thing using the standard library, like in this case.

Here, since Gazebo and gazebo_ros_pkgs already bring Boost as a dependency, this code is not adding a new dependency, so I think it's fine to leave it since an effort has been made to port it to use boost. But beware that we will most likely need to do the opposite effort of removing boost from here in the not-so-near future.

chapulina (approved):
Looks good to me, with the caveat that I think it would be better not to add the extra dependency on boost.

@jliviero , it would be great if you could add a note about the new topic to [this tutorial](http://gazebosim.org/tutorials?tut=ros_comm&cat=connect_ros#GazeboPublishedTopics).

ahcorde:

PR #603, opened 2017-07-27
add option to input xml from stdin to spawn_model

sputnick1124:
This allows for on-the-fly model description (e.g. with `sed`, `xacro.py`, `erb`) without the need to store a model in the database or create temp files for each. Useful for some swarming applications if plugins need different parameters for each instance of the robot.

Example usage:
`rosun gazebo_ros spawn_model -sdf -stdin "$(cat model.sdf | sed 's/arg/newarg/g')" -model modified_robot`
or
`rosrun gazebo_ros spawn_model -urdf -model model_name -stdin "$(rosrun xacro xacro --inorder path/to/model.xacro arg:=value)" -package_to_model`

### DIFF
--- a/gazebo_ros/scripts/spawn_model
+++ b/gazebo_ros/scripts/spawn_model
@@ -42,9 +42,9 @@ model_database_template = """<sdf version="1.4">
 def usage():
     print '''Commands:
     -[urdf|sdf|trimesh|gazebo] - specify incoming xml is urdf, sdf or trimesh format. gazebo arg is deprecated in ROS Hydro
-    -[file|param|database] [<file_name>|<param_name>|<model_name>] - source of the model xml or the trimesh file
+    -[file|param|database|stdin] [<file_name>|<param_name>|<model_name>|<xml>] - source of the model xml or the trimesh file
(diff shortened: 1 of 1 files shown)


j-rivero:

kev-the-dev (changes requested):
I like this idea! However, I think the argument "stdin" is misleading. This seems to actually just read the model_xml from the program arguments. I think a better style is to use ```sys.stdin```, so the usage would be like
 ```cat model.sdf | sed 's/arg/newarg/g' | rosun gazebo_ros spawn_model -sdf -stdin -model modified_robot```

sputnick1124:
Great points, @ironmig! I'll make some changes to the interface to make it more pipe-friendly. I agree that calling it stdin is a bit misleading. I think I originally was going for exactly what you described, but got sidetracked halfway through testing the proof of concept.

kev-the-dev:
Thanks for the updates. I'll be integrating this into my refactor of the script in #730 so it may be merged there instead of here.

kev-the-dev:
Merged into PR 730, thanks for the idea and code contribution!

PR #605, opened 2017-08-08
Fix for relative frame errors

hamzamerzic:
I noticed a lot of errors when using relative reference frames in Gazebo. When spawning the object, wrong multiplication order of quaternions is applied (as per [here](https://bitbucket.org/osrf/gazebo/src/a29d8dabed9a14fb6bafdcd31c84b160c00b127c/gazebo/math/Quaternion.cc?at=gazebo7&fileviewer=file-view-default#Quaternion.cc-391)). The operation q1 *= q2 results in q1 = q1 * q2, but we wish to have q1 = q2 * q1, since we apply rotation to q1. 

When obtaining the relative pose of the object w.r.t. to a frame, I noticed strange and inconsistent transformations. I used the reference for chaining transformations defined [here](http://osrf-distributions.s3.amazonaws.com/gazebo/api/dev/classgazebo_1_1math_1_1Pose.html) to make the transformations cleaner.

Also, instead of using the relative entity, in the whole code the entity is transformed to a link, which does not allow for specifying a model as a relative entity.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -625,7 +625,7 @@ bool GazeboRosApiPlugin::spawnSDFModel(gazebo_msgs::SpawnModel::Request &req,
   gazebo::math::Quaternion initial_q(req.initial_pose.orientation.w,req.initial_pose.orientation.x,req.initial_pose.orientation.y,req.initial_pose.orientation.z);
 
   // refernce frame for initial pose definition, modify initial pose if defined
-  gazebo::physics::LinkPtr frame = boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->GetEntity(req.reference_frame));
+  gazebo::physics::EntityPtr frame = world_->GetEntity(req.reference_frame);
   if (frame)
   {
     // convert to relative pose
@@ -818,7 +818,7 @@ bool GazeboRosApiPlugin::getModelState(gazebo_msgs::GetModelState::Request &req,
                                        gazebo_msgs::GetModelState::Response &res)
 {
   gazebo::physics::ModelPtr model = world_->GetModel(req.model_name);
-  gazebo::physics::LinkPtr frame = boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->GetEntity(req.relative_entity_name));
+  gazebo::physics::EntityPtr frame = world_->GetEntity(req.relative_entity_name);
   if (!model)
   {
     ROS_ERROR_NAMED("api_plugin", "GetModelState: model [%s] does not exist",req.model_name.c_str());
@@ -1063,7 +1063,7 @@ bool GazeboRosApiPlugin::getLinkState(gazebo_msgs::GetLinkState::Request &req,
                                       gazebo_msgs::GetLinkState::Response &res)
 {
   gazebo::physics::LinkPtr body = boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->GetEntity(req.link_name));
-  gazebo::physics::LinkPtr frame = boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->GetEntity(req.reference_frame));
+  gazebo::physics::EntityPtr frame = world_->GetEntity(req.reference_frame);
 
   if (!body)
   {
@@ -1413,7 +1413,7 @@ bool GazeboRosApiPlugin::setModelState(gazebo_msgs::SetModelState::Request &req,
   }
   else
   {
-    gazebo::physics::LinkPtr relative_entity = boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->GetEntity(req.model_state.reference_frame));
+    gazebo::physics::EntityPtr relative_entity = world_->GetEntity(req.model_state.reference_frame);
     if (relative_entity)
     {
       gazebo::math::Pose  frame_pose = relative_entity->GetWorldPose(); // - myBody->GetCoMPose();
@@ -1630,7 +1630,7 @@ bool GazeboRosApiPlugin::setLinkState(gazebo_msgs::SetLinkState::Request &req,
                                       gazebo_msgs::SetLinkState::Response &res)
 {
   gazebo::physics::LinkPtr body = boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->GetEntity(req.link_state.link_name));
-  gazebo::physics::LinkPtr frame = boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->GetEntity(req.link_state.reference_frame));
+  gazebo::physics::EntityPtr frame = world_->GetEntity(req.link_state.reference_frame);
   if (!body)
   {
     ROS_ERROR_NAMED("api_plugin", "Updating LinkState: link [%s] does not exist",req.link_state.link_name.c_str());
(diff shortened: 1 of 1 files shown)


hamzamerzic:
@j-rivero What does this label mean? Should I provide something else? 
Also, is the first test broken, or are there errors in the code?

j-rivero:
Hello Hamza:

> @j-rivero What does this label mean? Should I provide something else?

Nop. I use this label to ask for the help of the rest of Gazebo team since the changes exceeds my knowledge. I have the feeling that if we merge this code it will potentially affect all gazebo_ros_pkgs users changing the behaviour of the SpawnSDF model, right? If so, we need to find another way of fixing so we don't break previous behavior. 

> Also, is the first test broken, or are there errors in the code?

It is because of the compiler warnings which were never fixed for the gazebo8 version. Don't worry about it. Thanks!

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1082,9 +1081,9 @@ bool GazeboRosApiPlugin::getLinkState(gazebo_msgs::GetLinkState::Request &req,
...
     gazebo::math::Pose frame_pose = frame->GetWorldPose();
-    body_pose.pos = body_pose.pos - frame_pose.pos;
-    body_pose.pos = frame_pose.rot.RotateVectorReverse(body_pose.pos);
-    body_pose.rot *= frame_pose.rot.GetInverse();
+    gazebo::math::Pose body_rel_pose = body_pose - frame_pose;
+    body_pose.pos = body_rel_pose.pos;
+    body_pose.rot = body_rel_pose.rot;
 
     // convert to relative rates
     gazebo::math::Vector3 frame_vpos = frame->GetWorldLinearVel(); // get velocity in gazebo frame
```
v4hn: I believe you also want to have `RotateVectorInverse` right below.
You changed it in a similar situation further above.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1417,18 +1416,13 @@ bool GazeboRosApiPlugin::setModelState(gazebo_msgs::SetModelState::Request &req,
     if (relative_entity)
     {
       gazebo::math::Pose  frame_pose = relative_entity->GetWorldPose(); // - myBody->GetCoMPose();
-      gazebo::math::Vector3 frame_pos = frame_pose.pos;
-      gazebo::math::Quaternion frame_rot = frame_pose.rot;
 
-      //std::cout << " debug : " << relative_entity->GetName() << " : " << frame_pose << " : " << target_pose << std::endl;
-      //target_pose = frame_pose + target_pose; // seems buggy, use my own
-      target_pose.pos = model->GetWorldPose().pos + frame_rot.RotateVector(target_pos);
```
v4hn: Notice that the old code is plain *wrong and broken* here. It moves the object pos relative to it's current position instead of relative to the frame!
I just spent half an afternoon only to find old code not using a nice API because it "seemed buggy" years ago.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1417,18 +1416,13 @@ bool GazeboRosApiPlugin::setModelState(gazebo_msgs::SetModelState::Request &req,
...
-      target_pose.rot = frame_rot * target_pose.rot;
+      target_pose = target_pose + frame_pose;
```
v4hn: From the docs of `operator+`:

```      /// \brief Addition operator
      /// A is the transform from O to P specified in frame O
      /// B is the transform from P to Q specified in frame P
      /// then, B + A is the transform from O to Q specified in
 frame O
```

~~~So shouldn't this be the other way around as previously proposed in the commented code?~~~

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -1655,9 +1649,7 @@ bool GazeboRosApiPlugin::setLinkState(gazebo_msgs::SetLinkState::Request &req,
     gazebo::math::Quaternion frame_rot = frame_pose.rot;
 
     //std::cout << " debug : " << frame->GetName() << " : " << frame_pose << " : " << target_pose << std::endl;
-    //target_pose = frame_pose + target_pose; // seems buggy, use my own
-    target_pose.pos = frame_pos + frame_rot.RotateVector(target_pos);
-    target_pose.rot = frame_rot * target_pose.rot;
+    target_pose = target_pose + frame_pose;
```
v4hn:
~~~Same as above. Shouldn't this be the other way around. I haven't tested this myself.~~~

v4hn (commented):
Please see my attached comments.
I find it very surprising to see transform bugs in this code that lay dormant for the last two years and longer.

👍 @hamzamerzic for filing the request.
I would have proposed an inferior version today.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
hamzamerzic: Please correct me if I'm wrong. The way I see it is we want to specify the target pose in a different frame, so the target pose equates to the transform from P to Q, while the frame is then the transform O to P?

gazebo_ros/src/gazebo_ros_api_plugin.cpp
hamzamerzic: I guess I missed that bit. :+1:

hamzamerzic:
@v4hn From your responses, it is not fully clear to me which parts of the PR (if any) you agree with. In case some test cases are required, I would gladly try to provide them. And I agree that it is quite surprising to find bugs in the code sitting here for so long, and of course it is quite possible that I misunderstood the intended functionalities - I did not perform extensive checks here, and have in fact stopped using this code.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
v4hn: You are right, I'm sorry. I was too used to a different conversion.
I just tested your code and it works well.

v4hn:
@hamzamerzic 
First thing to say, I'm not a maintainer for this package or gazebo. I just stumbled upon this request because I hit one of the bugs you fix here.

I just tested the part I was unsure about myself and you were right.

I approve this patch as is. I would propose to also add (and verify) the additional `RotateVectorInverse` I mentioned in my review above.

> [I] have in fact stopped using this code.

I believe this is what many people did and it's probably the reason why it hasn't been fixed earlier.
Thanks again for taking the initiative on this matter!

@j-rivero 

> I have the feeling that if we merge this code it will potentially affect all gazebo_ros_pkgs users changing the behaviour of the SpawnSDF model, right?
> If so, we need to find another way of fixing so we don't break previous behavior.

The old code is plain broke beyond usability at least in the "SetModelState" method.
The other transforms are wrong too and you break expectations of users that rely on well-defined API.
If people rely on the broken transformations without reporting them to upstream, I find it very problematic to retain broken behavior for them.

If possible, please give this some priority.

v4hn:
ping @j-rivero ?

scpeters:
I've seen some recent discussion about the `*` operator seeming backwards:

* https://bitbucket.org/ignitionrobotics/ign-math/pull-requests/205/fixing-const-correctness-for-operator-of/diff#comment-48482192

j-rivero:
> I've seen some recent discussion about the * operator seeming backwards:
> 
>     https://bitbucket.org/ignitionrobotics/ign-math/pull-requests/205/fixing-const-correctness-for-operator-of/diff#comment-48482192

Thanks Steve. How that would affect the changes in this PR?

j-rivero:
Thanks for the work @hamzamerzic and thank to @v4hn for the review and comments. Ready to merge. Could you please merge with current `kinetic-devel` to solve the conflicts?

hamzamerzic:
@j-rivero I am glad the bug is confirmed and everything is ready to be merged. There is one relative transformation that needs to be fixed, as noticed by @v4hn. I will try to include that and resolve the merge conflicts as soon as possible.

j-rivero:
Thanks for resolving the conflicts. Here we go.

PR #606, opened 2017-08-10
[gazebo_ros] don't overwrite parameter "use_sim_time"

manuelilg:
gazebo_ros doesn't overwrite the parameter "use_sim_time", if the parameter is present on parameter-server


j-rivero:
Looks good to me. Thanks very much for contributing Manuel.

PR #607, opened 2017-08-10
[gazebo_ros] don't overwrite parameter "use_sim_time" (lunar-devel)

j-rivero:
{ port of pull request #606 }
gazebo_ros doesn't overwrite the parameter "use_sim_time", if the parameter is present on parameter-server


j-rivero:

PR #609, opened 2017-08-17
gazebo_plugins: throw error in case a joint is not found

kekeblom:
Gazebo would segfault if the joint_state_publisher plugin is used with a joint which is not defined. This is kind of hard to debug so I throw a fatal error.

This can be tested by taking a working urdf model using the joint state publisher with a joint name that does not exist. For example:
```
   <gazebo>
     <plugin name="joint_state_publisher" filename="libgazebo_ros_joint_state_publisher.so">
       <jointName>joint_name_that_does_not_exist</jointName>
     </plugin>
   </gazebo>
``` 

The issue was initially posted to the gazebo repository here: https://bitbucket.org/osrf/gazebo/issues/2330/gazebo-segfaults-on-launch.

Cheers!


j-rivero:
The error in the ROS buildfarm is due to a network problem. The error in the gazebo8 version is just the usual compiler warnings. And the error in the regression test is due to lack of infrastructure in the third party software used. Sorry for the noise.

The change looks good to me. Thanks!

PR #613, opened 2017-09-28
Feature/make gazebo 9 compatible

ffurrer:
These were the modification I had to do in order to compile with Gazebo9 and hence also with ignition::math.

I didn't try the backward compatibility.

**EDIT:**
So according to ci it works with gazebo8 but not earlier versions, which is expected, I guess :).


j-rivero:
Thanks for the PR. We are in the middle of a clean up which will modify and partially override this PR. As soon as we finish with it, I'll back to see if there is anything still missing in your changes.

scpeters:
we have fixed gazebo9 support in other pull requests

PR #620, opened 2017-11-06
kinnect code optimization

dseredyn:


j-rivero:
Thanks for the PR. Before merging it I would like to check that the produced code is identical or equivalent with the current version and the optimized changes. @dseredyn do you have any testing method at hand?

ahcorde:

PR #624, opened 2017-11-13
Install spawn_model using catkin_install_python (lunar-devel)

j-rivero:
{ port of pull request #621 }


scpeters:
The regression build is complaining about hector_gazebo for some reason

j-rivero:
> The regression build is complaining about hector_gazebo for some reason

I use hector_gazebo as a third party software to test how changes would affect to software that uses gazebo_ros_pkgs. hector_gazebo does not have a lunar branch yet. I made a [patch to change the software for lunar](https://bitbucket.org/osrf/release-tools/pull-requests/470/use-different-third-party-software-for/diff)

PR #628, opened 2017-11-14
Fix gazebo8 warnings part 2: replace private member gazebo::math types with ignition

scpeters:
The gazebo::math types are deprecated in gazebo8. This PR focuses on gazebo::math variables used as private member variables in `gazebo_plugins` header fiels and replaces them with ignition::math types. It also replaces local gazebo::math variables in those files. I also removed a few old compiler directive blocks.


scpeters:
Net reduction of over 200 warnings

PR #632, opened 2017-11-16
Fix gazebo8 warnings part 3: more ign-math in plugins (lunar-devel)

j-rivero:
{ port of pull request #631 }
Convert local more local variables in gazebo_plugins to use ignition math and remove a few outdated header guards in tricycle plugin.


scpeters:
This needs to be rebased on top of #630

scpeters:
that looks better

PR #633, opened 2017-11-16
Fix gazebo8 warnings part 4: convert remaining local variables in plugins to ign-math

scpeters:
Also remove some outdated compiler directives.


scpeters:
Fixed about 35 warnings net

* https://build.osrfoundation.org/job/ros_gazebo8_pkgs-ci-pr_any_kinetic-xenial-amd64/34/

scpeters:
I added one more quick change to use the World Get and Set Gravity functions since they don't require compiler directives. That's the last addition to this PR

PR #635, opened 2017-11-16
Fix gazebo8 warnings part 5: ignition math in gazebo_ros

scpeters:
Replace gazebo::math variables with ignition::math in several places in the private section of gazebo_ros_api_plugin.h and all local variables in the cpp file. Also, change some of the API's to pass const references to objects instead of copying them.


scpeters:
fixed about 220 warnings

PR #637, opened 2017-11-20
Don't ignore robotNamespace

sfocke:
When creating the NodeHandle for reading the PID parameters, the model_nh was always ignored. Instead, all parameters were read from /gazebo_ros_control/pid_gains/<joint_name>/* instead of /<robot_name>/gazebo_ros_control/pid_gains/<joint_name>/*.

This commit restores the intended behavior, i.e., the parameters will now read from <robot_name>/..., where <robot_name> is specified via the robotNamespace plugin parameter or the parent name.

### DIFF
--- a/gazebo_ros_control/src/default_robot_hw_sim.cpp
+++ b/gazebo_ros_control/src/default_robot_hw_sim.cpp
@@ -207,7 +207,7 @@ bool DefaultRobotHWSim::initSim(
       // joint->SetParam("vel") to control the joint.
       const ros::NodeHandle nh(robot_namespace + "/gazebo_ros_control/pid_gains/" +
                                joint_names_[j]);
-      if (pid_controllers_[j].init(nh, true))
+      if (pid_controllers_[j].init(nh))
       {
         switch (joint_control_methods_[j])
         {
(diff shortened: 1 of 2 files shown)


sfocke:
Please note that this changes the current behavior to what was obviously intended by the author. A different possibility would be this:

```diff
-      const ros::NodeHandle nh(model_nh, "/gazebo_ros_control/pid_gains/" +
+      const ros::NodeHandle nh("/gazebo_ros_control/pid_gains/" +
```

... which doesn't change the behavior, only makes it clearer in the code what is actually happening.

j-rivero:
The change looks correct to me but we have to keep in mind the problem of current compatibility with code out there where people relies on the current behavior. I want to see the topics paths a contract like an API. In order to get this change into Kinetic, we should find a solution similar of what we did in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/397 (legacy_mode parameter or similar).

sfocke:
I've added a legacyModeNS parameter to the plugin to deal with compatibility issues.

j-rivero:

j-rivero:

j-rivero:

kev-the-dev (approved):
Preserves current behavior by default and already merged into lunar/melodic

kev-the-dev:

j-rivero:
There is a failing test not related to this PR. Thanks @focke-85 for the contribution.

PR #639, opened 2017-11-27
Fix gazebo8 warnings part 6: ifdef's for functions without Get

scpeters:
Add ifdef blocks for functions that had `Get` prefix removed, and remove some old ifdef blocks. This focuses on the least-commonly used functions that had deprecation warnings. The next PR's will focus on smaller groups of more commonly used functions.


scpeters:
Fixes about 50 warnings relative to the build for #635

PR #640, opened 2017-11-28
Fix gazebo8 warnings part 6: ifdef's for functions without Get (lunar-devel)

j-rivero:
{ port of pull request #639 }
Add ifdef blocks for functions that had `Get` prefix removed, and remove some old ifdef blocks. This focuses on the least-commonly used functions that had deprecation warnings. The next PR's will focus on smaller groups of more commonly used functions.


scpeters:
It seems like this pull request missed the changes to gazebo_ros_diff_drive.cpp that were in #639. It appears that those changes were included in #647 instead.

PR #641, opened 2017-11-28
Fix for relative frame errors (lunar-devel)

j-rivero:
{ port of pull request #605 }
I noticed a lot of errors when using relative reference frames in Gazebo. When spawning the object, wrong multiplication order of quaternions is applied (as per [here](https://bitbucket.org/osrf/gazebo/src/a29d8dabed9a14fb6bafdcd31c84b160c00b127c/gazebo/math/Quaternion.cc?at=gazebo7&fileviewer=file-view-default#Quaternion.cc-391)). The operation q1 *= q2 results in q1 = q1 * q2, but we wish to have q1 = q2 * q1, since we apply rotation to q1. 

When obtaining the relative pose of the object w.r.t. to a frame, I noticed strange and inconsistent transformations. I used the reference for chaining transformations defined [here](http://osrf-distributions.s3.amazonaws.com/gazebo/api/dev/classgazebo_1_1math_1_1Pose.html) to make the transformations cleaner.

Also, instead of using the relative entity, in the whole code the entity is transformed to a link, which does not allow for specifying a model as a relative entity.


j-rivero:
needs to be created again. Bad merge.

Issue #643, opened 2017-12-05
[gazebo_ros] gzserver script overwrites GAZEBO_MODEL_DATABASE_URI


rsmallegoor:
When using Gazebo without ROS, it is possible to set a custom GAZEBO_MODEL_DATABASE_URI. This is required to load models from a specif model server instead of the [default server](http://gazebosim.org/models).

Unfortunately, when using `gazebo_ros`, this variable is overwritten [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/0b99225ae4c3f098a8cbc3873ac312376dd9a1be/gazebo_ros/scripts/gzserver#L30). Is there anything that can be done to prevent this?

If not, should the [script](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/0b99225ae4c3f098a8cbc3873ac312376dd9a1be/gazebo_ros/scripts/gzserver) be modified to handle the GAZEBO_MODEL_DATABASE_URI the same way as the GAZEBO_MASTER_URI variable?

rsmallegoor:
[PR](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/644)

j-rivero:
I think that we can fix the problem with the same solution applied to the `GAZEBO_MASTER_URI` as you suggested.

Another interesting experiment would be to remove completely the call to the setup.sh. Gazebo build system set [default values for all the variable if they are not defined](https://bitbucket.org/osrf/gazebo/src/b27e852937eec3b5f1f3829cc4535b42c1c66a00/gazebo/common/CMakeLists.txt?at=default&fileviewer=file-view-default#CMakeLists.txt-218).

rsmallegoor:
@j-rivero PM #644

j-rivero:
Merged and ported to lunar-devel. Thanks for the PR.

PR #644, opened 2017-12-05
[gazebo_ros] Prevents GAZEBO_MODEL_DATABASE_URI from being overwritten (fix for issue #643)

rsmallegoor:
Fix for issue [#643](https://github.com/ros-simulation/gazebo_ros_pkgs/issues/643)


starthal:
The Jenkins failure seems to be due to compiler warnings. Are these actually relevant to this change?

j-rivero:
> The Jenkins failure seems to be due to compiler warnings. Are these actually relevant to this change?

Not relevant to this change. The PR looks good to me.

PR #642, opened 2017-11-28
Fix gazebo8 warnings part 7: ifdef's for Joint::GetAngle and some cleanup

scpeters:

PR #647, opened 2017-12-14
Fix gazebo8 warnings part 7: ifdef's for Joint::GetAngle and some cleanup (lunar-devel)

j-rivero:
{ port of pull request #642 }


scpeters:
This doesn't appear to have any of the changes from #642. It seems to have the changes from #639 that were missing in #640.

PR #650, opened 2017-12-15
Fix gazebo8 warnings part 8: ifdef's for GetWorldPose

scpeters:
The uses of GetWorldPose were organized such that I could use the following sed scripts to assist in adding ifdef's in 0691135.

~~~
sed -i -e 's@.*GetWorldPose.*@#if GAZEBO_MAJOR_VERSION >= 8\
__REPLACE__&\
\#else\
&\
\#endif
~~~

~~~
sed -i -e \
  's@^__REPLACE__\(.*\)GetWorldPose()\.Ign\(.*\)@\1WorldPose\2@' \
  $(grep -rlI GetWorldPose gazebo_* | grep -v vacuum)
~~~

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_diff_drive.cpp
+++ b/gazebo_plugins/src/gazebo_ros_diff_drive.cpp
@@ -417,7 +417,11 @@ void GazeboRosDiffDrive::publishOdometry ( double step_time )
     }
     if ( odom_source_ == WORLD ) {
         // getting data form gazebo world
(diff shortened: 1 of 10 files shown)


scpeters:
fixes about 80 warnings, down to 203

* https://build.osrfoundation.org/job/ros_gazebo8_pkgs-ci-pr_any_kinetic-xenial-amd64/45/warnings21Result

j-rivero:
getting close. Thanks Steven!

PR #651, opened 2017-12-15
Fix gazebo8 warnings part 8: ifdef's for GetWorldPose (lunar-devel)

j-rivero:
{ port of pull request #650 }
The uses of GetWorldPose were organized such that I could use the following sed scripts to assist in adding ifdef's in 0691135.

~~~
sed -i -e 's@.*GetWorldPose.*@#if GAZEBO_MAJOR_VERSION >= 8\
__REPLACE__&\
\#else\
&\
\#endif
~~~

~~~
sed -i -e \
  's@^__REPLACE__\(.*\)GetWorldPose()\.Ign\(.*\)@\1WorldPose\2@' \
  $(grep -rlI GetWorldPose gazebo_* | grep -v vacuum)
~~~


j-rivero:
Incomplete port. Failing merge.

PR #656, opened 2017-12-20
Fix gazebo8 warnings part 10: ifdefs for GetModel, GetEntity, Light

scpeters:
This adds more ifdefs for functions that let you get pointers to physics objects, like Model, Entity, and Light. The only remaining warnings are from `World::GetSimTime` and `World::GetPhysicsEngine`.


scpeters:
Fixed 40 warnings, down now to 67:

* https://build.osrfoundation.org/job/ros_gazebo8_pkgs-ci-pr_any_kinetic-xenial-amd64/50/warningsResult/

PR #660, opened 2017-12-23
Fix gazebo8 warnings part 7: retry #642 on lunar

scpeters:
ifdef's for Joint::GetAngle and some cleanup (#642)

* fix major version check >= 8, instead of > 8

* gazebo_ros_bumper: use new API in commented code

* gazebo_ros_api_plugin: world pose in local vars

* worldLinearVel as local var in hand of god plugin

* gazebo8+: Joint::GetAngle -> Joint::Position


j-rivero:
67 left. Getting closer ...

PR #668, opened 2018-01-15
Adding support for multiple (greater than two) cameras with gazebo_ros_multicam

jbrookshire:
Adding support for multiple (greater than two) cameras with gazebo_ros_multicam


j-rivero:
Thanks for the PR.

If I'm not wrong, the PR is modifying the API so at this point would hard to get it into `lunar` and probably a really bad idea for `kinetic`. We are on time to include the change for the next melodic or try to addapt the changes not to modify the API.

ahcorde:

PR #669, opened 2018-01-25
Adding velocity to joint state publisher gazebo plugin

ahcorde:
I checked [Gazebo >= 8 API](gazebosim.org/api/8.0.0/classgazebo_1_1physics_1_1Joint.html) and [Gazebo < 8](http://osrf-distributions.s3.amazonaws.com/gazebo/api/7.1.0/classgazebo_1_1physics_1_1Joint.html) API and It seems that joint velocity method it's call in the same way.


scpeters:
Thanks for submitting this. We had made a similar change on an indigo branch (30a3d912ace523cd7860708f25cf3eca4b55ef95) while working on a project but hadn't managed to merge it forward.

cc @gerkey

scpeters (approved):

gerkey:
Great minds think alike!

jacknlliu:
@ahcorde I test this plugin with [universal_robot](https://github.com/ros-industrial/universal_robot) and ROS kinetic, gazebo 7, but it seems that this plugin can only get the first joint velocity, other joints all have zero velocity.

PR #670, opened 2018-01-25
Load the libgazebo_ros_api_plugin when starting gzclient

gerkey:
This change makes the ROS event loop turn over, which is required when you have a client-side Gazebo plugin that uses ROS. Otherwise if you use ROS in a gzclient plugin, ROS subscriptions and such will appear to work, but then nothing will happen because the event loop isn't turning over.


scpeters:
@osrf-jenkins run test please

scpeters:

scpeters:

iche033 (approved):

PR #671, opened 2018-01-25
Adding velocity to joint state publisher gazebo plugin (lunar-devel)

j-rivero:
{ port of pull request #669 }
I checked [Gazebo >= 8 API](gazebosim.org/api/8.0.0/classgazebo_1_1physics_1_1Joint.html) and [Gazebo < 8](http://osrf-distributions.s3.amazonaws.com/gazebo/api/7.1.0/classgazebo_1_1physics_1_1Joint.html) API and It seems that joint velocity method it's call in the same way.


j-rivero:
Thanks @ahcorde !

PR #674, opened 2018-02-07
[WIP] Create a rosbag playback plugin

tfoote:
I worked on this but it's been backgrounded so this is currently for visibility. 

Open TODOs:
 * Move the plugin into a separate package
 * Resolve an issue that the event doesn't get triggered when gazebo is in playback mode.


ahcorde:

PR #676, opened 2018-02-16
Load the libgazebo_ros_api_plugin when starting gzclient (lunar-devel)

j-rivero:
{ port of pull request #670 }
This change makes the ROS event loop turn over, which is required when you have a client-side Gazebo plugin that uses ROS. Otherwise if you use ROS in a gzclient plugin, ROS subscriptions and such will appear to work, but then nothing will happen because the event loop isn't turning over.


j-rivero:

PR #678, opened 2018-02-22
Fix tests and compiler warnings on kinetic-devel

scpeters:
Don't record state logs with `-r` in rostest's; add --verbose instead.

Fix gazebo9 deprecation warnings by removing Set.*Accel calls since they don't do anything


scpeters:
I don't know why, but some of the tests have been failing since the latest gazebo7 and gazebo8 minor releases; removing `-r` seems to fix it.

scpeters:
@iche033 pointed out that the following issue is the likely cause of the test failures:

* https://bitbucket.org/osrf/gazebo/issues/2425/significantly-lower-rtf-when-state-logging

I think we can make this change anyway, though, since we shouldn't need to record a state log during these tests

iche033 (approved):

scpeters:
I'm going to merge this and rebase some other PR's to fix their CI

PR #683, opened 2018-03-06
Fix sensors after time reset

jkooij:
Fixes various gazebo sensor plugins not publishing ROS messages after resetting the Gazebo world (CTRL+R).

Tested on gazebo9. Currently includes:

- block laser
- camera
- joint_state_publisher
- p3d
- range
These are the sensors on the prius vehicle in github.com/osrf/car_demo .

In all these plugins, resetting the time in Gazebo results in negative time difference between the current time and last update time, which prevents any new messages being send. The fix explicitly detects such negative time differences. With it, the simulation in osrf/car_demo can be reset without affecting ROS.

By the way, this seems related to #532, though kennysharm'a pull request, nor #540 fixes this.


j-rivero:

j-rivero:

PR #685, opened 2018-03-08
gazebo_plugins: unique names for distortion tests

scpeters:
The distortion tests have been using the same gtest code, but that gives them the same test name, which makes it hard to see which test failed in the results. This moves the gtest code to a function in a header file and uses distinct cpp files with different test names for each test case.

PR #686, opened 2018-03-08
gazebo_plugins: unique names for distortion tests (lunar-devel)

j-rivero:
{ port of pull request #685 }
The distortion tests have been using the same gtest code, but that gives them the same test name, which makes it hard to see which test failed in the results. This moves the gtest code to a function in a header file and uses distinct cpp files with different test names for each test case.


j-rivero:

Issue #681, opened 2018-03-05
Test failure: gazebo_plugins.rosunit-distortion_barrel_test


scpeters:
The [barrel distortion test](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/test/camera/distortion_barrel.test) is [failing](https://build.osrfoundation.org/job/ros_gazebo8_pkgs-ci-pr_any_kinetic-xenial-amd64/70/testReport/junit/(root)/DistortionTest/cameraDistortionTest_2/). I see the following console output on my desktop machine (gazebo9 deb, xenial, kinetic):

~~~
-- run_tests.py: execute commands
  /opt/ros/kinetic/share/rostest/cmake/../../../bin/rostest --pkgdir=/home/scpeters/ws/gazebo_ros_pkgs/src/gazebo_ros_pkgs/gazebo_plugins --package=gazebo_plugins --results-filename test_camera_distortion_barrel.xml --results-base-dir /data_fast/scpeters/ws/gazebo_ros_pkgs/build/gazebo_plugins/test_results /home/scpeters/ws/gazebo_ros_pkgs/src/gazebo_ros_pkgs/gazebo_plugins/test/camera/distortion_barrel.test 
... logging to /home/scpeters/.ros/log/rostest-t2-28237.log
[ROSUNIT] Outputting test results to /data_fast/scpeters/ws/gazebo_ros_pkgs/build/gazebo_plugins/test_results/gazebo_plugins/rostest-test_camera_distortion_barrel.xml
Gazebo multi-robot simulator, version 9.0.0
Copyright (C) 2012 Open Source Robotics Foundation.
Released under the Apache 2 License.
http://gazebosim.org

[ INFO] [1520029634.783625276]: Finished loading Gazebo ROS API Plugin.
[ INFO] [1520029634.783996684]: waitForService: Service [/gazebo/set_physics_properties] has not been advertised, waiting...
[Msg] Waiting for master.
[Msg] Connected to gazebo master @ http://127.0.0.1:11345
[Msg] Publicized address: 172.23.2.90
[ INFO] [1520029636.095044292]: Camera Plugin: The 'robotNamespace' param did not exit
[ INFO] [1520029636.097818691]: Camera Plugin: The 'robotNamespace' param did not exit
[ INFO] [1520029636.099974611]: Camera Plugin (ns = )  <tf_prefix_>, set to ""
[...]

PR #687, opened 2018-03-10
Triggered camera / multicamera plugins

scpeters:

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_triggered_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_triggered_camera.cpp
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2013 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+/*
+ @mainpage
+   Desc: GazeboRosTriggeredCamera plugin for simulating cameras in Gazebo
+   Author: John Hsu
+   Date: 24 Sept 2008
+*/
+
+#include "gazebo_plugins/gazebo_ros_triggered_camera.h"
+
+#include <float.h>
+#include <string>
+
+#include <gazebo/sensors/Sensor.hh>
+#include <gazebo/sensors/CameraSensor.hh>
+#include <gazebo/sensors/SensorTypes.hh>
+
+namespace gazebo
+{
+// Register this plugin with the simulator
+GZ_REGISTER_SENSOR_PLUGIN(GazeboRosTriggeredCamera)
+
+////////////////////////////////////////////////////////////////////////////////
+// Constructor
+GazeboRosTriggeredCamera::GazeboRosTriggeredCamera()
+{
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Destructor
+GazeboRosTriggeredCamera::~GazeboRosTriggeredCamera()
+{
+  ROS_DEBUG_STREAM_NAMED("camera","Unloaded");
+}
+
+void GazeboRosTriggeredCamera::Load(sensors::SensorPtr _parent, sdf::ElementPtr _sdf)
+{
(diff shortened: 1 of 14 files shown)


scpeters:
I filed an issue with gazebo because the first triggered image looks different (and wrong) compared to subsequent images:

* https://bitbucket.org/osrf/gazebo/issues/2438/first-camera-image-has-incorrect-visuals

gazebo_plugins/src/gazebo_ros_triggered_camera.cpp
```diff
@@ -0,0 +1,129 @@
...
+          std::bind(&GazeboRosTriggeredCamera::PreRender, this));
+}
+
+void GazeboRosTriggeredCamera::Load(sensors::SensorPtr _parent,
+  sdf::ElementPtr _sdf,
+  const std::string &_camera_name_suffix,
+  double _hack_baseline)
+{
+  GazeboRosCameraUtils::Load(_parent, _sdf, _camera_name_suffix, _hack_baseline);
+
```
iche033, line 79: add another `this->SetCameraEnabled(false);` here in this overloaded `Load` function?

gazebo_plugins/test/camera/triggered_camera.cpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+  EXPECT_EQ(images_received_, 2);
+
+  // then wait for 3 seconds to confirm that we don't receive any more images
+  for (unsigned int i = 0; i < 30; ++i)
+  {
+    ros::spinOnce();
+    ros::Duration(0.1).sleep();
+  }
+  EXPECT_EQ(images_received_, 2);
```
iche033: is there logic to filter out mgs close together?  I'm getting an error:

~~~
[gazebo_plugins.rosunit-triggered_camera/cameraSubscribeTest][FAILURE]----------
/home/osrf/code/ian/gz_ws/src/gazebo_ros_pkgs/gazebo_plugins/test/camera/triggered_camera.cpp:103
Value of: 2
Expected: images_received_
Which is: 3
~~~

but the tests are passing on jenkins..

scpeters:
I just rebased on kinetic-devel in order to get the renamed distortion tests.

gazebo_plugins/test/camera/triggered_camera.cpp
scpeters: hm...it is kind of a race condition, but it was pretty consistent on my machine

are you consistently getting this result?

gazebo_plugins/src/gazebo_ros_triggered_camera.cpp
scpeters: good call

scpeters: squashed into 7875cbd85f87a1f4990b8b61fa3b4c0ff1c6e4d4

gazebo_plugins/test/camera/triggered_camera.cpp
iche033: It was pretty consistent when I was testing it before but I just ran the test 10 more times and it only failed once so likely a race condition.

scpeters: an alternative is to try to change the behavior so that each trigger signal leads to a new image. this would require a slight refactoring (use `int` instead of `bool` in a few places) but would be more predictable

iche033: I think I've fixed the race condition in cb9f299. Now the plugin uses an `int` to keep track of the number of triggers received from the topic.

iche033:
mmm the new builds are failing because `cv_bridge` is not found

j-rivero:
@scpeters would you mind fixing the conflicts when you have some free time? Thanks.

scpeters:
@j-rivero sorry for the delay; I just resolved the conflicts

scpeters:
I tested locally with gazebo9 and the tests pass except for the barrel distortion test (as expected due to #681 )

scpeters:
> mmm the new builds are failing because `cv_bridge` is not found

maybe we aren't declaring an explicit dependency on it; let me check

scpeters:
Someone else had the `cv_bridge` problem too ( https://github.com/ros-perception/vision_opencv/issues/209 ), and it's been fixed and released in `ros-kinetic-cv-bridge` 1.12.8, which is currently in shadow-fixed. I'll check to see when that will sync.

scpeters:

scpeters:

scpeters:

scpeters:
The builds are now unstable, with just the distortion camera test failures

iche033:
just tested everything again, works as expected.

iche033 (approved):

j-rivero:
I've added the missing entry to `SENSORS.md` in f1b28c93b9815bafd43123b0010d47ab597390fb

PR #688, opened 2018-03-16
Fix #612 for Gazebo9

### LINKED ISSUE #612: [gazebo_ros_control] Strange behaviour when adding gazebo_ros_control to a robot
I was modeling up an arm, and everything seemed to be working until I added the ros control plugin:
```xml
  <gazebo>
    <plugin filename="libgazebo_ros_control.so" name="gazebo_ros_control">
      <robotSimType>gazebo_ros_control/DefaultRobotHWSim</robotSimType>
    </plugin>
  </gazebo>
```

The only difference between these two arms is that the arm in front has gazebo_ros_control plugin enabled, and the one in back has those lines commented out.

comparison

I've verified the inertial properties of are all set right, tried it with collisions disabled completely and without, its the same both ways. I am still able to control the joints while the arm is falling, and they behave as expected, with normal speed, though sluggish acceleration.

Even more curious, when I manually change any one of the four revolute joints on the gazebo_ros_control arm to a fixed joint it behaves perfectly fine.

I was observing a similar issue with turtlebot_arm that I had originally attributed to inertial properties, but it seems like it may be a deeper seated issue.
turtlebot/turtlebot_arm/issues/22

mintar:
This commit fixes #612, but only for Gazebo9. Fixing it for Gazebo7
(the version used in ROS Kinetic) requires the following PR to be
backported to Gazebo 7 and 8:

https://bitbucket.org/osrf/gazebo/pull-requests/2814/fix-issue-2111-by-providing-options-to/diff

Once that PR has been backported, we can remove the GAZEBO_MAJOR_VERSION
guards from this PR so that the fix is active for the older Gazebo
versions as well.

Tested on Gazebo7 (where it compiles, but doesn't change anything) and
Gazebo9 (where it compiles and fixes the bug). I've tested it using the
instructions I've put into this repo:

https://github.com/mintar/mimic_joint_gazebo_tutorial

### DIFF
--- a/gazebo_ros_control/src/default_robot_hw_sim.cpp
+++ b/gazebo_ros_control/src/default_robot_hw_sim.cpp
@@ -308,10 +308,8 @@ void DefaultRobotHWSim::writeSim(ros::Time time, ros::Duration period)
       case POSITION:
 #if GAZEBO_MAJOR_VERSION >= 9
         sim_joints_[j]->SetPosition(0, joint_position_command_[j], true);
-#elif GAZEBO_MAJOR_VERSION >= 4
-        sim_joints_[j]->SetPosition(0, joint_position_command_[j]);
 #else
-        sim_joints_[j]->SetAngle(0, joint_position_command_[j]);
+        sim_joints_[j]->SetPosition(0, joint_position_command_[j]);
 #endif
         break;
 
--- a/gazebo_ros_control/src/default_robot_hw_sim.cpp
+++ b/gazebo_ros_control/src/default_robot_hw_sim.cpp
@@ -306,7 +306,9 @@ void DefaultRobotHWSim::writeSim(ros::Time time, ros::Duration period)
         break;
 
       case POSITION:
-#if GAZEBO_MAJOR_VERSION >= 4
+#if GAZEBO_MAJOR_VERSION >= 9
+        sim_joints_[j]->SetPosition(0, joint_position_command_[j], true);
+#elif GAZEBO_MAJOR_VERSION >= 4
         sim_joints_[j]->SetPosition(0, joint_position_command_[j]);
 #else
         sim_joints_[j]->SetAngle(0, joint_position_command_[j]);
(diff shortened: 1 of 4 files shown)


gazebo_ros_control/src/default_robot_hw_sim.cpp
```diff
@@ -306,7 +306,9 @@ void DefaultRobotHWSim::writeSim(ros::Time time, ros::Duration period)
         break;
 
       case POSITION:
-#if GAZEBO_MAJOR_VERSION >= 4
+#if GAZEBO_MAJOR_VERSION >= 9
+        sim_joints_[j]->SetPosition(0, joint_position_command_[j], true);
+#elif GAZEBO_MAJOR_VERSION >= 4
         sim_joints_[j]->SetPosition(0, joint_position_command_[j]);
 #else
         sim_joints_[j]->SetAngle(0, joint_position_command_[j]);
```
scpeters: kinetic requires gazebo7 so we can ditch `SetAngle` and just put the old `SetPosition` call in an `#else` block

scpeters (approved):

raequin:
Please verify that the following plan should get rid of #612.  Also, if this is an inappropriate place to ask such a question then kindly let me know where I ought to take it.

My computer currently has ros-kinetic-gazebo9-ros-control installed via apt, so here's the plan:


```
sudo apt remove --purge ros-kinetic-gazebo9-ros-control ros-kinetic-gazebo9-ros-pkgs

cd ~/catkin_ws/src

git clone https://github.com/ros-simulation/gazebo_ros_pkgs.git -b kinetic-devel

cd ~/catkin_ws/

catkin_make
```


Is there anything I need to do in order to get behavior #612 to stop or will this version of gazebo_ros_pkgs automatically take care of that?

mintar:
> if this is an inappropriate place to ask such a question then kindly let me know where I ought to take it.

It would probably have been better if you had asked this question directly at #612, but here is also okay. 

What this PR does is that it makes gazebo_ros_pkgs use a feature that was introduced in Gazebo 9. So you'll need both Gazebo 9 (from debs is okay) and the newest version of gazebo_ros_pkgs (from source, until the next release happens).

So you plan is fine, except that you should *not* remove Gazebo 9. Afterwards, the bug #612 should be fixed for you.

raequin:
Thanks again!  With that information in hand, I will simply execute the final four lines in my plan above.

edit: After installing this version of gazebo_ros_pkgs, the bug is fixed as predicted.

PR #691, opened 2018-03-20
Add warnings when the user is affected by #612

mintar:


j-rivero:
Thanks for the PR. If we want only gazebo8 or lower users to be warn, we could some preprocessor tricks:
```
#if GAZEBO_MAJOR_VERSION < 9
```

mintar:
@j-rivero : The warnings are already enclosed in the GAZEBO_MAJOR_VERSION guards, so that's already happening.

j-rivero:
> @j-rivero : The warnings are already enclosed in the GAZEBO_MAJOR_VERSION guards, so that's already happening.

arg, sorry about it, I did not check the code above. The errors are not related to this PR. Thanks for the change Martin.

PR #695, opened 2018-04-04
Strip comments from model to be spawned

azhural:
Remove comments from urdf before trying to find packages specified by "package://". Otherwise commented lines referencing non-existant packages produce a fatal error, even though the package is not used.


j-rivero:
looks good to me. Thanks!

PR #707, opened 2018-04-09
End of legacy for diff drive plugin

j-rivero:
In old versions of the code the right and left wheel were changed to fix a former code issue. To fix an old package you have to exchange left wheel by the right wheel. This PR ends with the option to set legacy in a ROS parameter.


j-rivero:

kev-the-dev (approved):
Removes all trace of this from repo and appears correct to me.

j-rivero:

PR #708, opened 2018-04-09
Remove deprecated plugins in Melodic

j-rivero:
In the list:

 * gazebo_ros_joint_trajectory: the only removed in this PR. Migration to `gazebo_ros_joint_pose_trajectory`.

 * `camera_synchronizer` and `vision_reconfigure`: I don't know if there is still someone using them. My inclination is not remove the code.

* `gazebo_ros_imu`: should be migrated to use a plugin in Gazebo but I think that we don't have a replacement at this point. We must keep it.


kev-the-dev (approved):
Users were warned of deprecation and replacement exists. Confirmed no other traces of this exist in this repo.

j-rivero:

PR #709, opened 2018-04-09
Remove legacy in gazebo_ros_control for robotNamespace

j-rivero:
Change proposed in pull request #637. Now without a legacy option in Melodic.


j-rivero:

PR #710, opened 2018-04-10
Fix model_state_test by removing the state recording

j-rivero:


kev-the-dev (approved):
Test works in either case on my machine, I assume this is for CI.

Issue #485, opened 2016-08-30
ctrl-q doesn't tear down gzserver


jackieokay:
On 16.04 with the ros-kinetic-gazebo-ros debian, whenever I use CTRL-Q to exit Gazebo launched with `rosrun gazebo_ros gazebo`, I get the following error message:

```
/opt/ros/kinetic/lib/gazebo_ros/gazebo: 36: kill: invalid signal number or name: SIGINT
```

The gzserver process remains alive in the background and I have to kill it manually.

scpeters:
The signal names seem to have changed on some OS's. We fixed OSX in #364 but haven't fixed xenial yet. I suggested the following in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/364#issuecomment-231419305 but I haven't tested it:

```
diff --git a/gazebo_ros/scripts/gazebo b/gazebo_ros/scripts/gazebo
index 392dac7..3edbf83 100755
--- a/gazebo_ros/scripts/gazebo
+++ b/gazebo_ros/scripts/gazebo
@@ -6,10 +6,8 @@ final="$@"
 final=`echo "$final" | sed 's/[^ ]*:=[^ ]* \?//g'`

 EXT=so
-SIGNAL=SIGINT
 if [ $(uname) = "Darwin" ]; then
     EXT=dylib
-    SIGNAL=INT
 fi

 # add ros path plugin if it does not already exist in the passed in arguments
@@ -40,4 +38,4 @@ GAZEBO_MASTER_URI="$desired_master_uri" gzserver $final &
 GAZEBO_MASTER_URI="$desired_master_uri" gzclient $client_final

 # Kill the server
-kill -s $SIGNAL $!
+kill -n 2 $!
```

scpeters:
fixed in #711 for lunar

PR #711, opened 2018-04-10
Use generic SIGINT parameter in kill command for gazebo script

j-rivero:
issue #485

Use the `-N` notation for send `-2` to the kill command of `gzserver`. It is supported in both Mac and Linux. I added some checks with `kill -0` to check that the process is indeed running.

### DIFF
--- a/gazebo_ros/scripts/gazebo
+++ b/gazebo_ros/scripts/gazebo
@@ -44,12 +44,12 @@ final=$(relocate_remappings "${final}")
 GAZEBO_MASTER_URI="$desired_master_uri" GAZEBO_MODEL_DATABASE_URI="$desired_model_database_uri" gzserver $final &
 gzserver_pid="$!"
 # Use kill -0 to check if the process is running
-if $(kill -0 $! > /dev/null); then
+if $(kill -0 $! 2> /dev/null); then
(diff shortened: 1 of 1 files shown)


gazebo_ros/scripts/gazebo
```diff
@@ -43,8 +41,15 @@ fi
 final=$(relocate_remappings "${final}")
 
 # Combine the commands
-GAZEBO_MASTER_URI="$desired_master_uri" GAZEBO_MODEL_DATABASE_URI="$desired_model_database_uri" gzserver $final & 
-GAZEBO_MASTER_URI="$desired_master_uri" GAZEBO_MODEL_DATABASE_URI="$desired_model_database_uri" gzclient $client_final
+GAZEBO_MASTER_URI="$desired_master_uri" GAZEBO_MODEL_DATABASE_URI="$desired_model_database_uri" gzserver $final &
+gzserver_pid="$!"
+# Use kill -0 to check if the process is running
+if $(kill -0 $! > /dev/null); then
```
kev-the-dev: Recommend redirecting stderr ```2>``` instead of stdout because kill will write to stderr if that process does not exist.

kev-the-dev: Otherwise worked great in my testing!

scpeters: good call:  b332777

scpeters (approved):

PR #714, opened 2018-05-07
Triggered camera / multicamera plugins (lunar-devel)

j-rivero:
{ port of pull request #687 }

### DIFF
--- a/SENSORS.md
+++ b/SENSORS.md
@@ -211,6 +211,15 @@ details about which code should be submitted to this repository.
    - ***example:*** gazebo_plugins/test/tricycle_drive/
               xacro, launch, world, rviz, etc.
 
+ * ***gazebo_ros_triggered_camera / gazebo_ros_triggered_multicamera***
+   - ***description:*** These camera sensors do not publish unless triggered. They have an
+     additional topic (default name `image_trigger`) that subscribes to std_msgs/Empty messages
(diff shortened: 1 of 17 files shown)


j-rivero:
@scpeters @iche033 In this port of the pull request #687 there are two failing tests (`rostest.runner.RosTest.testdistortion_pincushion_test`) instead of just one like in the `kinetic-devel`branch. Any hint about what can be happening?

scpeters:
I'm not sure. I think @mjcarroll has an idea about the distortion tests (#681), so maybe we'll figure it out soon.

scpeters:
There are two compiler warnings in the gazebo9 job; I think 14b01bc1fc018f53a96005c235c1b1b94f3cfd04 didn't get ported forward to lunar

iche033:
the previous [gazebo9 lunar build](https://build.osrfoundation.org/view/All/job/ros_gazebo9_pkgs-ci-pr_any_lunar-xenial-amd64/21/) also has two failures so it could be an existing problem?

scpeters:
~~yeah, I think the CI is ok for this~~

I've made a separate pull request to forward-port the fix for gazebo9 warnings in #728

scpeters:
I take that back, two of the distortion tests are failing because the distortion worlds don't exist on this branch. I'm not sure why, but I've added them back.

scpeters (approved):

Issue #692, opened 2018-03-22
gazebo_ros_diff_drive should add a 'enable_odom_tf' parameter


sychaichangkun:
I want to use `robot_pose_ekf` package which merges IMU, VO and Odometry and outputs the tf odom -> base_footprint.  But diff_drive plugin is always broadcasting the tf odom -> base_footprint,  these two transform conflicts.  I think there should be a param  to enable or disable the tf in diff drive plugin.

sychaichangkun:
I read the source code of gazebo_ros_diff_drive.cpp and found that in the function `GazeboRosDiffDrive::publishOdometry()`, odom tf is broadcasted and odom topic is published at the same time. There is one param `publishTf`, but it turns on/off the topic and tf altogether.
In my case I just want the /odom topic, as the input of `robot_pose_ekf`, but I dont want tf of odom->base_link, which `robot_pose_ekf`  also broadcasts.

I suggest to add two params `publishOdomTF` and `publishOdomTopic` , that would be more clear for users. And I would like to fix the bug.

KrzysztofZurad:
Hi, any progress on this one? I have the same problem to solve, need to fuse data with imu and use robot_localization.

j-rivero:
> Hi, any progress on this one? I have the same problem to solve, need to fuse data with imu and use robot_localization.

Thanks for the proposal guys, seems reasonable to me. Feel free to make the necessary changes, be sure of not changing the default behavior for the rest of users (not hard, just leave the new feature disabled by default).

kev-the-dev:
Merged. Can close issue now.

PR #716, opened 2018-05-16
Add publishOdomTF flag (#692)

KrzysztofZurad:
Add parameter to disable broadcasting odom tf information.


kev-the-dev (approved):
Useful feature with no API break as default is old behavior.

j-rivero:
thanks @KrzysztofZurad for the PR, thanks Kevin for the comment. Ready to go.

PR #718, opened 2018-05-24
ROS UTILS: prevent segfault when using alternative GazeboRos constructor

### LINKED ISSUE #478: GazeboRos constructor causes segfault
Dear developers,

I found that GazeboRos constructor for Sensor in "gazebo_ros_utils.h" causes segfault.
It's because that the member variable "rosnode_" was not initialized before using it.
- in GazeboRos( sensors::SensorPtr _parent, ... ) constructor.
- that calls ["GazeboRos::readCommonParameter()".](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h#L173)
- the member function uses "rosnode_" variable to initialize ["tf_prefix_"](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_plugins/src/gazebo_ros_utils.cpp#L87) variable.
- but the "rosnode_" variable was not initialized in the GazeboRos( sensors ... ) constructor.

I hope, it could be helpful.

kev-the-dev:
Fix #478. This line was likely omitted by accident. When users use this constructor, rosnode_ is not initialized and the program will segfault when rosnode_ is used.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h
@@ -164,6 +164,7 @@ public:
         } else {
             ss << "the 'robotNamespace' param did not exit";
         }
+        rosnode_ = boost::shared_ptr<ros::NodeHandle> ( new ros::NodeHandle ( namespace_ ) );
         info_text = plugin_ + "(ns = " + namespace_ + ")";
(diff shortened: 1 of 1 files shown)


scpeters (approved):

scpeters:
Thanks for the fix. This is a bug, so we can backport it to lunar and kinetic too.

PR #721, opened 2018-05-25
ROS UTILS: prevent segfault when using alternative GazeboRos construc…

kev-the-dev:
…tor (kinetic-devel)

Backport of #718


j-rivero:
Thanks Kevin for the PR. The `DistortionTest.barrelDistortion` test failing is not related to this PR.

PR #726, opened 2018-05-25
Fix model_state_test by removing the state recording (melodic-devel)

j-rivero:
{ port of pull request #710 }


scpeters:
I actually think the fix for this should come from #678

Issue #151, opened 2013-11-12
gazebo_ros_control: Joint limits interfaces don't work when control period != sim period


jim-rothrock:
I've been testing pull request #144 to make sure that it really fixes issue #125. I found that if Gazebo is running with a simulation period of 0.001 seconds, and gazebo_ros_control's controlPeriod parameter is set to 0.01 seconds, the front wheels of the vehicle defined in the ackermann_vehicle package steer at 1/10th the correct speed. The cause is that the joint limits interface calls in writeSim() (e.g. pj_sat_interface_.enforceLimits(period)) expect to be passed the _controller_ period (in this case, 0.01 seconds), not the simulation period (0.001 seconds). In gazebo_ros_control_plugin.cpp, the problem can be fixed like this:

```
#ifdef NOTDEF
  robot_hw_sim_->writeSim(sim_time_ros, sim_time_ros - last_write_sim_time_ros_);
#else
  robot_hw_sim_->writeSim(sim_time_ros, sim_period);
#endif
```

I'm willing to make the changes (including removing last_write_sim_time_ros_, since it would no longer be needed) and submit a pull request, but I don't know whether passing the controller period to writeSim() breaks its interface, i.e. I don't whether writeSim()'s period parameter is supposed to be the simulation period or the controller period. If no one objects to de facto defining it as the controller period, I'll make the necessary changes.

hsu:
Hi Jim,

Sorry I am a bit confused by this issue.  Did you find issue #125 not fixed by pull request #144?  I tested pull request #144 against 33eac07 where `controlPeriod` is set to `0.1`, and the _hover controller_ in that example was setting force correctly on every simulation step (`0.001` second).

On a separate note, where is the code that actually exercises `sim_period` variable in

```
robot_hw_sim_->writeSim(sim_time_ros, sim_period);
```

example?  If we decide on a convention, we should add doxygen to [`robot_hw_sim.h`](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_ros_control/include/gazebo_ros_control/robot_hw_sim.h#L79).

Thanks!
John

jim-rothrock:
> Did you find issue #125 not fixed by pull request #144?

[...]

PR #733, opened 2018-05-30
GAZEBO_ROS_CONTROL: call readSim every world update (issue #151)

kev-the-dev:
* Address issue #151 by calling simulated hardware's readSim every Update loop, not just when controller is updated.
* improve comments so this behavior is clear

I believe this would considered no API change. readSim will be called more frequently (in some cases), but the time and period parameter will still be accurate.


j-rivero (approved):
The changes looks good to me although I'm far from being an expert in the area.

j-rivero:
@ianjfrosst would you mind give a try to these changes and seems if the solve the problem reported in issue #151 ?

edenfrosst:
I can't test this code any more, as I don't have access to the code base where the issue came up. However, this is the same fix that I made to get rid of that issue.

edenfrosst (approved):

Issue #666, opened 2018-01-09
gazebo_ros_control should allow multiple RobotHW plugins


mikepurvis:
Right now you can only specify one tag, however if would be nice to take advantage of the new [combined_robot_hw](http://wiki.ros.org/combined_robot_hw) to allow specifying multiple RobotHW plugins— my use case in particular would be a simulator where I want the behaviour of the DefaultRobotHW for joints/transmissions/actuators, but I also want another plugin that loads a different type of control interface altogether that has nothing to do with joints.

Right now I manage this by having my plugin extend DefaultRobotHW, but this shouldn't be necessary.

kev-the-dev:
I was thinking of implementing this by creating ```CombinedRobotHWSim``` which will act like CombinedRobotHW but only allow loading of classes which extend RobotHWSim. Is this roughly what you meant?

bochen87:
This would be really good! Also, is it possible to then add a "blacklist" or "whitelist" of joints that should be powered / not powered by the DefaultRobotHW?

alextoind:
Any news on merging this feature?

PR #738, opened 2018-06-05
GAZEBO_ROS_CONTROL: port combined_robot_hw for gazebo_ros_control (issue #666)

kev-the-dev:
Implement feature request #666. Port an equivalent of the combined_robot_hw class which inherits from RobotHWSim to allow users to load multiple RobotHWSim plugins.


alextoind:
I've tested on my code and it works like a charm!
Thank you very much @kev-the-dev!

PR #740, opened 2018-06-06
gazebo_plugins: install triggered camera plugins

scpeters:
Fixes #739.

PR #743, opened 2018-06-07
Triggered camera / multicamera plugins (melodic-devel)

j-rivero:
{ port of pull request #714 }
{ port of pull request #687 }


scpeters:
We should include #740 as well, to make sure the new plugins are installed

kev-the-dev:
Bump to trigger rebuild (current build is 404)

kev-the-dev (approved):

Issue #127, opened 2013-10-10
merge gazebo_ros_depth_camera and gazebo_ros_openni_kinect plugins and deprecate one of them


hsu:
They started as dirty duplicate code, need to consolidate and deprecate one of them.

kev-the-dev:
Still still needs to be done. These two plugins have diverged over the years causing many issues (and a lot of maintenance overhead).

My plan is to deprecate gazebo_ros_openni_kinect, having it just print a Deprecated warning then load the gazebo_ros_depth_camera plugin. It seems that the openni_kinect plugin is used more, because it is referenced [in the tutorial](http://gazebosim.org/tutorials?tut=ros_gzplugins&cat=connect_ros#DepthCamera), but the name is less accurate.

I'll also resolve the differences between the two, mostly bringing fixes from openni_kinect into depth_camera.

It should be possible to do this in a way that doesn't break API.

PR #749, opened 2018-06-19
Depth camera plugin: Reconvert Point Cloud to original height and width

marting87:
Point cloud height and width was not reset to the correct height and width after the flat reshape.

Inherited change from _gazebo_ros_openni_kinect.cpp_.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
@@ -454,6 +454,11 @@ bool GazeboRosDepthCamera::FillPointCloudHelper(
     }
   }
 
+  // reconvert to original height and width after the flat reshape
+  point_cloud_msg.height = rows_arg;
+  point_cloud_msg.width = cols_arg;
+  point_cloud_msg.row_step = point_cloud_msg.point_step * point_cloud_msg.width;
+
   return true;
 }
 


kev-the-dev:
This copy looks good to me. However, as we have to do this copy back and forth every time one of these changes, I like the idea proposed in #127 to get rid of one of these plugins so we can avoid these bugs in the future. I'm gonna try to do this, but I'll come back and merge this if it ends up being more complicated than it seems

kev-the-dev:
Out of curiosity (deciding which of the two to remove), do you use the depth_camera plugin or the openni_kinect and why? Thanks!

marting87:
I'm using openni_kinect since it let me receive the frame height and width directly from the point cloud. I would like to use depth_camera since the camera I'm trying to simulate is not related to the kinect camera. 

Idea:
Maybe we could keep them both and derive openni_kinect from depth_camera? Keeping all the shared code in the base class depth_camera. This ensures us we do not break anyones code which depends on openni_kinect. It also gives us the opportunity to extend the package with more cameras.

kev-the-dev:
Good idea! I did essentially that in the PR referenced above (includes your commit too). After reading through both of them, the only difference was that openni_kinnect had a max depth parameter, which I added into depth camera with a default of None to maintain current behavior. I agree it is awkward to use a plugin called openni_kinnect for platforms which don't use a kinect, so it will be deprecated in the future (for now will just warn users to migrate).

marting87:
Could we get this simple fix approved? This commit will bring both cameras closer together making #757 easier. We can continue to discuss merging of the 2 depth cameras in #757.

Issue #752, opened 2018-06-25
gazebo_ros_control_plugin - incorrect information on URDF parameter


msadowski:
Hi!

I'm on ROS kinetic (all packages installed through apt-get). And I run into a problem where I wouldn't be able to start diff_drive_controller as Gazebo was showing me an error:

```
[ERROR] [1529904002.814702511, 5.676000000]: Robot descripion couldn't be retrieved from param server.
[ERROR] [1529904002.814760134, 5.676000000]: Failed to initialize the controller
[ERROR] [1529904002.814814065, 5.676000000]: Initializing controller 'mobile_base_controller' failed
```

Even though gazebo_ros_control displays this message when it's starting:

`[ INFO] [1529903997.363384038, 0.360000000]: gazebo_ros_control plugin is waiting for model URDF in parameter [/robot_description] on the ROS param server.`

I believe it is not correct because once I remapped my namespaces so that my robot description was in parameter with /my_namespace/robot_description things started working correctly. In short, having my robot description in /robot_description wasn't working, while having it in /some_namespace/robot_description is working perfectly.

If you would like to quickly check it I made a [repo](https://github.com/msadowski/dd_bot) where I was experimenting (47b8c9d662 is the last commit, where the issue can be seen, I think).

kev-the-dev:
Looks like you specify ```<robotNamespace>``` in your gazebo_ros_control plugin definition. This is the namespace robot_description is expected to be in.  You can also override this by setting ```<robotParam>``` to the parameter you set your robot description to.

msadowski:
@ironmig That's what I thought. I'm wondering whether it's possible to include the <robotNamespace> parameter in the message visible from Gazebo? I think this would make it much clearer

kev-the-dev:
So after looking at the code the behavior is to search for ```<robotParam>``` (default ```robot_description```) within ```<robotNamespace>``` _or any parent namespaces_, waiting until it is found. IMO the behavior should simply be to get the param from ```robot_namespace/robot_description``` or error and close. 

I submitted the PR referenced above to make the logging more verbose/accurate while preserving the current behavior for compatibility for other users. Thanks for noticing the issue!

Now prints:
```[ WARN] [1530126972.799640061, 0.097000000]: Waiting for parameter [test] in namespace [/test_namespace] or parent namespaces.```

PR #754, opened 2018-06-27
GAZEBO_ROS_CONTROL: more verbose logging for missing URDF parameter

kev-the-dev:
* Change log level on missing robot_description parameter to WARNING instead of INFO as it blocks the plugin from running and users may not see why.
* Change logged messages in this condition to more accurately describe what the plugin is doing and how to set the parameter correctly

The behavior other than the logging is the same.

Addresses issue #752

### DIFF
--- a/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp
+++ b/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp
@@ -268,29 +268,36 @@ std::string GazeboRosControlPlugin::getURDF(std::string param_name) const
 {
   std::string urdf_string;
 
-  // search and wait for robot_description on param server
-  while (urdf_string.empty())
+  // Search for param_name in the robot namespace and parent namespaces, waiting until it is found
+  std::string search_param_name;
+  while (true)
   {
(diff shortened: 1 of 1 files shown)


gazebo_ros_control/src/gazebo_ros_control_plugin.cpp
```diff
@@ -268,29 +268,36 @@ std::string GazeboRosControlPlugin::getURDF(std::string param_name) const
...
+    {
+      ROS_WARN_ONCE_NAMED("gazebo_ros_control", "Waiting until parameter [%s] is not an empty string",
+        search_param_name.c_str());
+      continue;
+    }
+    else
+      break;
 
+    // Sleep 0.1 seconds to allow parameter to be set
     usleep(100000);
```
chapulina, line 297: Since all conditions above have either `continue` or `break`, I think we'll never reach this sleep.

chapulina (changes requested):
The change to warning and the extra checks look good to me. I just have the comment below about sleep, and also note that this needs to be rebased to resolve conflicts.

Issue #484, opened 2016-08-30
[gazebo_plugins] RGB channel flipped with <image><format>R8G8B8</format></image> with openni_camera plugin


wkentaro:
Do you have idea how to fix this? Bug in gazebo library?

```
<image>
  <format>R8G8B8</format>
</image>
```

![format_rgb](https://cloud.githubusercontent.com/assets/4310419/18098221/45b8d20c-6f1c-11e6-9c5d-931dd16bd393.png)

```
<image>
  <format>B8G8R8</format>
</image>
```

![format_bgr](https://cloud.githubusercontent.com/assets/4310419/18098251/5fcc75b8-6f1c-11e6-9fd3-0df6ed1e7081.png)

iche033:
[related issue](https://bitbucket.org/osrf/gazebo/issues/1644/rgb-channel-swapping-in-ros-visualization) in gazebo

scpeters:
@iche033 beat me to it, I was just about to link to that issue.

kev-the-dev:
This can still be reproduced in kinetic/gazebo7 (and possibly other versions). Note that the point cloud from the openni_kinect plugin is generated by this plugin (not gazebo) based on the depth image. Looking at [the code](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_openni_kinect.cpp#L361), it is indeed hard coded to assume the image is RGB. I'll submit a PR to look at the image format and form the point cloud accordingly.

shmpwk:
Is this issue solved?
I'm having similar problem on `libgazebo_ros_depth_camera.so` on ROS melodic + gazebo-plugins `2.8.7` 

-  `B8G8R8`
![Screenshot from 2020-08-10 05-22-00](https://user-images.githubusercontent.com/42209144/89741151-da103b80-dac9-11ea-99dd-a931185334fe.png)

- `R8G8B8`
![Screenshot from 2020-08-10 05-23-59](https://user-images.githubusercontent.com/42209144/89741155-e7c5c100-dac9-11ea-843b-6df39d4c3bf2.png)

PR #757, opened 2018-06-29
Deprecate openni_kinect in favor of depth_camera

kev-the-dev:
For several years there has been two nearly identical plugins, openni_kinnect and depth_camera. This is confusing for users and had lead to several bugs from not porting changes in one to the other. 

This PR makes openni_kinect inherit from depth_camera and print a deprecation warning on load so users will switch to depth_camera. Some fixes/features from openni_kinnect are also ported into depth_camera. 

The API for using both plugins should remain the same, but ABI for both had to be broken (this isn't very important as they are dynamically loaded anyways). 

Also added a test for openni_kinnect to ensure it still works

Implements #127 
Replaces #749 
Fixes #569
Replaces #570

PR #759, opened 2018-07-02
DEPTH CAMERA: fix pointcloud for BGR/mono image encodings (issue 484)

kev-the-dev:
**Note: branched from #757, which should be merged first**

Fix issue #484. Depth camera plugin will now look at image encoding and form the point cloud accordingly, supporting bgr8, rgb8, and mono8, and defaulting to no color/intensity fields for unrecognized encoding formats.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
@@ -455,6 +455,11 @@ bool GazeboRosDepthCamera::FillPointCloudHelper(
     }
   }
 
+  // reconvert to original height and width after the flat reshape
+  point_cloud_msg.height = rows_arg;
+  point_cloud_msg.width = cols_arg;
+  point_cloud_msg.row_step = point_cloud_msg.point_step * point_cloud_msg.width;
+
   return true;
 }
 
--- a/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
@@ -353,13 +353,15 @@ void GazeboRosDepthCamera::FillPointdCloud(const float *_src)
   this->point_cloud_msg_.row_step = this->point_cloud_msg_.point_step * this->width;
 
   ///copy from depth to point cloud message
-  FillPointCloudHelper(this->point_cloud_msg_,
-                 this->height,
-                 this->width,
-                 this->skip_,
-                 (void*)_src );
+  bool success = FillPointCloudHelper(this->point_cloud_msg_,
+                                      this->height,
+                                      this->width,
+                                      this->skip_,
+                                      (void*)_src );
 
-  this->point_cloud_pub_.publish(this->point_cloud_msg_);
+  // If filling point cloud failed, don't publish it (user has already been warned)
+  if (success)
+    this->point_cloud_pub_.publish(this->point_cloud_msg_);
 
   this->lock_.unlock();
 }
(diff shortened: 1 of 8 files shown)


gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -303,34 +310,43 @@ void GazeboRosDepthCamera::OnNewImageFrame(const unsigned char *_image,
   if (!this->initialized_ || this->height_ <=0 || this->width_ <=0)
     return;
 
+<<<<<<< HEAD
   //ROS_ERROR_NAMED("depth_camera", "camera_ new frame %s %s",this->parentSensor_->GetName().c_str(),this->frame_name_.c_str());
 # if GAZEBO_MAJOR_VERSION >= 7
+=======
+>>>>>>> abf673f... GAZEBO_PLUGINS: Deprecate openni_kinect
```
marting87: Something goes wrong here, left overs from merge

marting87 (commented):
I don't think OnNewRGBPointCloud is used anymore. Maybe we can get rid of it?

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -381,14 +397,73 @@ bool GazeboRosDepthCamera::FillPointCloudHelper(
     uint32_t rows_arg, uint32_t cols_arg,
     uint32_t step_arg, void* data_arg)
 {
+  // Skip if color image has not been filled yet
+  if (this->image_msg_.data.empty())
+  {
+    ROS_DEBUG_STREAM_NAMED("depth_camera", "Image is empty, skipping point cloud generation. "
+                           << "This should only happen once at the beginning");
+    return false;
```
marting87, line 403: The very first time false is returned, this it not checked by FillPointdCloud.
```

  ///copy from depth to point cloud message
   FillPointCloudHelper(this->point_cloud_msg_,
                  this->height,
                  this->width,
                  this->skip_,
                  (void*)_src );
 
    this->point_cloud_pub_.publish(this->point_cloud_msg_);
```
This results in publishing an empty point cloud with no fields in the beginning. Maybe change to:
```
///copy from depth to point cloud message
   if(FillPointCloudHelper(this->point_cloud_msg_,
                  this->height,
                  this->width,
                  this->skip_,
                  (void*)_src )){
       this->point_cloud_pub_.publish(this->point_cloud_msg_);
   }
```

marting87:
I'm curious what your opinions are on the encoding of the fields in the point cloud. 

For example mono8 results in **float x, float y, float z, uint8_t intensity**. 

I believe it is more common to output intensity as a float, to comply with **pcl::PointXYZI**

kev-the-dev:
> I don't think OnNewRGBPointCloud is used anymore. Maybe we can get rid of it?

It is not well documented, but if the ```<output>points<output>``` is set in the camera sensor, like in [this example](https://bitbucket.org/osrf/gazebo/src/01c7f8b1d68448bc618b575ad1c7ec13fee2b87f/test/worlds/pointcloud_camera.world#lines-42), OnNewRGBPointCloud will be called _instead of_ OnNewDepthImage.

> I believe it is more common to output intensity as a float, to comply with pcl::PointXYZI

It does seem that intensity is more common as a float, I'll switch it

marting87:
Any updates? Would love to see this code committed.

kev-the-dev:
This can be merged once 757 is merged, which will happen once another maintainer approves of it. Hopefully in a few weeks. In the mean time, you can always download this branch into your workspace and compile it from source to get this feature.

PR #760, opened 2018-07-02
RST-410: More realistic bump sensor simulation

matthewbwhitaker:
Addressing [RST-410](https://locusrobotics.atlassian.net/browse/RST-410)

I _think_ this should function in the same way as the real bump topic:

- Only publish the boolean bump data to the topic when the bump status changes.
- Make the bump data topic latching, as per the real system.

Added and activated using a configurable locusBumpTopicName URDF element to switch-on and name the Locus bump data topic (associated PR in locus_bots here: ).


matthewbwhitaker:
Urgh, those conflicts look fun. Will resolve and update PR.

PR #761, opened 2018-07-03
Add auto_distortion parameter to camera utils

kev-the-dev:
For users who chose to model distortion in their cameras, they have to specify the parameters twice in their SDF like in this example:

```
        <sensor type="camera" name="camera_distorted">
          <update_rate>30.0</update_rate>
          <camera name="head">
              ...
            <distortion>
              <k1>0.1</k1>
              <k2>0.1</k2>
              <p1>0</p1>
              <p2>0</p2>
              <k3>0.1</k3>
              <center>0.5 0.5</center>
            </distortion>
          </camera>
          <plugin name="camera_controller" filename="libgazebo_ros_camera.so">
             ...
            <distortionK1>0.1</distortionK1>
            <distortionK2>0.1</distortionK2>
            <distortionT1>0.0</distortionT1>
            <distortionT2>0.0</distortionT2>
            <distortionK3>0.1</distortionK3>
            <borderCrop>false</borderCrop>
          </plugin>
        </sensor>
```

This adds an sdf paramter ```<autoDistortion>``` which will set the distortion coefficents published with camera info to be the same as the gazebo sensor, so users can do this instead:

```
        <sensor type="camera" name="camera_distorted">
          <update_rate>30.0</update_rate>
          <camera name="head">
              ...
            <distortion>
              <k1>0.1</k1>
              <k2>0.1</k2>
              <p1>0</p1>
              <p2>0</p2>
              <k3>0.1</k3>
              <center>0.5 0.5</center>
            </distortion>
          </camera>
          <plugin name="camera_controller" filename="libgazebo_ros_camera.so">
             ...
             <autoDistortion>true</autoDistortion>
          </plugin>
        </sensor>
```

This should probably be the default to pull the coefficients from the sensor (like width, height, etc), but for backwards compatibility this parameter is false by default.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
@@ -157,6 +157,7 @@ namespace gazebo
     protected: double distortion_t1_;
     protected: double distortion_t2_;
 
+    protected: bool auto_distortion_;
     protected: bool border_crop_;
 
     protected: boost::shared_ptr<camera_info_manager::CameraInfoManager> camera_info_manager_;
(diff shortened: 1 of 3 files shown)


j-rivero (approved):
I think that is a good improvement over the current situation. Thanks Kevin.

PR #762, opened 2018-07-04
Fix handling of boolean values

tik0:
since Gazebo API returns 'true'/'false' as '1'/'0' strings

Signed-off-by: Timo Korthals <tkorthals@cit-ec.uni-bielefeld.de>

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_utils.cpp
+++ b/gazebo_plugins/src/gazebo_ros_utils.cpp
@@ -98,7 +98,7 @@ void GazeboRos ::getParameterBoolean(bool &_value, const char *_tag_name, const 
     _value = _default;
     if (!sdf_->HasElement(_tag_name)) {
         ROS_WARN_NAMED("utils", "%s: missing <%s> default is %s",
-                 info(), _tag_name,  (_default?"ture":"false"));
+                 info(), _tag_name,  (_default?"true":"false"));
     } else {
         getParameterBoolean(_value, _tag_name);
     }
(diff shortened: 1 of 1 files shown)


tik0:
I am wondering about the test results. Is this sufficient? Because it seems not related to my pull request.

j-rivero:
> I am wondering about the test results. Is this sufficient? Because it seems not related to my pull request.

It is not related , a known bug that @ironmig has been fixing in upstream Gazebo.

j-rivero (approved):
The acceptance of "1" and "0" values seems reasonable to me.

tik0:
Thanks. It is a bit wired that this bug appears first in melodic, so I think the Gazebo guys messed around with the parsing in their last major upgrade.

gazebo_plugins/src/gazebo_ros_utils.cpp
```diff
@@ -108,22 +108,22 @@ void GazeboRos ::getParameterBoolean(bool &_value, const char *_tag_name) {
...
-        else if(boost::iequals(value, std::string("false")))
+        else if(boost::iequals(value, std::string("false")) || boost::iequals(value, std::string("0")))
         {
             _value = false;
         }
         else
         {
-            ROS_WARN_NAMED("utils", "%s: <%s> must be either true or false",
-                     info(), _tag_name);
+            ROS_WARN_NAMED("utils", "%s: <%s> must be either true or false but is '%s'",
```
kev-the-dev, line 121: Warning should be updated to reflect new options
```"%s: <%s> must be either true/1 or false/0 but is '%s'"```

tik0: Just to make it clear: My intention with this PR is not to introduce a new option, but to introduce the new behavior of the Gazebo API (melodic/gazebo9) to the evaluation of boolean strings in sdf. 'true'/'false' in an sdf are evaluated as '1'/'0' by the function `sdf_->GetElement(_tag_name)->Get<std::string>();`.
Just try yourself by e.g. setting up the diffkin plugin. You cannot configure it anymore without my PR.

kev-the-dev:
Seems like this was changed in upstream SDF. Thanks for the fix!

tik0:
Just one off-topic question: Do you know how long it takes for the PR to be actually included into the dep packages of the ROS/Ubuntu ppa (s.t. release cycle)? I am using the plugin in some robot for lessons at my uni.

kev-the-dev:
There is not a fixed time span for releases, they are done manually. I'll lobby to get a new release for melodic next week so this fix can be distributed. Then is is usually a couple more weeks until the release is synced to the debs. Of course, you can always clone the repo into your catkin workspace and build it

PR #374, opened 2015-12-17
Added option to specify x and y components of camera focal length.

carlosmccosta:

PR #767, opened 2018-07-11
Addition of fx and fy camera intrinsics

carlosmccosta:
Added the option to specify the X and Y components of the focal length of a gazebo camera.
This PR is the update of #374 for ROS Melodic in Ubuntu 18.04.

Let me know if I need to perform changes / improvements.

Related PRs:
- https://bitbucket.org/osrf/gazebo/pull-requests/2058/added-camera-intrinsics-fx-fy-cx-cy-s
- https://bitbucket.org/osrf/sdformat/pull-requests/248/added-camera-intrinsics-fx-fy-cx-cy-s

Testing world:
- https://github.com/inesc-tec-robotics/gazebo_projection_mapping/blob/hydro-devel/worlds/camera_intrinsics.world

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
@@ -149,7 +149,8 @@ namespace gazebo
     protected: double cx_prime_;
     protected: double cx_;
     protected: double cy_;
-    protected: double focal_length_;
+    protected: double focal_length_x_;
+    protected: double focal_length_y_;
     protected: double hack_baseline_;
     protected: double distortion_k1_;
     protected: double distortion_k2_;
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
@@ -152,6 +152,7 @@ namespace gazebo
     protected: double focal_length_ ROS_DEPRECATED;
     protected: double focal_length_x_;
     protected: double focal_length_y_;
+    protected: double axis_skew_;
     protected: double hack_baseline_;
     protected: double distortion_k1_;
     protected: double distortion_k2_;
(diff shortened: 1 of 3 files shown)


christian-rauch:
This seems to be a reasonable addition to the camera plugin.
Could one of the maintainers have a look at this?

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
```diff
@@ -149,7 +149,8 @@ namespace gazebo
     protected: double cx_prime_;
     protected: double cx_;
     protected: double cy_;
-    protected: double focal_length_;
```
chapulina, line 152: This PR is tagetting Melodic, and we can't remove protected members in a released version because it would break downstream users.

gazebo_plugins/src/gazebo_ros_camera_utils.cpp
```diff
@@ -518,22 +539,23 @@ void GazeboRosCameraUtils::Init()
...
   {
-    this->focal_length_ = computed_focal_length;
+    this->focal_length_x_ = computed_focal_length;
+    this->focal_length_y_ = computed_focal_length;
   }
   else
   {
     // check against float precision
-    if (!ignition::math::equal(this->focal_length_, computed_focal_length))
+    if (!ignition::math::equal(this->focal_length_x_, computed_focal_length))
```
chapulina: Should `focal_length_y_` also be checked?

chapulina (changes requested):
LGTM, we just need to make sure it's backwards compatible. A test would also be nice.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
carlosmccosta: Good afternoon :)
I added back the focal_length_ as deprecated and made it backwards compatible (it was computed from HFOV, so it is the same as focal_length_x_).

gazebo_plugins/src/gazebo_ros_camera_utils.cpp
carlosmccosta: Good afternoon :)
I added the checking of focal_length_y_ as suggested.

carlosmccosta:
Good afternoon,
Besides the focal lengths and principal point, I also added the setting of the axis skew.

peci1:
Hi, what about moving on with this? What is still needed?

PR #770, opened 2018-07-17
ROS2: Port gazebo_dev to ament package

kev-the-dev:
Convert gazebo_dev into an ament/colcon compatible package. This will allow for gazebo_ros and gazebo_plugins to depend on gazebo_dev to get headers/libraries from gazebo.

In package.xml, the dependency on gazebo is commented out with a TODO because colcon and ament fail to find gazebo on my system (though the cmake code works fine).


gazebo_dev/package.xml
```diff
@@ -17,8 +18,15 @@
```
chapulina, line 18: Sorry, I just realized the url lines have been removed for `gazebo_msgs`, mind bringing them back?

kev-the-dev: 94aba73

gazebo_dev/package.xml
```diff
@@ -17,8 +18,15 @@
 
   <author>Johannes Meyer</author>
 
-  <buildtool_depend>catkin</buildtool_depend>
+  <buildtool_depend>ament_cmake</buildtool_depend>
 
+<!-- Currently causes colcon/ament to fail even when they are installed
+     TODO: fix and uncomment
   <build_export_depend>libgazebo9-dev</build_export_depend>
   <exec_depend>gazebo9</exec_depend>
```
chapulina: @mikaelarguedas just helped me debug this a bit and for me, upgrading the cmake minimum version to 3.5 solves the issue. Could you give that a try?

`cmake_minimum_required(VERSION 3.5)`

kev-the-dev: Worked!

chapulina (approved):

PR #771, opened 2018-07-17
ROS2: Add base class for Node in gazebo plugins

kev-the-dev:
Adds a class gazebo_ros::Node, which is a child of rclcpp::Node but manages a rclcpp::exeuctor among all it's instances. This allows gazebo plugins to create ROS nodes without worrying about setting up a thread to spin, like in the ros1 version of this package. It also handles ROS initialization and shutdown.


chapulina:
Do you know how github figured out that a file had been renamed and displayed the appropriate diff [here](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/770/files#diff-ddcaebb8ba7211f67158911619630222)? It [seems](https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv) that `git mv` wouldn't help with that... I'm asking because it would be more convenient to review the diffs that way...

gazebo_ros/include/gazebo_ros/executor.hpp
```diff
@@ -0,0 +1,52 @@
...
+#ifndef GAZEBO_ROS_EXECUTOR_HPP
+#define GAZEBO_ROS_EXECUTOR_HPP
+
+#include <rclcpp/rclcpp.hpp>
+#include <gazebo/common/common.hh>
+
+namespace gazebo_ros
+{
+
+/// Executor run in a seperate thread to handle events from all #gazebo_ros::Node instances
```
chapulina: sep**a**rate

gazebo_ros/include/gazebo_ros/executor.hpp
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * \class Executor executor.hpp <gazebo_ros/executor.hpp>
+ */
+class Executor : public rclcpp::executors::MultiThreadedExecutor
+{
+public:
+  /// Create an instance and start the internal thread
+  Executor();
+
+  /// Stops the internal exeuctor and joins with the internal thread
```
chapulina: exe**cu**tor

gazebo_ros/include/gazebo_ros/executor.hpp
```diff
@@ -0,0 +1,52 @@
...
+  virtual ~Executor();
+
+private:
+  /// Thread where the exeuctor spins until destruction
+  std::thread spin_thread_;
```
chapulina: Style nitpick: members should come after functions, see [guide](https://google.github.io/styleguide/cppguide.html#Declaration_Order)

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -0,0 +1,153 @@
...
+  /// Destructor
+  virtual ~Node();
+
+  /// Create a #gazebo_ros::Node and add it to the global #gazebo_ros::Executor.
+  /**
+   * \note Must NOT called until #InitROS has been called
+   * \param[in] node_name Name for the new node to create
+   * \return A shared pointer to a new #gazebo_ros::Node
+   */
+  static SharedPtr Create(const std::string& node_name);
```
chapulina: I *believe* the pointer style rule [here](https://github.com/ros2/ros2/wiki/Developer-Guide#pointer-syntax-alignment) also applies in this case, i.e. `std::string & node_name`. I also looked through existing ROS2 code and that seems to be the case.

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -0,0 +1,153 @@
...
+  static std::mutex lock_;
+
+  /// Points to an #gazebo_ros::Executor shared between all #gazebo_ros::Node instances
+  static std::weak_ptr<Executor> static_executor_;
+};
+
+template <typename ...Args> 
+Node::SharedPtr Node::Create(Args && ...args)
+{
+  // Throw exception is Node is created before ROS is initialized
```
chapulina: is Node -> if Node?

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -0,0 +1,153 @@
...
+    node->executor_ = std::make_shared<Executor>();
+    static_executor_ = node->executor_;
+  }
+
+  // Add new node to the exeuctor so its callbacks are called
+  node->executor_->add_node(node);
+  return node;
+}
+
+
```
chapulina: Nit: 2 blank lines

chapulina (commented):
Style and documentation comments

chapulina:
How do you recommend this PR to be tested manually?

Also, ideally we would start writing automated tests for all new functionality. This PR seems like a good place to start.

kev-the-dev:
I have an example plugin using this in ironmig/gazebo_ros_pkgs#2, I'll try to convert that into a test

chapulina:
@ironmig , heads up, while you work on a test, I'm [enabling linters](https://github.com/ros2/ros2/wiki/Migration-Guide#linters) and fixing linter issues :wink:

chapulina:
Check out https://github.com/ros-simulation/gazebo_ros_pkgs/commit/dee249ee51e63eb10df6160fa28a5908985b4bff for the linters, feel free to pull the changes into this PR.

gazebo_ros/include/gazebo_ros/node.hpp
chapulina: Addressed by adding the linters

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  // Contruct Node by forwarding arguments
+  Node::SharedPtr node = std::make_shared<Node>(std::forward<Args>(args) ...);
+
+  std::lock_guard<std::mutex> l(lock_);
+  // Store shared pointer to static executor in this object
+  node->executor_ = static_executor_.lock();
+  // If executor has not been contructed yet, do so now
+  if (!node->executor_) {
```
chapulina: If I'm understanding correctly, you may want to check `if(!node->static_executor_)` here instead.

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * \note Must NOT called until #InitROS has been called
+   * \param[in] node_name Name for the new node to create
+   * \return A shared pointer to a new #gazebo_ros::Node
+   */
+  static SharedPtr Create(const std::string & node_name);
+
+  /// Create a #gazebo_ros::Node and add it to the global #gazebo_ros::Executor.
+  /**
+   * \todo Implement
```
chapulina: How about we only add the declaration once the function is implemented? It can be mentioned on the design document so that we don't lose it.

chapulina:
This is looking good, @ironmig ! I have some suggestions on https://github.com/ros-simulation/gazebo_ros_pkgs/commit/d34452a733d3f88fb788f6bbc21289b2cb575a89, could you take a look and let me know what you think? The main ideas are:

* Remove the wrapper around the `rclcpp::init` call (i.e. `Node::InitROS`) so developers can use the original function directly
* Rearranged the test directory and added a couple more test cases

kev-the-dev:
Appreciate the additional test cases. I have two thoughts on the init change:

* Do we need to worry about concurrent attempts to call rclcpp::Init? One nice thing about the wrapper is that it is mutex locked
* Perhaps we should keep it wrapped in gazebo_ros::Node until the is_initialized function is merged/released with a TODO to eventually let rclcpp track the state

chapulina:
That's a good point about concurrency. But in any case, the Node's mutex can't protect from init calls coming from outside the class, right?

I'm working right now on calling the init function from within the node without any arguments in case it hasn't been called yet and I can wrap that in a mutex. I think from our end, that's the most we can do for now. Once we move on to `is_initialized`, it should make it robust against init calls outside the class.

chapulina:
@ironmig , check out the latest version of branch `ros2-devel-node_chapulina`. The node calls init if it hasn't been called yet, that call is protected by a mutex, and I adjusted tests accordingly. I also printed a message as you suggested.

kev-the-dev:
Looks good to me, integrated your changes

chapulina (approved):
LGTM

PR #773, opened 2018-07-20
ROS2 msgs bridge mapping rule

chapulina:
Addresses https://github.com/ros2/ros1_bridge/issues/128

I haven't had a chance to try it together with ROS 1 though. @crdelsey, it would be great to get confirmation that it works. Thanks!

PR #774, opened 2018-07-20
Adding ros1_bridge mapping rules for ODEJointStates message.

ghost:
The hiStop and loStop fields in this message were renamed to
hi_stop and lo_stop to comply with ROS 2 naming rules.

This prevented the automatic mapping rules in ros1_bridge from
generating mapping code.

This change adds a manual mapping to allow the ros1_bridge to
successfully generate mappings for this message.

Related to ros-simulation/gazebo_ros_pkgs#773


chapulina (approved):
Thanks for doing it the proper way. Looks good to me.

PR #775, opened 2018-07-21
[ros2] gazebo_ros_init plugin

chapulina:
* Added the `gazebo_ros_init` system plugin, whose single responsibility is to forward command line arguments from `gzserver` to `rclcpp::init`.

* Also added a simple launch file `empty_world.launch.py`, which launches `gzserver` with the `gazebo_ros_init` plugin. It's not really useful right now, because as I understand it, launch files currently don't support arguments (see https://github.com/ros2/launch/issues/107). So you might as well call `gzserver` directly.

    You can try it like this:

   `ros2 launch gazebo_ros empty_world.launch.py`

* Added a test world plugin `ros_world_plugin` and test world `ros_world_plugin.world` to test that command line arguments passed together with `gazebo_ros_init` affect all nodes launched afterwards. 

    For example, something like this should remap the world plugin's topic:

    `GAZEBO_PLUGIN_PATH=${GAZEBO_PLUGIN_PATH}:build/gazebo_ros/test/; gzserver --verbose -s libgazebo_ros_init.so src/gazebo_ros_pkgs/gazebo_ros/test/worlds/ros_world_plugin.world ros_world_plugin:/test:=/new_test`

    And this should change the name of the node in the world plugin:

    `GAZEBO_PLUGIN_PATH=${GAZEBO_PLUGIN_PATH}:build/gazebo_ros/test/; gzserver --verbose -s libgazebo_ros_init.so src/gazebo_ros_pkgs/gazebo_ros/test/worlds/ros_world_plugin.world ros_world_plugin:__node:=new_ros_world_plugin`


chapulina:
Bah, I need to learn to rebase, my fingers are just doing mercurial stuff

chapulina:
I messed up by merging when I should have rebased (I assume) and now cleaning up is even more confusing to a Mercurial user getting used to Git :sweat_smile: See https://github.com/ros-simulation/gazebo_ros_pkgs/pull/776 instead.

PR #776, opened 2018-07-23
[ros2] gazebo_ros_init plugin

chapulina:
Redo of https://github.com/ros-simulation/gazebo_ros_pkgs/pull/775

---

* Added the `gazebo_ros_init` system plugin, whose single responsibility is to forward command line arguments from `gzserver` to `rclcpp::init`.

* Also added a simple launch file `empty_world.launch.py`, which launches `gzserver` with the `gazebo_ros_init` plugin. It's not really useful right now, because as I understand it, launch files currently don't support arguments (see https://github.com/ros2/launch/issues/107). So you might as well call `gzserver` directly.

    You can try it like this:

   `ros2 launch gazebo_ros empty_world.launch.py`

* Added a test world plugin `ros_world_plugin` and test world `ros_world_plugin.world` to test that command line arguments passed together with `gazebo_ros_init` affect all nodes launched afterwards. 

    For example, something like this should remap the world plugin's topic:

    `GAZEBO_PLUGIN_PATH=${GAZEBO_PLUGIN_PATH}:build/gazebo_ros/test/; gzserver --verbose -s libgazebo_ros_init.so src/gazebo_ros_pkgs/gazebo_ros/test/worlds/ros_world_plugin.world ros_world_plugin:/test:=/new_test`

    And this should change the name of the node in the world plugin:

    `GAZEBO_PLUGIN_PATH=${GAZEBO_PLUGIN_PATH}:build/gazebo_ros/test/; gzserver --verbose -s libgazebo_ros_init.so src/gazebo_ros_pkgs/gazebo_ros/test/worlds/ros_world_plugin.world ros_world_plugin:__node:=new_ros_world_plugin`


gazebo_ros/test/test_gazebo_ros_init.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+public:
+  void TearDown();
+  int pid_{-1};
+};
+
+void GazeboRosInitTest::TearDown()
+{
+  // If fork failed, don't need to do anything
+  if (pid_ < 0) {
```
kev-the-dev: Wonder if this test could be refactored to avoid the duplicate code for forking / subscribing / verifying messages received that is in the other tests.

gazebo_ros/test/CMakeLists.txt
```diff
@@ -22,8 +23,18 @@ foreach(plugin ${plugins})
...
 endforeach()
 
+# Worlds
+set(worlds
+  ros_world_plugin
+)
+
+foreach(world ${worlds})
+  configure_file(worlds/${world}.world ${world}.world COPYONLY)
+endforeach()
```
kev-the-dev: May be simpler to just install the worlds directory unless we plan on having non-world files in there

kev-the-dev (approved):
Looks good! A couple of optional ideas for simplicity

gazebo_ros/test/CMakeLists.txt
chapulina: You mean just configuring the whole directory (i.e. copying to the build folder) or actually installing? I agree with configuring, but I don't think it's a good idea to install them since they're specific to tests; otherwise we'd need to install the plugins too for example.

gazebo_ros/test/test_gazebo_ros_init.cpp
chapulina: Yeah you're right I'll do that. I had started writing a completely different test but then they ended up very similar so it should be possible to move this to `test_plugins`.

chapulina: Done on https://github.com/ros-simulation/gazebo_ros_pkgs/pull/776/commits/10be195185bca8015c02a391f628cfbabf944216

gazebo_ros/test/CMakeLists.txt
chapulina: Just copying the whole worlds directory to build: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/776/commits/10be195185bca8015c02a391f628cfbabf944216

Issue #780, opened 2018-07-25
[ros2] Flakey tests


kev-the-dev:
The gazebo plugin tests on the ros2 branch occasionally fail. It seems the callbacks for the ROS timers in these plugins sometimes are never called, therefore the messages are not published. This occurs both when the plugins are launched by the test or if run independently.

kev-the-dev:
I noticed that after closing some other processes on my machine, this occurs much less often. Perhaps the executor is just being starved when the system is under load.

chapulina:
I observed once that something happened that let `gzserver` keep running even after the tests were over. This made all subsequent test runs to fail because a new `gzserver` couldn't be started. I wonder if this is the same case and whether we can make sure the `gzserver` process is killed even if the test is interrupted.

dhood:
fyi @ironmig (in response to a comment that has been edited out now): if you want docker containers to not have cross-talk (for ROS 2) you can start them [like this](https://github.com/ros2/ci/pull/204)

dhood:
edit: I meant to link to the PR not the issue: https://github.com/ros2/ci/pull/204

kev-the-dev:
Thanks!

kev-the-dev:
@chapulina I think gazebo is shutting down okay. Here's one of the logs:

```
8: -- run_test.py: invoking following command in '/home/developer/workspaces/build/gazebo_ros/test':
8:  - /home/developer/workspaces/build/gazebo_ros/test/test_plugins --gtest_output=xml:/home/developer/workspaces/build/gazebo_ros/test_r
esults/gazebo_ros/test_plugins.gtest.xml  
8: [==========] Running 5 tests from 1 test case.
8: [----------] Global test environment set-up.
8: [----------] 5 tests from Plugins/TestPlugins
8: [ RUN      ] Plugins/TestPlugins.TestTopicsReceived/0
8: Gazebo multi-robot simulator, version 9.2.0              
8: Copyright (C) 2012 Open Source Robotics Foundation.
8: Released under the Apache 2 License.                           
8: http://gazebosim.org                                 
8:                                            
8: [WARN] [proper_init]: Publishing                   
8: [WARN] [proper_init]: Publishing    
8: [Msg] Waiting for master.
8: [Msg] Connected to gazebo master @ http://127.0.0.1:11345
8: [Msg] Publicized address: 172.17.0.2               
[...]

PR #781, opened 2018-07-25
Add Node::Create with sdf element

kev-the-dev:
Implement the ```Node::Create(const std::string & node_name, sdf::ElementPtr _sdf)``` method discussed in the [ros2 page](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-integration).

The only difference was I decided to parse the ```<parameter>``` and ```<argument>``` tag in the root of the element instead of within a ```<parameters>```/```<arguments>``` tag to avoid the unnecessary complexity.

Some other changes bundled in:
* Renamed the variadic ```Node::Create(Args && ... args)``` to ```Node::CreateWithArgs```. Without this change, the compiler will never use the alternative ```Node::Create``` definitions and always assume the variadic form is used.
* Move ament linting back to main CmakeList so linting is run on all source files
* Various style fixes
* Small refactor in test_plugins.cpp to do setup operations in the SetUp function as per gtest convention
* Only catch RCL_NOT_INIT exception in Node::Create, instead of any exception.
* Add larger timeouts to tests 

I've noticed that the tests are a bit flakey on my machine with and without these changes. Ticketed to #780

### DIFF
--- a/gazebo_ros/CMakeLists.txt
+++ b/gazebo_ros/CMakeLists.txt
@@ -43,6 +43,8 @@ ament_export_dependencies(gazebo_dev)
 
 if(BUILD_TESTING)
   add_subdirectory(test)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
 endif()
 
 ament_package()
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -46,19 +46,63 @@ public:
   /**
    * \details This will call rclcpp::init if it hasn't been called yet.
    * \param[in] node_name Name for the new node to create
-   * \return A shared pointer to a new #gazebo_ros::Node, or nullptr if failed to create.
+   * \return A shared pointer to a new #gazebo_ros::Node
    */
   static SharedPtr Create(const std::string & node_name);
 
   /// Create a #gazebo_ros::Node and add it to the global #gazebo_ros::Executor.
   /**
    * \details This will call rclcpp::init if it hasn't been called yet.
+   * \details Sets namespace, remappings, and parameters from SDF.
+   * SDF is in the form:
+   * \code{.xml}
+   * <!-- Optional configurations for a plugin's Node -->
+   * <ros>
+   *  <!-- Name of node, only use if plugin is one node -->
+   *  <node_name>my_node</node_name>
+   *  <!-- Namespace of the node -->
+   *  <namespace>/my_ns</namespace>
+   *  <!-- Command line arguments sent to Node's constructor for remappings -->
+   *  <argument>my_topic:=new_topic</argument>
+   *  <argument>__name:=super_cool_node</argument>
+   *  <!-- Initial ROS params set for node -->
+   *  <parameter name="max_velocity" type="int">55</parameter>
+   *  <parameter name="publish_odom" type="bool">True</parameter>
+   * </ros>
+   * \endcode
+   * \param[in] node_name Name of node to create
+   * \param[in] _sdf An SDF element in the style above or containing a <ros> tag in the style above
+   * \return A shared pointer to a new #gazebo_ros::Node
+   */
+  static SharedPtr Create(const std::string & node_name, sdf::ElementPtr _sdf);
+
+  /// Create a #gazebo_ros::Node and add it to the global #gazebo_ros::Executor.
+  /**
+   * \details This will call rclcpp::init if it hasn't been called yet.
    * \details Forwards arguments to the constructor for rclcpp::Node
    * \param[in] args List of arguments to pass to <a href="http://docs.ros2.org/latest/api/rclcpp/classrclcpp_1_1_node.html">rclcpp::Node</a>
-   * \return A shared pointer to a new #gazebo_ros::Node, or nullptr if failed to create.
+   * \return A shared pointer to a new #gazebo_ros::Node
    */
   template<typename ... Args>
-  static SharedPtr Create(Args && ... args);
+  static SharedPtr CreateWithArgs(Args && ... args);
+
+  /// Convert an sdf element to an rclcpp::Parameter
+  /* \details SDF must have a type and name attribute
+   *          where type is either 'int', 'float', 'double','bool' or 'string'.
+   *          Note that 'float' and 'double' are both stored as double.
+   * Examples:
+   * \code{.xml}
+   * <parameter type="int" name="my_int">55</parameter>
+   * <parameter type="bool" name="my_bool">false</parameter>
+   * <parameter type="double" name="my_double">13.37</parameter>
+   * <parameter type="float" name="my_float">41.18</parameter>
+   * <parameter type="string" name="my_string">Hello World</parameter>
+   * \endcode
+   * \param[in] _sdf An sdf element in the style mentioned above
+   * \return The ROS parameter with the same name, type, and value as the sdf element
+   *         On failure, the return parameter.get_type() == rclcpp::ParameterType::PARAMETER_NOT_SET
+   */
+  static rclcpp::Parameter sdf_to_ros_parameter(sdf::ElementPtr const & _sdf);
 
 private:
   /// Inherit constructor
@@ -73,10 +117,13 @@ private:
 
   /// Points to an #gazebo_ros::Executor shared between all #gazebo_ros::Node instances
   static std::weak_ptr<Executor> static_executor_;
+
+  /// Gets a logger to log information internal to the node
+  static rclcpp::Logger internal_logger();
 };
 
 template<typename ... Args>
-Node::SharedPtr Node::Create(Args && ... args)
+Node::SharedPtr Node::CreateWithArgs(Args && ... args)
 {
   std::lock_guard<std::mutex> l(lock_);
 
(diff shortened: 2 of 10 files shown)


gazebo_ros/src/node.cpp
```diff
@@ -27,9 +29,98 @@ Node::~Node()
...
+  // Set namespace if tag is present
+  if (_sdf->HasElement("namespace")) {
+    ns = _sdf->GetElement("namespace")->Get<std::string>();
+  }
+
+  // Get list of arguments from SDF
+  sdf::ElementPtr argument_sdf = _sdf->GetElement("argument");
+  while (argument_sdf) {
+    std::string argument = argument_sdf->Get<std::string>();
+    RCLCPP_WARN(internal_logger(), "Got argument %s", argument.c_str());
```
chapulina: Was the warning here for debugging purposes?

gazebo_ros/src/node.cpp
```diff
@@ -27,9 +29,98 @@ Node::~Node()
...
 {
-  return Create<const std::string &>(node_name);
+  return CreateWithArgs(node_name);
+}
+
+rclcpp::Parameter Node::sdf_to_ros_parameter(sdf::ElementPtr const & _sdf)
+{
+  if (!_sdf->HasAttribute("name")) {
+    RCLCPP_WARN(internal_logger(),
+      "Ignoring tag because it has no attribute 'name'. Tag: %s", _sdf->ToString("").c_str());
```
chapulina: Suggestion, to be more explicit, maybe say "Ignoring parameter because..."

gazebo_ros/src/node.cpp
```diff
@@ -27,9 +29,98 @@ Node::~Node()
...
+    arguments, initial_parameters);
+}
+
 Node::SharedPtr Node::Create(const std::string & node_name)
 {
-  return Create<const std::string &>(node_name);
+  return CreateWithArgs(node_name);
+}
+
+rclcpp::Parameter Node::sdf_to_ros_parameter(sdf::ElementPtr const & _sdf)
```
chapulina: I think we can remove the `_` from `_sdf`, that's usually like that because of Gazebo's style guide, but it looks out of place if that's the only input param doing it.

gazebo_ros/src/node.cpp
```diff
@@ -27,9 +29,98 @@ Node::~Node()
...
+  }
+  if (!_sdf->HasAttribute("type")) {
+    RCLCPP_WARN(internal_logger(),
+      "Ignoring tag because it has no attribute 'name'. Tag: %s", _sdf->ToString("").c_str());
+    return rclcpp::Parameter();
+  }
+
+  std::string name = _sdf->Get<std::string>("name");
+  std::string type = _sdf->Get<std::string>("type");
+  sdf::ParamPtr value = _sdf->GetValue();
```
chapulina: `value` is never used

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -46,19 +46,61 @@ class Node : public rclcpp::Node
...
-   * \return A shared pointer to a new #gazebo_ros::Node, or nullptr if failed to create.
+   * \return A shared pointer to a new #gazebo_ros::Node
    */
   template<typename ... Args>
-  static SharedPtr Create(Args && ... args);
+  static SharedPtr CreateWithArgs(Args && ... args);
+
+  /// Convert an sdf element to an rclcpp::Parameter
+  /* \details SDF must have a type and name attribute
+   *          where type is either 'int', 'float', 'bool' or 'string'
```
chapulina: Consider documenting here that  'double' is also accepted, and double and float are treated the same way.

gazebo_ros/test/plugins/sdf_node.cpp
```diff
@@ -0,0 +1,81 @@
...
+// limitations under the License.
+
+#include <gazebo/common/Plugin.hh>
+
+#include <std_msgs/msg/string.hpp>
+#include <gazebo_ros/node.hpp>
+
+#include <memory>
+
+/// Simple example of a gazebo system plugin which uses a ROS2 node with gazebo_ros::Node.
```
chapulina: system -> world (probably copy-paste error, `ros_world_plugin` is also wrong :sweat_smile: )

gazebo_ros/test/plugins/sdf_node.cpp
```diff
@@ -0,0 +1,81 @@
...
+class SDFNode : public gazebo::WorldPlugin
+{
+public:
+  /// Called by gazebo to load plugin. Creates #node, #timer_, and #pub
+  /// \param[in] argc Argument count.
+  /// \param[in] argv Argument values.
```
chapulina: different params

gazebo_ros/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ ament_export_dependencies(gazebo_dev)
 
 if(BUILD_TESTING)
   add_subdirectory(test)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
```
chapulina, line 47: Thanks for catching this! I was really suspicious that suddenly I was not making any new linting errors :smile: On my machine, I need these extra changes for all tests to pass, is the same happening to you?

```diff
diff --git a/gazebo_ros/launch/empty_world.launch.py b/gazebo_ros/launch/empty_world.launch.py
index 6d42d90..aae73c1 100644
--- a/gazebo_ros/launch/empty_world.launch.py
+++ b/gazebo_ros/launch/empty_world.launch.py
@@ -15,7 +15,7 @@
 """Launch a Gazebo server with an empty world and initialize ROS with command line arguments."""
 
 import launch
-import launch_ros.actions
+
 
 def generate_launch_description():
     # TODO(anyone): Forward command line arguments once that's supported, see
@@ -28,4 +28,3 @@ def generate_launch_description():
     return launch.LaunchDescription([
         gzserver
     ])
-
```

gazebo_ros/test/test_plugins.cpp
```diff
@@ -34,94 +34,89 @@ struct TestParams
 class TestPlugins : public ::testing::TestWithParam<TestParams>
 {
 public:
-  // Documentation inherited
+  /// Documentation inherited
```
chapulina: The reason why it was just a normal comment was to indicate that doxygen should pull the documentation from the base function if available. By changing it to `///` we're overriding the original documentation with "Documentation inherited", which is probably not what we want to do.

gazebo_ros/test/test_plugins.cpp
```diff
@@ -34,94 +34,89 @@ struct TestParams
...
+        (void) msg;
+        // If this is the first message from this topic, increment the counter
+        if (!received[i]) {
+          received[i] = true;
+          ++topics_received_from;
+        }
+      }));
   }
 
   // Wait until message is received or timeout after 5 seconds
```
chapulina: 15 seconds

gazebo_ros/src/node.cpp
```diff
@@ -27,9 +29,98 @@ Node::~Node()
...
+  } else {
+    RCLCPP_WARN(internal_logger(),
+      "Ignoring tag because attribute 'type' is invalid. Tag: %s", _sdf->ToString("").c_str());
+    return rclcpp::Parameter();
+  }
+}
+
+rclcpp::Logger Node::internal_logger()
+{
+  return rclcpp::get_logger("gazebo_ros_node");
```
chapulina, line 121: This means that all messages get printed with `gazebo_ros_node` in front instead of the namespaced node name. This could make it more difficult to identify where the messages are coming from. What's the motivation for using this common logger?

gazebo_ros/test/worlds/sdf_node_plugin.world
```diff
@@ -0,0 +1,17 @@
...
+    <plugin name="sdf_plugin" filename="./libsdf_node.so">
+      <ros>
+        <node_name>my_node</node_name>
+        <namespace>/foo</namespace>
+        <argument>test:=my_topic</argument>
+        <parameter name="my_string" type="string">str</parameter>
+        <parameter name="my_bool_true" type="bool">True</parameter>
+        <parameter name="my_bool_false" type="bool">false</parameter>
+        <parameter name="my_int" type="int">42</parameter>
+        <parameter name="my_double" type="double">13.37</parameter>
```
chapulina, line 13: Suggestion: add a malformed parameter (for example missing type) and check that the parameter is not found afterwards.

gazebo_ros/test/plugins/sdf_node.cpp
```diff
@@ -0,0 +1,81 @@
...
+  if (strcmp("my_node", node_name)) {
+    RCLCPP_ERROR(node->get_logger(), "Node name not remapped in SDF");
+    return;
+  }
+
+  // Create a publisher
+  auto pub = node->create_publisher<std_msgs::msg::String>("test");
+
+  // Attempt to get the parameters set by the SDF.
+  // If any of them fail, an exception will be thrown causing the test to fail
```
chapulina: It looks like it doesn't really throw an exception. The test times out because the plugin never publishes a message. We may want to change this comment.

chapulina (commented):
Looking good! Thanks for re-enabling the linter and refactoring the test. I wasn't able to reproduce https://github.com/ros-simulation/gazebo_ros_pkgs/issues/780, but let's keep an eye on this.

gazebo_ros/src/node.cpp
kev-the-dev: This is a ```private static``` method used internally for info/errors in Node::Create (such as when ROS is initialized, parameters cannot be parsed, etc.

chapulina: Ah gotcha :wink: Thanks!

gazebo_ros/CMakeLists.txt
kev-the-dev: Ah my flake8 was broken so it wasn't showing this... patch applied

gazebo_ros/test/worlds/sdf_node_plugin.world
kev-the-dev: added a few failure examples

kev-the-dev:
Thanks for the feedback, everything should be addressed now

chapulina (approved):
Works for me! Thanks for addressing all the comments.

PR #783, opened 2018-07-27
[ros2] Port gazebo_ros_template and add more instructions

chapulina:


gazebo_plugins/include/gazebo_plugins/gazebo_ros_template.hpp
```diff
@@ -0,0 +1,53 @@
...
+  void Load(physics::ModelPtr model, sdf::ElementPtr sdf);
+
+protected:
+  /// Optional callback to be called at every simulation iteration.
+  virtual void OnUpdate();
+
+private:
+  /// Recommended PIMPL pattern. This variable should hold all private
+  /// data members.
+  std::unique_ptr<GazeboRosTemplatePrivate> d_ptr_;
```
kev-the-dev: I personally like naming it ```impl_```, especially if we plan to put methods (not just data) in the private class

gazebo_plugins/src/gazebo_ros_template.cpp
```diff
@@ -0,0 +1,68 @@
...
+public:
+  /// Connection to world update event. Callback is called while this is alive.
+  event::ConnectionPtr update_connection_;
+
+  /// Node for ROS communication.
+  gazebo_ros::Node::SharedPtr ros_node_;
+};
+
+GazeboRosTemplate::GazeboRosTemplate()
+: d_ptr_(new GazeboRosTemplatePrivate)
```
kev-the-dev: ```make_unique``` is slightly cleaner I think. Maybe ```new``` just scares me in a modern c++ world

gazebo_plugins/CMakeLists.txt
```diff
@@ -12,11 +12,44 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
+  ${gazebo_ros_INCLUDE_DIRS}
+)
+
+# gazebo_ros_template
+add_library(gazebo_ros_template SHARED
+  src/gazebo_ros_template.cpp
+)
+target_link_libraries(gazebo_ros_template
+  ${gazebo_ros_LIBRARIES}
+)
```
kev-the-dev: I believe this is handled by ```ament_target_dependencies```.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_template.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef GAZEBO_PLUGINS__GAZEBO_ROS_TEMPLATE_HPP_
+#define GAZEBO_PLUGINS__GAZEBO_ROS_TEMPLATE_HPP_
+
+#include <gazebo/common/Plugin.hh>
+
+// For std::unique_ptr, could be removed
+#include <memory>
+
+namespace gazebo
```
kev-the-dev: Not sure if we should put all of these in the gazebo namespace like in ROS1. Perhaps put it in the gazebo_plugins namespace and declare ```using namespace gazebo```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_template.hpp
```diff
@@ -0,0 +1,53 @@
...
+{
+// Forward declaration of private data class.
+class GazeboRosTemplatePrivate;
+
+class GazeboRosTemplate : public ModelPlugin
```
kev-the-dev: Would be helpful to mention that other base classes (WorldPlugin, SensorPlugin, etc) are possible

kev-the-dev (commented):
Looks good! Have some minor comments (some of which are just my style opinions)

gazebo_plugins/include/gazebo_plugins/gazebo_ros_template.hpp
chapulina: Yeah I had the same thought, but since I couldn't think of a strong reason to change it, I just left the same to reduce the amount of things changing. One reason to keep it would be to follow the same pattern as the plugins [shipped with Gazebo](https://bitbucket.org/osrf/gazebo/src/01c7f8b1d68448bc618b575ad1c7ec13fee2b87f/plugins/?at=default), but there's little meaning if that's the only pattern they share... I'll change it to `gazebo_plugins` and we can revisit this in case someone complains.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_template.hpp
chapulina: I like that, changed https://github.com/ros-simulation/gazebo_ros_pkgs/pull/783/commits/2bc125b356bcaa39aef6c139c48e697f1f701c5e

gazebo_plugins/src/gazebo_ros_template.cpp
chapulina: making unique https://github.com/ros-simulation/gazebo_ros_pkgs/pull/783/commits/2bc125b356bcaa39aef6c139c48e697f1f701c5e

gazebo_plugins/include/gazebo_plugins/gazebo_ros_template.hpp
chapulina: done in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/783/commits/2bc125b356bcaa39aef6c139c48e697f1f701c5e

gazebo_plugins/include/gazebo_plugins/gazebo_ros_template.hpp
chapulina: Good idea, added mentions: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/783/commits/2bc125b356bcaa39aef6c139c48e697f1f701c5e

gazebo_plugins/CMakeLists.txt
chapulina: You're right, it seems to work without it. I wonder what I was doing before that I needed to add this. Removing for now: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/783/commits/c86ee0cd8b77cf161724f4c1bc78419ef62c3694

chapulina:
Thanks for the review, @ironmig ! I've addressed all your comments and also added an example world which can be tested like this:

`gzserver --verbose install/gazebo_plugins/share/gazebo_plugins/worlds/gazebo_ros_template_demo.world`

While testing it, I noticed a couple of SDF warnings coming from `node.cpp` and also fixed those.

kev-the-dev (approved):

PR #778, opened 2018-07-23
 Combine various ray plugins into one (issue 772)

kev-the-dev:
There are 4 plugins in gazebo_plugins which attach to a ray or gpu_ray sensor and are mostly duplicate code which is confusing for users and is difficult to maintain (changes in one need to be copied to the other):
* gazebo_ros_laser: Attaches to a ray sensor and publishes a sensor_msgs/LaserScan
* gazebo_ros_gpu_laser: Exact duplicate of gazebo_ros_laser but attaches to a gpu_ray
* gazebo_ros_block_laser: Attaches to a ray sensor and publishes a sensor_msgs/PointCloud
* gazebo_ros_range: Attaches to a ray sensor and publishes a sensor_msgs/Range (this one is not affect by this PR)

This PR creates a new plugin gazebo_ros_ray_sensor which can be attached to _either_ a ray or gpu_ray sensor and can publish LaserScan, PointCloud, or PointCloud2 by changing the sdf parameter ```<outputType>```. The old plugins are changed to print a deprecation warning then load this plugin with the necessary configuration to have the same behavior for backwards compatibility during the deprecation "tick".

Implements #772 
Implements #50
Fix #768

PR #785, opened 2018-07-31
[ros2] Add gazebo_ros_ray_sensor

kev-the-dev:
Add gazebo_ros_ray_sensor, a plugin to replace gazebo_ros_laser, gazebo_ros_gpu_laser, gazebo_ros_block_laser, and gazebo_ros_range. It attaches to a ray or gpu_ray sensor and publishes an output message in ROS based with the type configured via an sdf parameter.

~~The test is currently non-exhaustive and only checks the content of the range type, ensuring that that all types just publish messages. I will be adding more extensive tests.~~

raysensor


gazebo_plugins/CMakeLists.txt
```diff
@@ -61,10 +72,10 @@ ament_package()
 install(DIRECTORY include/
         DESTINATION include)
 
-install(
-  TARGETS
+install(TARGETS
     gazebo_ros_force
     gazebo_ros_template
+    gazebo_ros_ray_sensor
```
chapulina: It would be convenient to keep these in alphabetical order

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <gazebo/common/Plugin.hh>
+#include <gazebo/sensors/RaySensor.hh>
+#include <gazebo/sensors/GpuRaySensor.hh>
+#include <rclcpp/rclcpp.hpp>
+#include <sensor_msgs/msg/laser_scan.hpp>
+#include <sensor_msgs/msg/point_cloud.hpp>
+#include <sensor_msgs/msg/point_cloud2.hpp>
+#include <sensor_msgs/msg/range.hpp>
+#include <gazebo_ros/node.hpp>
+#include <boost/variant.hpp>
```
chapulina: Most of these includes can be (re)moved

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+
+class GazeboRosRaySensorPrivate;
+
+/// Plugin to attach to a gazebo ray or gpu_ray sensor and publish its data to ROS
+/**
+  \details
+  SDF parameters:
+  \verbatim
+  <outputType>: Optional. The message type of the plugin's output. Can be any of the following:
```
chapulina: It always seemed odd to me that some of the tags were `camelCase`. Both URDF and SDF use the `underscore_case` convention, and `gazebo_ros::Node`, as of now, is also using that.

So I think it would be nice to also convert all tags to this style. Thoughts?

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+  \verbatim
+  <radiationType>: The radiation type to publish when the output type is sensor_msgs/Range.
+                   Can be either "infrared" or "ultrasonic".
+                   Default: "infrared".
+  \endverbatim
+
+  Example SDF:
+  \code{.xml}
+    <plugin name="my_ray_senor_plugin" filename="libgazebo_ros_ray_sensor.so">
```
chapulina: sen**s**or

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+public:
+  /// \brief Constructor
+  GazeboRosRaySensor();
+
+  /// \brief Destructor
+  virtual ~GazeboRosRaySensor();
+
+  // Documentation Inherited
+  void Load(gazebo::sensors::SensorPtr _parent, sdf::ElementPtr _sdf);
```
chapulina, line 82: Can remove `_` prefix

gazebo_plugins/test/worlds/gazebo_ros_ray_sensor.world
```diff
@@ -0,0 +1,199 @@
...
+            <cylinder>
+              <radius>0.05</radius>
+              <length>0.5</length>
+            </cylinder>
+          </geometry>
+          <laser_retro>60.0</laser_retro>
+        </collision>
+      </link>
+    </model>
+    <light type="point" name="light_1">
```
chapulina: I'm curious why not to use the `sun` directional light.

gazebo_ros/include/gazebo_ros/conversions.hpp
```diff
@@ -65,5 +67,22 @@ geometry_msgs::msg::Vector3 Convert(const ignition::math::Vector3d & vec)
   msg.z = vec.Z();
   return msg;
 }
+
+/// Gets the base name of a gazebo scoped name
+/// \details Example: given "my_world::my_robot::my_link", returns "my_link"
+/// \param[in] str Input scoped name, see example
+/// \return Input string with all base scopes removed, see example
+std::string ScopedNameBase(const std::string & str)
```
chapulina: Not sure about this living in the `conversions` header, we could consider renaming the file something like `helpers` or moving this to a new file.

I should also mention the Gazebo issue [Add helper functions to handle scoped names](https://bitbucket.org/osrf/gazebo/issues/1735/add-helper-functions-to-handle-scoped); we could move this function to `gazebo::common` or `ignition::common` eventually.

gazebo_ros/include/gazebo_ros/test/gazebo_process.hpp
```diff
@@ -0,0 +1,112 @@
...
+  /// \note The path and --verbose are automaticaly added
+  explicit GazeboProcess(const std::vector<const char *> & args);
+
+  ~GazeboProcess();
+
+  /// Start gzserver with the arguments passed to constructuor
+  /// \return The result of fork(), either the pid of gazebo process or error if < 0
+  int run();
+
+  /// Terminate the child gzserve process
```
chapulina: gzserve**r**

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,428 @@
...
+{
+
+class GazeboRosRaySensorPrivate
+{
+public:
+  /// Node for ROS communication.
+  gazebo_ros::Node::SharedPtr rosnode_;
+
+  /// Sensor this plugin is attached to
+  boost::variant<gazebo::sensors::RaySensorPtr, gazebo::sensors::GpuRaySensorPtr> sensor_;
```
chapulina: `sensor_` is never used, we can remove the member variable

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,428 @@
...
+  using PointCloud2 = sensor_msgs::msg::PointCloud2;
+  using Range = sensor_msgs::msg::Range;
+  using LaserScanPub = rclcpp::Publisher<LaserScan>::SharedPtr;
+  using PointCloudPub = rclcpp::Publisher<PointCloud>::SharedPtr;
+  using PointCloud2Pub = rclcpp::Publisher<PointCloud2>::SharedPtr;
+  using RangePub = rclcpp::Publisher<sensor_msgs::msg::Range>::SharedPtr;
+
+  /// Publisher of output
+  /// \todo use std::variant one c++17 is supported in ROS2
+  boost::variant<LaserScanPub, PointCloudPub, PointCloud2Pub, RangePub> pub_;
```
chapulina, line 48: It won't look as elegant and the migration to `std::variant` won't be as clean, but I think it may be worth it having 4 separate publishers just to remove the reliance on `boost::variant`. @j-rivero , any thoughts?

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,428 @@
...
+    } else {
+      RCLCPP_ERROR(
+        impl_->rosnode_->get_logger(), "Invalid <radiationType> [%s]. Can be ultrasound or infared",
+        _sdf->Get<std::string>("radiationType").c_str());
+      return;
+    }
+  }
+
+  if (!_sdf->HasElement("minIntensity")) {
+    impl_->min_intensity_ = 0.0;
```
chapulina: This could be initialized at declaration

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+  /// brief Radiation type to report when output type is range
+  uint8_t range_radiation_type_;
+
+  /// Gazebo node used to subscribe to laser scan
+  gazebo::transport::NodePtr gazebo_node_;
+  /// Gazebo subscribe to parent sensor's laser scan
+  gazebo::transport::SubscriberPtr laser_scan_sub_;
+
+  int rayCount;
```
chapulina: Double-check which of these variables are actually needed and:

* remove the unused ones
* document and add `_` suffix to the used ones

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,428 @@
...
+        &GazeboRosRaySensorPrivate::PublishPointCloud2, this);
+  } else if (pub_.type() == typeid(RangePub)) {
+    laser_scan_sub_ = gazebo_node_->Subscribe(sensor_topic_,
+        &GazeboRosRaySensorPrivate::PublishRange, this);
+  } else {
+    RCLCPP_ERROR(rosnode_->get_logger(), "Publisher is an invalid type. This is an internal bug.");
+  }
+}
+
+void GazeboRosRaySensorPrivate::PublishLaserScan(ConstLaserScanStampedPtr & _msg)
```
chapulina: Do you think it makes sense to have functions like `sensor_msgs::msg::LaserScan Convert(const gazebo::msgs::LaserScanStamped & msg)` which do most of the heavy-lifting and then the plugin just sets extra stuff like `frame_id`?

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,428 @@
...
+GazeboRosRaySensor::GazeboRosRaySensor()
+: impl_(std::make_unique<GazeboRosRaySensorPrivate>())
+{
+}
+
+GazeboRosRaySensor::~GazeboRosRaySensor()
+{
+}
+
+void GazeboRosRaySensor::Load(gazebo::sensors::SensorPtr _parent, sdf::ElementPtr _sdf)
```
chapulina: Consider renaming `_parent` to `sensor`, so it is clear that it isn't the parent model or link for example.

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,428 @@
...
+  ls.time_increment = 0;
+  ls.scan_time = 0;
+  ls.range_min = _msg->scan().range_min();
+  ls.range_max = _msg->scan().range_max();
+
+  // If there are multiple vertical beams, use the one in the middle
+  size_t start = (verticalRangeCount / 2) * rangeCount;
+
+  // Copy ranges and intensities over
+  ls.ranges.resize(rangeCount);
```
chapulina: Why not use `_msg->scan().ranges_size()`?

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,428 @@
...
+  boost::get<PointCloud2Pub>(pub_)->publish(pc);
+}
+
+void GazeboRosRaySensorPrivate::PublishRange(ConstLaserScanStampedPtr & _msg)
+{
+  sensor_msgs::msg::Range range_msg;
+  range_msg.header.frame_id = frame_name_;
+  range_msg.header.stamp.sec = _msg->time().sec();
+  range_msg.header.stamp.nanosec = _msg->time().nsec();
+  range_msg.field_of_view = std::max(pDiff, yDiff);
```
chapulina: On `gazebo_ros_range` the FOV is set from the SDF but now it comes from the sensor itself. If the `<fov>` param wasn't useful, let's document it.

chapulina (changes requested):
I've done a first high level pass over the code.

There are a few changes to the math that I'd like to take a closer look at. If we want to say with some degree of confidence that users should be able to migrate from some specific old sensor type to this new one, we will need to:

* Write detailed migration guides explaining, for example:
    * what parameters changed
    * how noise can be configured to behave the same way as before
    * any expected changes in sensor output, and why

* Add some tests that show behaviour is still the same

I realize these tasks can become arbitrarily long. We need to weigh the effort of providing a robust migration path against the effort of maintaining, at least for now, 4 different plugins that do similar things but at least behave as they did in ROS 1.

Let's start a dedicated wiki page for the migration documentation: https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Ray-sensors

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
kev-the-dev: Agreed

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
kev-the-dev: I think this would be a problem if there is greater than one vertical ray as this is just a planar scan. rangeCount is the number of horizontal rays, where as _msg->scan().ranges_size() is all the rays

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
kev-the-dev: I decided to use variant because I saw it was already needed for  [sdformat](https://bitbucket.org/osrf/sdformat/src/13546ead54939168420eb770497d189459da6195/include/sdf/Param.hh#lines-219) so I believe it doesn't represent an additional dependency.

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
kev-the-dev: That would be a lot cleaner. I have to first make sure the gazebo message has everything needed to do the conversion, but this sounds like a good refactor

gazebo_plugins/test/worlds/gazebo_ros_ray_sensor.world
kev-the-dev: This was copied from the ROS1 version of this test, sun should work fine.

kev-the-dev:
Migration notes with examples [added to the wiki](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Ray-sensors). I'm going to recommit the conversion/utility functions in a separate PR than come back to address the specific issues for gazebo_ros_ray_sensor

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
chapulina: Ok sounds good. We've been trying to stop using boost, but you're right that it is already a dependency.

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,220 @@
...
+  auto pc2 = gazebo_ros::Convert<sensor_msgs::msg::PointCloud2>(*_msg, min_intensity_);
+  // Set tf frame
+  pc2.header.frame_id = frame_name_;
+  // Publish output
+  boost::get<PointCloud2Pub>(pub_)->publish(pc2);
+}
+
+void GazeboRosRaySensorPrivate::PublishRange(ConstLaserScanStampedPtr & _msg)
+{
+  // Convert Laser scan to PointCloud2
```
chapulina: `PointCloud2` -> `Range`

gazebo_plugins/test/test_gazebo_ros_ray_sensor.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+protected:
+  /// command line arguments to pass to gzserver
+  static const std::vector<const char *> args;
+  /// Wrapper for gzserver run with plugin
+  static std::unique_ptr<gazebo_ros::GazeboProcess> gazebo_process_;
+  /// ROS node used to verify output of plugin
+  static rclcpp::Node::SharedPtr node_;
+  /// ROS executor to call node_'s callbacks
+  static rclcpp::executor::Executor::SharedPtr executor_;
```
chapulina: Just curious, why are you making these static?

chapulina:
@ironmig , could you also add a demo world besides the test world? On the world, add comments explaining how to launch RViz to visualize the various sensors.

gazebo_plugins/test/test_gazebo_ros_ray_sensor.cpp
kev-the-dev: Rewrote test to actually verify correctness of output using gazebo's ground truth in 
3f117fe

kev-the-dev:
Added the test world and expanded the test to verify the correctness of the message content's using gazebo's ground truth

mikaelarguedas:

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
chapulina: This has been done

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
chapulina: I'm now leaning towards leaving this in, sorry for the roundabout, as far as I know, this doesn't conflict with any ROS2 styles, and it's convenient.

gazebo_plugins/test/worlds/gazebo_ros_ray_sensor.world
chapulina: The test has been changed to use the sun

gazebo_ros/include/gazebo_ros/conversions.hpp
chapulina: This was moved to `utils.hpp` in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/787

gazebo_ros/include/gazebo_ros/test/gazebo_process.hpp
chapulina: Fixed and code now lives in `testing_utils.hpp`

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
chapulina: It's been removed

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
chapulina: It's been cleanup

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
chapulina: Done

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
chapulina: It's been [documented](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Ray-sensors#additional-changes-3)

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
chapulina: it's fixed

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
chapulina: It's fixed

gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
chapulina: It's been addressed

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ray_sensor.hpp
chapulina: Done in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/785/commits/63a3e1f13adce497fff81a8940bff67c02ad9666

chapulina:
Thanks for the demo world, @ironmig! I added a few more examples to it on https://github.com/ros-simulation/gazebo_ros_pkgs/pull/785/commits/9c74f9c7f51188c82a5e23df461a4dc3fc41be51 and I noticed some possible problems which I need to investigate further:

* I couldn't get the `LaserScan` messages to show on RViz for either `ray` or `gpu_ray` sensors. The Status is ok, messages are received, but it always says "Showing [0] points from [0] messages". We could be forgetting to publish some field? Or maybe it's RViz's fault? I see stuff when echoing the messages.

* When I have only one vertical ray, the point cloud's position looks wrong. See how the 1-vertical-ray `gpu_ray` is way below the output for the 300-vertical-rays `ray` sensor:

gpu_ray_1vert_pc

This happens even if I put one vertical ray for the `ray` sensor, note how the intensities are properly colored but the result is also under the ground:

ray_1vert_pc

chapulina (dismissed):

chapulina:
The issue with the `LaserScan` not being visible on RViz seems to be due to https://github.com/ros2/rviz/issues/332

chapulina:
The issue with 1 vertical ray was happening if there was a `<vertical>` tag with 1 sample, but not if there was no `<vertical>` tag. So I changed the example world and also added a warning.

I think this is good to go if the CI comes clean. I'll add a note to the migration page about https://github.com/ros-simulation/gazebo_ros_pkgs/pull/778#issuecomment-412195870 by @dhood.

PR #792, opened 2018-08-07
[ros2] NoiseVariance as a pointer, ROS quaternion -> ignition conversion

kev-the-dev:
* Add conversion function going from geometry msgs quaternion to ignition quaternion
* Add a new signature for NoiseVariance, which takes a NoisePtr and return 0.0 if it is null (which indicates no noise)
* Add SensorFrameID, to standardize how the various sensor plugins select the frame to tf publish in

### DIFF
--- a/gazebo_ros/src/utils.cpp
+++ b/gazebo_ros/src/utils.cpp
@@ -49,4 +49,15 @@ std::string ScopedNameBase(const std::string & str)
   return str.substr(idx + 2);
 }
 
+std::string SensorFrameID(const gazebo::sensors::Sensor & _sensor, const sdf::Element & _sdf)
+{
+  // Return frame from sdf tag if present
+  if (_sdf.HasElement("frame_name")) {
+    return _sdf.Get<std::string>("frame_name");
+  }
+
+  // Otherwise return the unscoped parent link name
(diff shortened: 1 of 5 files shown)


gazebo_ros/src/utils.cpp
```diff
@@ -43,4 +49,15 @@ std::string ScopedNameBase(const std::string & str)
   return str.substr(idx + 2);
 }
 
+std::string SensorFrameID(const gazebo::sensors::Sensor & _sensor, const sdf::Element & _sdf)
+{
+  // Return frame from sdf tag if present
+  if (_sdf.HasElement("frame_name")) {
```
chapulina, line 55: Will this tag usually be inside `<ros>`?

chapulina (approved):
I just have a question, but this looks good to me.

gazebo_ros/src/utils.cpp
kev-the-dev: I think we should reserve `<ros>` for configuring a `gazebo_ros::Node` specifically. As this is used in the imu PR, the tag appears in the root of the plugin's SDF.

chapulina: Cool, thanks for the clarification. Let's merge this!

PR #793, opened 2018-08-07
[ros2] Port gazebo_ros_imu_sensor

kev-the-dev:
Port the gazebo_ros_imu_sensor to ROS2, following the new conventions.

A [wiki page](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-IMU-Sensors) has been created to guide users in migrating.


gazebo_plugins/include/gazebo_plugins/gazebo_ros_imu_sensor.hpp
```diff
@@ -0,0 +1,66 @@
...
+class GazeboRosImuSensor : public gazebo::SensorPlugin
+{
+public:
+  /// Constructor.
+  GazeboRosImuSensor();
+  /// Destructor.
+  virtual ~GazeboRosImuSensor();
+
+  // Documentation Inherited
+  virtual void Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _sdf);
```
dhood: could add `override`

gazebo_plugins/src/gazebo_ros_imu_sensor.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <string>
+
+namespace gazebo_plugins
+{
+
+class GazeboRosImuSensorPrivate
+{
+public:
+  /// Node for ros communication
+  gazebo_ros::Node::SharedPtr rosnode_;
```
dhood: `ros_node_` to match e.g. https://github.com/ros-simulation/gazebo_ros_pkgs/blob/4fe1967e3e6c1ee1cc176dfd253f7dd0d9ae51f9/gazebo_plugins/src/gazebo_ros_joint_state_publisher.cpp#L58

gazebo_plugins/src/gazebo_ros_imu_sensor.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+  impl_->rosnode_ = gazebo_ros::Node::Create("gazebo_ros_imu_sensor", _sdf);
+
+  impl_->sensor_ = std::dynamic_pointer_cast<gazebo::sensors::ImuSensor>(_sensor);
+  if (!impl_->sensor_) {
+    RCLCPP_ERROR(impl_->rosnode_->get_logger(), "Parent is not an imu sensor. Exiting.");
+    return;
+  }
+
+  impl_->pub_ = impl_->rosnode_->create_publisher<sensor_msgs::msg::Imu>("~/out");
```
dhood: I will say it explicitly in case anyone disagrees with me: though we have a `sensor_data` [QoS profile](https://github.com/ros2/ros2/wiki/About-Quality-of-Service-Settings#qos-profiles) I'd say that's more appropriate for subscribers. It's better for this to publish as reliable so that reliable subscribers will still be able to connect.

dhood (commented):
Just a partial review so far

gazebo_plugins/src/gazebo_ros_imu_sensor.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  // Create message to be reused
+  auto msg = std::make_shared<sensor_msgs::msg::Imu>();
```
dhood: @mikaelarguedas can you confirm that this will initialise the other elements of the covariance matrix fields to 0?
From https://github.com/ros2/geometry2/pull/64 I think the answer is yes, but http://design.ros2.org/articles/generated_interfaces_cpp.html says the default constructor causes default/value initialised, and doesn't say the default is zero initialised.

mikaelarguedas: TL;DR: yes all the fields of the array will be set to zero

Longer answer:
From the doc: `The default constructor initializes all members with their default value; if a field doesn’t have a default value, then the member is value-initialized.`

As Imu doesn't provide custom default values for its' fields (https://github.com/ros2/common_interfaces/blob/60afcc4b3256d5a0f606939a5d02e173ff15bdd9/sensor_msgs/msg/Imu.msg#L18) they will initialized with ROS' default.

These default are language independent so they are defined in the [Interface Definition](http://design.ros2.org/articles/interface_definition.html) doc.

```
A field can optionally specify a default value. If no default value is specified a common default value is used:

for bool it is false
for numeric types it is a 0 value
for string it is an empty string
for static size arrays it is an array of N elements with its fields zero-initialized
for bounded size arrays and dynamic size arrays it is an empty array []
```

The part we're interested in here is `for static size arrays it is an array of N elements with its fields zero-initialized`

dhood: mmmm I didn't realise from the design article that it would use a ROS default and not just a default specified in the .msg. I will try to update that design doc based on your answer, thanks!!

gazebo_plugins/test/worlds/gazebo_ros_imu_sensor.world
```diff
@@ -0,0 +1,42 @@
...
+          <geometry>
+            <box>
+              <size>1 1 1</size>
+            </box>
+          </geometry>
+        </visual>
+        <sensor name="my_imu" type="imu">
+          <always_on>true</always_on>
+          <update_rate>30</update_rate>
+          <imu>
```
dhood: if this empty tag is needed could you add a comment why?

mikaelarguedas:

dhood:
After speaking with ironmig offline I reworked the test to not need [the step_for_msgs function](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/793/commits/ecf2850230f9e4f09f54e454287053f7938cf091#diff-d966e67ca1b4db645730d08da8e16054R54) anymore, following [the ros laser test](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/785/files#diff-8eb9f4bbce3120fe78b26b75d8f2b81bR108) instead: it now steps forward once: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/793/files#diff-d966e67ca1b4db645730d08da8e16054R57

I also made it apply the force for multiple steps, and now the test checks the magnitude of the values.

I noticed that the linear acceleration flips between positive and negative, though this happens with the readings of the IMU directly from gazebo, it's not a conversion problem. Not sure if that happens with the gazebo_ros_imu plugin too, but it's not something that's been introduced in the porting of this plugin. Example of the values during the test:
```
9: Angular velocity: 1.443684
9: Linear acceleration: 7.542811
9: Yaw: 0.167626
9: Angular velocity: 1.808677
9: Linear acceleration: 7.469551
9: Yaw: 0.220460
9: Angular velocity: 1.583626
9: Linear acceleration: -3.116650
9: Yaw: 0.276416
9: Angular velocity: 1.948953
9: Linear acceleration: 7.004602
9: Yaw: 0.344101
9: Angular velocity: 2.277386
9: Linear acceleration: 7.697588
9: Yaw: 0.414009
9: Angular velocity: 2.024620
9: Linear acceleration: -4.554927
9: Yaw: 0.489143
9: Angular velocity: 2.360117
9: Linear acceleration: 7.322274
9: Yaw: 0.572661
9: Angular velocity: 2.605397
9: Linear acceleration: 6.003855
9: Yaw: 0.657134
9: Angular velocity: 2.964912
9: Linear acceleration: 6.983869
9: Yaw: 0.741163
9: Angular velocity: 2.702542
9: Linear acceleration: -6.040410
9: Yaw: 0.840170
```

gazebo_plugins/src/gazebo_ros_imu_sensor.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  // Get frame for message
+  msg->header.frame_id = gazebo_ros::SensorFrameID(*_sensor, *_sdf);
+
+  // Fill covariances
+  // TODO(anyone): covariance for IMU's orientation once this is added to gazebo
+  using SNT = gazebo::sensors::SensorNoiseType;
+  msg->angular_velocity_covariance[0] =
+    gazebo_ros::NoiseVariance(impl_->sensor_->Noise(SNT::IMU_ANGVEL_X_NOISE_RADIANS_PER_S));
```
dhood, line 76: I'm seeing `[Err] [Sensor.cc:454] Get noise index not valid` from these lines. looking into why.

dhood: it's caused by https://bitbucket.org/osrf/gazebo/issues/2513/ in general, but even if the noise is specified in the SDF, the error still shows in tests because the noise isn't applied via the ServerFixture: https://github.com/ros-simulation/gazebo_ros_pkgs/issues/799

dhood:
Followup work will be to prevent the noise error mentioned in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/793/files#r211036103: it's caused by https://bitbucket.org/osrf/gazebo/issues/2513/ in general, but even if the noise is specified in the SDF, the error still shows in tests because the noise isn't applied via the ServerFixture: #799. We don't have to let that block this PR because the plugin is working as expected (noise applied when specified), it just prints an unnecessary error if launched with verbose e.g.
```
gazebo ~/ros2_ws/src/ros2/gazebo_ros_pkgs/gazebo_plugins/test/worlds/gazebo_ros_imu_sensor.world --verbose
```

dhood (approved):

dhood:
@chapulina realised that the test was behaving as expected from a clean build. #799 turned out to be a non-issue, just user error.

https://github.com/ros-simulation/gazebo_ros_pkgs/pull/801 added noise to the sensor used in the test so the error doesn't show in the test output, but sensors using this plugin will still have error output if the noise isn't specified: https://bitbucket.org/osrf/gazebo/issues/2513/

Issue #779, opened 2018-07-25
[ros2] gazebo_ros_api_plugin split


chapulina:
The `gazebo_ros_api_plugin` is quite large and offers a lot of functionality all at once. While porting it to ROS 2, it would be nice to take the opportunity to also break that plugin into smaller ones so that:

* Users don't need to "pay for what they don't use"
* It is easier to remove functionality which shouldn't be present (see for example https://github.com/ros-simulation/gazebo_ros_pkgs/pull/704 where a param was added to allow disabling **all** API at once)

See the table below for a proposal of a split based on what kind of entity (model / link / joint / etc) an API targets.

I also considered a split which takes into account whether an API provides access to **ground truth** that wouldn't be available in the real world, or whether it allows altering simulation with a **god's hand**, i.e. in a way a robot wouldn't be able to in the real world. But after marking the API below with these 2 characteristics (to the best of my knowledge), it became clear to me that they really go hand-in-hand, and maybe the most useful approach would make each API disable-able within their plugins.

Another question I have is whether the new plugins should be in `gazebo_plugins` instead of `gazebo_ros`.

API | Ground truth | God's hand | Proposed category | Comments
-- | -- | -- | -- | --
apply_joint_effort |   | ✓ | `gazebo_ros_force_system` | :heavy_check_mark: #941 
clear_joint_forces |   | ✓ | `gazebo_ros_force_system` | :heavy_check_mark: #941 
get_joint_properties |   |   | `gazebo_ros_properties` | :heavy_check_mark: #868 #972 
set_joint_properties |   | ✓ | `gazebo_ros_properties` | :heavy_check_mark: #868 #972
  |   |   |   |  
clear_body_wrenches |   | ✓ | `gazebo_ros_force_system` |  :heavy_check_mark: #941 
apply_body_wrench |   | ✓ | `gazebo_ros_force_system` | :heavy_check_mark: #941
link_states | ✓ |   | `gazebo_ros_state` |  :heavy_check_mark:  #969
get_link_properties |   |   | `gazebo_ros_properties` | :heavy_check_mark: #868 #972
set_link_properties |   | ✓ | `gazebo_ros_properties` | :heavy_check_mark: #868 #972
get_link_state | ✓ |   | `gazebo_ros_state` | :heavy_check_mark: #839
set_link_state |   | ✓ | `gazebo_ros_state` | :heavy_check_mark: #839
  |   |   |   |  
[...]

PR #794, opened 2018-08-07
[ros2] Add sim_time to gazebo_ros_init

kev-the-dev:
* Add a Throttler class to utils, something that can be reused in a few places
* Add publishing sim time to /clock in gazebo_ros_init
  * The time of a simulation starts at 0.
  * Subscribers that have transient local durability specified in [their QoS settings](https://github.com/ros2/ros2/wiki/About-Quality-of-Service-Settings) will be able to receive the previously-published time message(s) even if the simulation is paused (up to the queue depth of 10).

Usage:
```
$ ros2 launch gazebo_ros empty_world.launch.py
```
```
$ ros2 topic echo /clock
clock:
  sec: 7
  nanosec: 507000000
```

Wiki: https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-ROS-Clocks-and-sim-time

### DIFF
--- a/gazebo_ros/include/gazebo_ros/gazebo_ros_init.hpp
+++ b/gazebo_ros/include/gazebo_ros/gazebo_ros_init.hpp
@@ -15,10 +15,17 @@
 #ifndef GAZEBO_ROS__GAZEBO_ROS_INIT_HPP_
 #define GAZEBO_ROS__GAZEBO_ROS_INIT_HPP_
 
-
 #include <gazebo/common/Plugin.hh>
 
+#include <memory>
+
+namespace gazebo_ros
+{
+
+class GazeboRosInitPrivate;
+
 /// Initializes ROS with the system arguments passed to Gazebo (i.e. calls rclcpp::init).
+/// Also publishes the latest simtime to /clock
 class GazeboRosInit : public gazebo::SystemPlugin
 {
 public:
@@ -28,9 +35,12 @@ public:
   /// Destructor
   virtual ~GazeboRosInit();
 
-  /// Called by Gazebo to load plugin.
-  /// \param[in] argc Argument count.
-  /// \param[in] argv Argument values.
+  // Documentation inherited
   void Load(int argc, char ** argv);
+
+private:
+  std::unique_ptr<GazeboRosInitPrivate> impl_;
 };
+
+}  // namespace gazebo_ros
 #endif  // GAZEBO_ROS__GAZEBO_ROS_INIT_HPP_
--- a/gazebo_ros/include/gazebo_ros/gazebo_ros_init.hpp
+++ b/gazebo_ros/include/gazebo_ros/gazebo_ros_init.hpp
@@ -36,7 +36,7 @@ public:
   virtual ~GazeboRosInit();
 
   // Documentation inherited
-  void Load(int argc, char ** argv);
+  void Load(int argc, char ** argv) override;
 
 private:
   std::unique_ptr<GazeboRosInitPrivate> impl_;
--- a/gazebo_ros/src/gazebo_ros_init.cpp
+++ b/gazebo_ros/src/gazebo_ros_init.cpp
@@ -29,11 +29,11 @@ class GazeboRosInitPrivate
 {
 public:
   GazeboRosInitPrivate();
-  gazebo_ros::Node::SharedPtr rosnode_;
+  gazebo_ros::Node::SharedPtr ros_node_;
   rclcpp::Publisher<rosgraph_msgs::msg::Clock>::SharedPtr clock_pub_;
   gazebo::event::ConnectionPtr world_update_event_;
   gazebo_ros::Throttler throttler_;
-  static const gazebo::common::Time DEFAULT_PUBLISH_PERIOD;
+  static constexpr double DEFAULT_PUBLISH_FREQUENCY = 10.;
 
   void PublishSimTime(const gazebo::common::UpdateInfo & _info);
 };
@@ -52,21 +52,21 @@ void GazeboRosInit::Load(int argc, char ** argv)
   // Initialize ROS with arguments
   rclcpp::init(argc, argv);
 
-  impl_->rosnode_ = gazebo_ros::Node::Create("gazebo_ros_clock");
-  impl_->clock_pub_ = impl_->rosnode_->create_publisher<rosgraph_msgs::msg::Clock>("/clock");
+  impl_->ros_node_ = gazebo_ros::Node::Create("gazebo_ros_clock");
+  impl_->clock_pub_ = impl_->ros_node_->create_publisher<rosgraph_msgs::msg::Clock>("/clock");
 
   // Get publish rate from parameter if set
   rclcpp::Parameter rate_param;
-  if (impl_->rosnode_->get_parameter("publish_rate", rate_param)) {
+  if (impl_->ros_node_->get_parameter("publish_rate", rate_param)) {
     if (rclcpp::ParameterType::PARAMETER_DOUBLE == rate_param.get_type()) {
-      impl_->throttler_ = Throttler(gazebo::common::Time(1.0 / rate_param.as_double()));
-    } else if (rclcpp::ParameterType::PARAMETER_INTEGER != rate_param.get_type()) {
-      impl_->throttler_ = Throttler(gazebo::common::Time(1.0 / rate_param.as_int()));
+      impl_->throttler_ = Throttler(rate_param.as_double());
+    } else if (rclcpp::ParameterType::PARAMETER_INTEGER == rate_param.get_type()) {
+      impl_->throttler_ = Throttler(rate_param.as_int());
     } else {
-      RCLCPP_WARN(impl_->rosnode_->get_logger(),
+      RCLCPP_WARN(impl_->ros_node_->get_logger(),
         "Could not read value of param publish_rate [%s] as double/int, using default %ghz.",
         rate_param.value_to_string().c_str(),
-        1.0 / GazeboRosInitPrivate::DEFAULT_PUBLISH_PERIOD.Double());
+        GazeboRosInitPrivate::DEFAULT_PUBLISH_FREQUENCY);
     }
   }
 
@@ -74,12 +74,8 @@ void GazeboRosInit::Load(int argc, char ** argv)
     std::bind(&GazeboRosInitPrivate::PublishSimTime, impl_.get(), std::placeholders::_1));
 }
 
-// By default publish at 10 HZ (100 millisecond period)
-const gazebo::common::Time GazeboRosInitPrivate::DEFAULT_PUBLISH_PERIOD = gazebo::common::Time(0,
-    1E8);
-
 GazeboRosInitPrivate::GazeboRosInitPrivate()
-: throttler_(DEFAULT_PUBLISH_PERIOD)
+: throttler_(DEFAULT_PUBLISH_FREQUENCY)
 {
 }
 
--- a/gazebo_ros/src/gazebo_ros_init.cpp
+++ b/gazebo_ros/src/gazebo_ros_init.cpp
@@ -14,10 +14,32 @@
 
 #include "gazebo_ros/gazebo_ros_init.hpp"
 
-#include <rclcpp/rclcpp.hpp>
+#include <gazebo_ros/node.hpp>
+#include <gazebo_ros/utils.hpp>
+#include <gazebo_ros/conversions.hpp>
+#include <rosgraph_msgs/msg/clock.hpp>
 #include <gazebo/common/Plugin.hh>
 
+#include <memory>
+
+namespace gazebo_ros
+{
+
+class GazeboRosInitPrivate
+{
+public:
+  GazeboRosInitPrivate();
+  gazebo_ros::Node::SharedPtr rosnode_;
+  rclcpp::Publisher<rosgraph_msgs::msg::Clock>::SharedPtr clock_pub_;
+  gazebo::event::ConnectionPtr world_update_event_;
+  gazebo_ros::Throttler throttler_;
+  static const gazebo::common::Time DEFAULT_PUBLISH_PERIOD;
+
+  void PublishSimTime(const gazebo::common::UpdateInfo & _info);
+};
+
 GazeboRosInit::GazeboRosInit()
+: impl_(std::make_unique<GazeboRosInitPrivate>())
 {
 }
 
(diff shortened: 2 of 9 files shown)


gazebo_ros/include/gazebo_ros/gazebo_ros_clock.hpp
```diff
@@ -0,0 +1,47 @@
...
+  /// Constructor
+  GazeboRosClock();
+
+  /// Destructor
+  virtual ~GazeboRosClock();
+
+  /// Called by Gazebo to load plugin.
+  /// \param[in] argc Argument count.
+  /// \param[in] argv Argument values.
+  void Load(int argc, char ** argv);
```
chapulina: We could add `override`

gazebo_ros/src/gazebo_ros_clock.cpp
```diff
@@ -0,0 +1,120 @@
...
+  gazebo::event::ConnectionPtr world_created_event_;
+  gazebo::event::ConnectionPtr world_update_event_;
+  gazebo::physics::WorldPtr world_;
+  std::atomic_bool world_created_flag_;
+  gazebo::common::Time last_publish_time_;
+  gazebo::common::Time publish_period_;
+  static const gazebo::common::Time DEFAULT_PUBLISH_PERIOD;
+
+  void OnWorldCreated(std::string _world_name);
+  void PublishSimTime();
```
chapulina: Missing documentation

gazebo_ros/src/gazebo_ros_clock.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+GazeboRosClock::~GazeboRosClock()
+{
+}
+
+void GazeboRosClock::Load(int argc, char ** argv)
+{
+  (void) argc;
+  (void) argv;
```
chapulina: Should we try to pass these to `rclcpp::init` in case it hasn't been initialized yet?

gazebo_ros/include/gazebo_ros/gazebo_ros_clock.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <gazebo/common/Plugin.hh>
+
+#include <memory>
+
+namespace gazebo_ros
+{
+
+class GazeboRosClockPrivate;
+
+/// Initializes ROS with the system arguments passed to Gazebo (i.e. calls rclcpp::init).
```
chapulina: Documentation from a different plugin

gazebo_ros/src/gazebo_ros_clock.cpp
```diff
@@ -0,0 +1,120 @@
...
+  rclcpp::Publisher<rosgraph_msgs::msg::Clock>::SharedPtr clock_pub_;
+  gazebo::event::ConnectionPtr world_created_event_;
+  gazebo::event::ConnectionPtr world_update_event_;
+  gazebo::physics::WorldPtr world_;
+  std::atomic_bool world_created_flag_;
+  gazebo::common::Time last_publish_time_;
+  gazebo::common::Time publish_period_;
+  static const gazebo::common::Time DEFAULT_PUBLISH_PERIOD;
+
+  void OnWorldCreated(std::string _world_name);
```
chapulina: Make input a const reference

gazebo_ros/src/gazebo_ros_clock.cpp
```diff
@@ -0,0 +1,120 @@
...
+  world_ = gazebo::physics::get_world(_world_name);
+  if (!world_) {
+    RCLCPP_ERROR(rosnode_->get_logger(), "Cannot get world [%s]. /clock will NOT be published",
+      _world_name.c_str());
+    return;
+  }
+
+  clock_pub_ = rosnode_->create_publisher<rosgraph_msgs::msg::Clock>("/clock");
+  world_update_event_ = gazebo::event::Events::ConnectWorldUpdateBegin(
+    std::bind(&GazeboRosClockPrivate::PublishSimTime, this));
```
chapulina: You could connect to a callback that receives `gazebo::common::UpdateInfo`, so you can get the times from there and don't need to keep a pointer to the world. This should save the whole logic of waiting for a world to be created.

See an example [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2_joint_state_pub/gazebo_plugins/src/gazebo_ros_joint_state_publisher.cpp#L140).

dhood:
I don't know that it makes sense to have such a standalone plugin for a ROS clock, I would think that it goes well with `gazebo_ros_init`. I have seen that both approaches have been mentioned on https://github.com/ros-simulation/gazebo_ros_pkgs/issues/779 but not a discussion of pros/cons.

I'd say it makes sense that `gazebo_ros_init` sets up a reasonable amount of the default Gazebo-ROS integration, which includes clock publishing. No need to separate them if they will always be used together (I've never heard of anyone not wanting the clock to be published, and even if there's a use case for that, it might not warrant two separate plugins). What's the motivation for keeping them separate?

gazebo_ros/test/test_sim_time.cpp
```diff
@@ -0,0 +1,75 @@
...
+void TestSimTime::TearDown()
+{
+  ASSERT_GE(gazebo_process_->Terminate(), 0);
+  gazebo_process_.reset();
+}
+
+TEST_F(TestSimTime, TestClock)
+{
+  auto node = std::make_shared<rclcpp::Node>("my_node");
+  rclcpp::Clock clock;
```
dhood: don't think this is what you want here (an `rclcpp::Clock` might not be using sim time). You'd be wanting to check `node->now()` (which will use ROS time [if the `use_sim_time` parameter has been set](http://design.ros2.org/articles/clock_and_time.html#ros-time)). (When you do that, please set the `use_sim_time` parameter explicitly on the node and don't rely on https://github.com/ros2/rclcpp/issues/514)

gazebo_ros/test/test_sim_time.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  auto first_msg =
+    gazebo_ros::get_message_or_timeout<ClockMsg>(node, "/clock", rclcpp::Duration(5, 0));
+  ASSERT_NE(first_msg, nullptr);
+  auto first_msg_time = clock.now();
+
+  auto second_msg =
+    gazebo_ros::get_message_or_timeout<ClockMsg>(node, "/clock", rclcpp::Duration(5, 0));
+  ASSERT_NE(second_msg, nullptr);
+  auto second_msg_time = clock.now();
```
dhood: I imagine you also wanted to check that the time roughly matches what was published on the clock topic

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -29,7 +51,47 @@ void GazeboRosInit::Load(int argc, char ** argv)
...
+
+  impl_->rosnode_ = gazebo_ros::Node::Create("gazebo_ros_clock");
+  impl_->clock_pub_ = impl_->rosnode_->create_publisher<rosgraph_msgs::msg::Clock>("/clock");
+
+  // Get publish rate from parameter if set
+  rclcpp::Parameter rate_param;
+  if (impl_->rosnode_->get_parameter("publish_rate", rate_param)) {
+    if (rclcpp::ParameterType::PARAMETER_DOUBLE == rate_param.get_type()) {
+      impl_->throttler_ = Throttler(gazebo::common::Time(1.0 / rate_param.as_double()));
+    } else if (rclcpp::ParameterType::PARAMETER_INTEGER != rate_param.get_type()) {
```
dhood: should this be `==`?

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -14,10 +14,32 @@
...
+#include <memory>
+
+namespace gazebo_ros
+{
+
+class GazeboRosInitPrivate
+{
+public:
+  GazeboRosInitPrivate();
+  gazebo_ros::Node::SharedPtr rosnode_;
```
dhood: recommend `ros_node_` to match e.g. https://github.com/ros-simulation/gazebo_ros_pkgs/blob/4fe1967e3e6c1ee1cc176dfd253f7dd0d9ae51f9/gazebo_plugins/src/gazebo_ros_joint_state_publisher.cpp#L58

gazebo_ros/src/utils.cpp
```diff
@@ -60,4 +60,30 @@ std::string SensorFrameID(const gazebo::sensors::Sensor & _sensor, const sdf::El
   return gazebo_ros::ScopedNameBase(_sensor.ParentName());
 }
 
+Throttler::Throttler(const gazebo::common::Time & _period)
```
dhood: I don't think Times should be use to represent durations (at least in ROS we have different types; Time is meant for time stamps). Just the rate constructor seems like it'd be fine

gazebo_ros/src/utils.cpp
```diff
@@ -60,4 +60,30 @@ std::string SensorFrameID(const gazebo::sensors::Sensor & _sensor, const sdf::El
...
+}
+
+Throttler::Throttler(const double _hz)
+: Throttler(gazebo::common::Time(1.0 / _hz))
+{
+}
+
+bool Throttler::IsReady(const gazebo::common::Time & _now)
+{
+  if (_now < last_time_) {
```
dhood: is this for a backwards time jump? if so a comment would be good

gazebo_ros/test/test_sim_time.cpp
```diff
@@ -0,0 +1,75 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+#include <gazebo_ros/testing_utils.hpp>
+#include <rclcpp/rclcpp.hpp>
+#include <rosgraph_msgs/msg/clock.hpp>
+
+#include <memory>
+#include <string>
```
dhood, line 21: doesn't seem that this or utility are used

gazebo_ros/test/test_sim_time.cpp
```diff
@@ -0,0 +1,75 @@
...
+    gazebo_ros::get_message_or_timeout<ClockMsg>(node, "/clock", rclcpp::Duration(5, 0));
+  ASSERT_NE(first_msg, nullptr);
+  auto first_msg_time = clock.now();
+
+  auto second_msg =
+    gazebo_ros::get_message_or_timeout<ClockMsg>(node, "/clock", rclcpp::Duration(5, 0));
+  ASSERT_NE(second_msg, nullptr);
+  auto second_msg_time = clock.now();
+
+  ASSERT_GT(second_msg, first_msg);
```
dhood: be good to check that the time values published also seem reasonable individually

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -29,7 +51,47 @@ void GazeboRosInit::Load(int argc, char ** argv)
...
+    1E8);
+
+GazeboRosInitPrivate::GazeboRosInitPrivate()
+: throttler_(DEFAULT_PUBLISH_PERIOD)
+{
+}
+
+void GazeboRosInitPrivate::PublishSimTime(const gazebo::common::UpdateInfo & _info)
+{
+  if (!throttler_.IsReady(_info.realTime)) {return;}
```
dhood, line 91: @tfoote do you agree that specifying the rate of publishing to `/clock` wrt to real time is appropriate here? seems reasonable since then even if RTF is really low the clock will still be published regularly

gazebo_ros/test/test_utils.cpp
```diff
@@ -92,6 +92,40 @@ TEST(TestUtils, SensorFrameID)
...
+    Throttler t(-1);
+    EXPECT_TRUE(t.IsReady(Time()));
+    EXPECT_TRUE(t.IsReady(Time(50, 0)));
+    EXPECT_TRUE(t.IsReady(Time(20, 0)));
+    EXPECT_TRUE(t.IsReady(Time(1E3, 0)));
+  }
+  {
+    // Test frequency
+    Throttler t(2.0);
+    EXPECT_FALSE(t.IsReady(Time(0, 4E8)));
```
dhood: for the usual usage (with time stamps as an input), the first call to `IsReady` will be true because of the jump from 0. This isn't necessarily an issue, but want to check it's what we want since it's contrary to this specific test. If so would be good to document (and maybe make that behaviour consistent in all cases).

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -29,7 +51,47 @@ void GazeboRosInit::Load(int argc, char ** argv)
 {
   // Initialize ROS with arguments
   rclcpp::init(argc, argv);
-  RCLCPP_INFO(rclcpp::get_logger("gazebo_ros_init"), "ROS has been initialized.");
+
+  impl_->rosnode_ = gazebo_ros::Node::Create("gazebo_ros_clock");
+  impl_->clock_pub_ = impl_->rosnode_->create_publisher<rosgraph_msgs::msg::Clock>("/clock");
```
dhood: @tfoote I think setting durability on the `/clock` publisher's QoS settings would be appropriate for late-subscribers when the simulation's paused. Would you agree?

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -29,7 +51,47 @@ void GazeboRosInit::Load(int argc, char ** argv)
...
+        1.0 / GazeboRosInitPrivate::DEFAULT_PUBLISH_PERIOD.Double());
+    }
+  }
+
+  impl_->world_update_event_ = gazebo::event::Events::ConnectWorldUpdateBegin(
+    std::bind(&GazeboRosInitPrivate::PublishSimTime, impl_.get(), std::placeholders::_1));
+}
+
+// By default publish at 10 HZ (100 millisecond period)
+const gazebo::common::Time GazeboRosInitPrivate::DEFAULT_PUBLISH_PERIOD = gazebo::common::Time(0,
```
dhood: (genuine question): how come we don't just initialise this above?

gazebo_ros/include/gazebo_ros/gazebo_ros_init.hpp
```diff
@@ -28,9 +35,12 @@ class GazeboRosInit : public gazebo::SystemPlugin
   /// Destructor
   virtual ~GazeboRosInit();
 
-  /// Called by Gazebo to load plugin.
-  /// \param[in] argc Argument count.
-  /// \param[in] argv Argument values.
+  // Documentation inherited
   void Load(int argc, char ** argv);
```
dhood: could add `override` (attribution: think this might have been mentioned before)

dhood (commented):
exciting we will have sim time in ROS 2 soon!

gazebo_ros/src/utils.cpp
kev-the-dev: I agree the other constructor may be confusing. I'll continue to use time to store the period internally because Gazebo does not have a duration type

dhood: I would suggest instead to store the period as a double and compare doubles with doubles e.g. https://github.com/ros-simulation/gazebo_ros_pkgs/pull/793/files#diff-b9fc9fd135967672730a2b1c536338b7L71

kev-the-dev:
Rewrote test to actually test a node's sim time and sanity check on the /clock output.

gazebo_ros/test/test_sim_time.cpp
```diff
@@ -49,22 +48,32 @@ void TestSimTime::TearDown()
...
+  auto second_msg_time = node->now();
 
-  ASSERT_GT(second_msg, first_msg);
+  // The SIM time should be close to zero (start of simulation)
+  EXPECT_LT(rclcpp::Time(first_msg->clock), rclcpp::Time(1, 0, RCL_ROS_TIME));
+  EXPECT_LT(rclcpp::Time(second_msg->clock), rclcpp::Time(1, 0, RCL_ROS_TIME));
+  // Time should go forward
+  EXPECT_GT(second_msg, first_msg);
+  // The message from the clock should be the node's time
+  EXPECT_EQ(first_msg_time, rclcpp::Time(first_msg->clock));
```
dhood, line 75: Seems like this is subject to a race condition (additional clock message may have been received by the node after `first_msg_time` was stored), so we will likely benefit from a tolerance here and below

dhood (commented):
looks pretty good to me now!

mikaelarguedas:

gazebo_ros/src/gazebo_ros_init.cpp
tfoote: There's not a real standard, but I believe that the publish rate wrt to realtime makes sense. It's mostly limited by the clock communication bandwidth requirments which are hardware based and thus connect to realtime.

gazebo_ros/src/gazebo_ros_init.cpp
tfoote: Yes, adding durability would make sense so that a late starting node would get their time initialized.

gazebo_ros/test/test_sim_time.cpp
dhood: I take back what I said, since spinning only happens in `get_message_or_timeout`. It's not possible that additional messages are received between storing `first_msg` and `first_msg_time`.

dhood (dismissed):

gazebo_ros/src/gazebo_ros_init.cpp
dhood: added in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/794/commits/165cb300a794e4adb5bce40f99375ea9118987ce

dhood (approved):

gazebo_ros/src/gazebo_ros_init.cpp
dhood: interestingly, in ROS 1 it isn't published as latched: https://github.com/ros-simulation/gazebo_ros_pkgs/blob/6a8202411e3ea00fc22b197c248caac56f271fd9/gazebo_ros/src/gazebo_ros_api_plugin.cpp#L221

dhood:
I added example usage and a wiki page to the original post

PR #795, opened 2018-08-07
[ros2] gazebo_ros_joint_state_publisher

chapulina:
Try out the demo, make sure you either have the `double_pendulum_with_base` model locally or an internet connection:

`gazebo --verbose install/gazebo_plugins/share/gazebo_plugins/worlds/gazebo_ros_joint_state_publisher_demo.world`

joint_state_pub

---

#### SDF

ROS | ROS 2
------ | -------
`<robotNamespace>` | `<ros><namespace>`
`<jointName>joint1, joint2</jointName>` | `<joint_name>joint1</joint_name><joint_name>joint2</joint_name>`
`<updateRate>` | `<update_rate>`


gazebo_plugins/src/gazebo_ros_joint_state_publisher.cpp
```diff
@@ -0,0 +1,183 @@
...
+{
+}
+
+GazeboRosJointStatePublisher::~GazeboRosJointStatePublisher()
+{
+}
+
+void GazeboRosJointStatePublisher::Load(gazebo::physics::ModelPtr model, sdf::ElementPtr sdf)
+{
+  auto logger = rclcpp::get_logger("gazebo_ros_force");
```
kev-the-dev: Should be ```"gazebo_ros_joint_state_publisher"```, though I personally favor creating the Node in the beginning and using it's logger ```node->get_logger()```. This will avoid any inconsistency in logging if the node name is remapped

kev-the-dev (changes requested):
One typo, otherwise LGTM

gazebo_plugins/src/gazebo_ros_joint_state_publisher.cpp
chapulina: I actually started that way, but then I moved the node creation to the bottom so we don't need to destroy the node in case the load function returns early.

But I can move it back to the top if you have a strong preference, I agree that it would make the logging consistent.

kev-the-dev (approved):

PR #796, opened 2018-08-08
[ros2] Fix missing dependencies to run CI and update maintainers

j-rivero:
@ironmig, @chapulina I was not sure about adding your name to maintainers but it is useful to receive the CI mails.

### DIFF
--- a/gazebo_dev/package.xml
+++ b/gazebo_dev/package.xml
@@ -7,8 +7,7 @@
     Provides a cmake config for the default version of Gazebo for the ROS distribution.
   </description>
 
-  <maintainer email="hsu@osrfoundation.org">John Hsu</maintainer>
-  <maintainer email="davetcoleman@gmail.com">Dave Coleman</maintainer>
+  <maintainer email="jrivero@openrobotics.org">Jose Luis Rivero</maintainer>
 
   <license>Apache 2.0</license>
 
--- a/gazebo_ros/package.xml
+++ b/gazebo_ros/package.xml
@@ -27,6 +27,7 @@
   <exec_depend>rclcpp</exec_depend>
   <exec_depend>gazebo_dev</exec_depend>
 
+  <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
   <test_depend>std_msgs</test_depend>


j-rivero:
The CI is failing due to a bug in the Ubuntu libgazebo9-dev package. 
```
05:20:16 [100%] Linking CXX executable test_gazebo_ros_force
05:20:16 /usr/bin/x86_64-linux-gnu-ld: cannot find -lgazebo_test_fixture
05:20:16 collect2: error: ld returned 1 exit status
05:20:16 test/CMakeFiles/test_gazebo_ros_force.dir/build.make:188: recipe for target 'test/test_gazebo_ros_force' failed
05:20:16 make[2]: *** [test/test_gazebo_ros_force] Error 1
05:20:16 CMakeFiles/Makefile2:1090: recipe for target 'test/CMakeFiles/test_gazebo_ros_force.dir/all' failed
05:20:16 make[1]: *** [test/CMakeFiles/test_gazebo_ros_force.dir/all] Error 2
05:20:16 Makefile:140: recipe for target 'all' failed
05:20:16 make: *** [all] Error 2
05:20:16 --- stderr: gazebo_plugins
```
I've [reported sometime ago the bug to Ubuntu](https://bugs.launchpad.net/ubuntu/+source/gazebo/+bug/1773811) but it is still open. We could generate custom gazebo9 packages and upload them to the ros2 repo as a workaround.

kev-the-dev:
Could we add the gazebo apt source `http://packages.osrfoundation.org/gazebo/ubuntu-stable` to the ros2 CI process instead? It will be useful to test against new gazebo features eventually anyways

chapulina:
Thanks, @j-rivero ! On https://github.com/ros-simulation/gazebo_ros_pkgs/pull/795/files#diff-f437ab417bed30627dab7e3a6b14c0edR34 I also added `ament_cmake_gtest` to `gazebo_plugins`.

+1 to @ironmig's comment about adding the OSRF PPA. This should fix the test fixture issue and it's easier to maintain than having duplicate gazebo9 packages on the ROS2 repo.

chapulina (approved):

j-rivero:
> Could we add the gazebo apt source http://packages.osrfoundation.org/gazebo/ubuntu-stable instead? It will be useful to test against new gazebo features eventually anyways

I'm afraid that the ROS buildfarm does not accept any kind of PPA or external repositories. That is why I proposed the custom fixed packages as workaround. The other option would be to embed gazebo code in this repo to generate the gazebo_test_fixture.

chapulina:
Ah got it. I was talking to @wjwwood and something like [this](https://github.com/ros-infrastructure/ros_bootstrap_dependencies/blob/master/ubuntu/config/gazebo5.upstream.yaml) could also be an option.

mikaelarguedas:
> I'm afraid that the ROS buildfarm does not accept any kind of PPA or external repositories.

Yes, for security reasons we don't allow any external repositories.

> That is why I proposed the custom fixed packages as workaround.

That can work. Though what are the implications for users ? will that conflict with the upstream version? Making it impossible to install ROS melodic and ROS bouncy side by side for example ?

mikaelarguedas:
> Ah got it. I was talking to @wjwwood and something like this could also be an option.

Yeah this is the equivalent of what Jose suggested "Import custom packages to the ros buildfarm"

j-rivero:
> Yeah this is the equivalent of what Jose suggested "Import custom packages to the ros buildfarm"

Yes that was the idea. When I mentioned "custom packages" I was not thinking on importing packages.osrfoundation.org versions since they are several minor versions above (9.3.0 osrfoundation vs 9.0.0 ubuntu). 

My goal during these years was to provide users with only the ROS repo installed the same version shipped by Ubuntu repositories with the minimal intervention to fix the important bugs. The reasoning behind this is to minimize the maintenance and debug effort for the different options: 1) use gazebo from Ubuntu repo, 2) use gazebo from ROS repo (when available) and 3) use gazebo from packages.osrfoundation.org + ROS. If we built custom gazebo packages with small patches on top of 1) the solution delivered by 2) is almost the same. If we import a random version from packages.osrfoundation.org we will end up with three different scenarios.

mikaelarguedas:
Thanks for clarifying, I didn't catch the nuance between 2) and 3) originally :+1:

kev-the-dev (approved):
Regardless of the outcome of the gazebo apt source, this PR is an improvement

PR #802, opened 2018-08-20
Wait before publishing force msg to reduce test flakiness

dhood:
Saw this test fail when testing another PR: http://build.ros2.org/job/Bpr__gazebo_ros_pkgs__ubuntu_bionic_amd64/38/testReport/gazebo_plugins/GazeboRosForceTest/ApplyForceTorque/

(Note to reviewers: [durability QoS profiles](https://github.com/ros2/ros2/wiki/About-Quality-of-Service-Settings#qos-compatibilities) can be used to avoid race conditions like this in some situations, but in this case that would involve the force plugin's subscriber requesting transient local durability, and unfortunately that would then _require all_ publishers that interact with the plugin to offer transient local durability).

Locally this passed with `colcon test --event-handlers console_direct+ --packages-select gazebo_plugins --retest-until-fail 100 --ctest-args -R force`

### DIFF
--- a/gazebo_plugins/test/test_gazebo_ros_force.cpp
+++ b/gazebo_plugins/test/test_gazebo_ros_force.cpp
@@ -50,6 +50,8 @@ TEST_F(GazeboRosForceTest, ApplyForceTorque)
   ASSERT_NE(nullptr, node);
 
   auto pub = node->create_publisher<geometry_msgs::msg::Wrench>("test/force_test");
+  // Wait a bit for the force plugin to be initialized.
+  gazebo::common::Time::MSleep(1000);
 
   // Apply force on X
   auto msg = geometry_msgs::msg::Wrench();


tfoote (changes requested):
Instead of sleeping before publishing how about we put the publish call inside the loop so it repeats and the test is robust to one or more dropped packets.  Republishing it shouldn't adversely effect the test. And this will avoid one more magic system dependent number.

chapulina:
I think this is not yet available on `rclcpp`, but maybe something like [ros::Publisher::getNumSubscribers](http://docs.ros.org/jade/api/roscpp/html/classros_1_1Publisher.html#ad56c8e954020290dcb87cd6bbd330a2b) could be helpful here?

chapulina:
I haven't seen this test failing recently. Since this PR has been here for a while, I suggest we decline it for now, and revive it if we observe the flakiness again.

PR #784, opened 2018-07-28
[ros2] gazebo_ros_force plugin

chapulina:
* Ported plugin without changing behaviour
* Make use of new SDF param parsing
* Use `gazebo::ServerFixture` for testing
* Add demo world, see instructions on the file
* Start a series of conversion functions, the idea is to keep filling it with handy conversions between ROS / Gazebo / SDF / Ignition (Math/Msgs) types

ros2force

PR #803, opened 2018-08-22
[ros2] Split conversions into headers specific to message packages

dhood:
Otherwise any package that included `conversions.hpp` needed to build depend on all of the message packages included in that header.

Noticed the need for this because since https://github.com/ros-simulation/gazebo_ros_pkgs/pull/785 has been merged, `gazebo_ros` on the `ros2` branch can't be built in isolation (missing dependency on `sensor_msgs`):
```
In file included from /home/dhood/ros2_ws/src/ros2/gazebo_ros_pkgs/gazebo_ros/test/test_conversions.cpp:15:0:
/home/dhood/ros2_ws/src/ros2/gazebo_ros_pkgs/gazebo_ros/include/gazebo_ros/conversions.hpp:30:44: fatal error: sensor_msgs/msg/point_cloud2.hpp: No such file or directory
```

A similar thing happened when https://github.com/ros-simulation/gazebo_ros_pkgs/pull/784 was merged requiring `geometry_msgs` (fixed in 80f3a789e2b670ea3f7ccd534d245e3036d337b3).

@chapulina not sure how you have been building locally but since the buildfarm won't catch issues building in isolation, if you can switch to an isolated build that will help catch these issues earlier :smile: (`colcon build` will use isolation by default)

### DIFF
--- a/gazebo_ros/package.xml
+++ b/gazebo_ros/package.xml
@@ -25,7 +25,6 @@
   <depend>gazebo_dev</depend>
   <depend>rclcpp</depend>
 
-  <build_export_depend>builtin_interfaces</build_export_depend>
   <build_export_depend>geometry_msgs</build_export_depend>
   <build_export_depend>sensor_msgs</build_export_depend>
 
--- a/gazebo_ros/package.xml
+++ b/gazebo_ros/package.xml
@@ -31,6 +31,7 @@
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
+  <test_depend>sensor_msgs</test_depend>
   <test_depend>std_msgs</test_depend>
 
   <export>
--- a/gazebo_ros/package.xml
+++ b/gazebo_ros/package.xml
@@ -25,6 +25,10 @@
   <depend>gazebo_dev</depend>
   <depend>rclcpp</depend>
 
+  <build_export_depend>builtin_interfaces</build_export_depend>
+  <build_export_depend>geometry_msgs</build_export_depend>
+  <build_export_depend>sensor_msgs</build_export_depend>
+
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
--- a/gazebo_ros/package.xml
+++ b/gazebo_ros/package.xml
@@ -21,16 +21,14 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <build_depend>rclcpp</build_depend>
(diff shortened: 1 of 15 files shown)


chapulina:
humm strange, I've always just used plain `colcon build` locally and can't reproduce the issue :thinking: But it sounds like the fix is important anyway, thanks for catching that.

Regarding the separation of files, this is something I agree is a problem and the issue had already crossed my mind. The only issue I see with the current split is that it only considers one side of the conversion, which seems to be always a ROS message. But each conversion file will still potentially need to include various Gazebo, Ignition and SDF headers.

So I suggest we split according to pairs. For example, [tf2 conversions](https://github.com/ros/geometry2) are separated into packages which convert from tf2 to something else. So we'd get headers like these:

* `sensor_msgs-ign_math`
* `sensor_msgs-gz_msgs`
* `sensor_msgs-ign_msgs`
* `sensor_msgs-gz_common`
* etc.

Does this sound too fine-grained?

tfoote (commented):
As we mentioned in the meeting this looks good with the addition of the necessary build_export_depend http://www.ros.org/reps/rep-0140.html#build-export-depend-multiple tags

chapulina:
Conclusions from the meeting:
* This is an improvement for package dependencies, let's go with the current split for now
* In the future we may split further not to bring unnecessary includes
* In the future, TF2's conversion functions could be promoted to a more common package such as rcl-utils so that they can be used for non-tf conversions as well

dhood:
@tfoote does this look alright now after https://github.com/ros-simulation/gazebo_ros_pkgs/pull/803/commits/9803d5c5ba5fe0ddee689508b1a3589abd88cccc? gazebo_ros has the conversions, and so exports a dependency on the three message packages for which conversions are supported. It build depends on builtin_interfaces because gazebo_ros_init uses it, and it additionally test depends on the other two message packages because test_conversions uses them.

gazebo_ros/package.xml
```diff
@@ -21,16 +21,19 @@
...
-  <build_depend>rclcpp</build_depend>
-  <build_depend>gazebo_dev</build_depend>
-  <build_depend>geometry_msgs</build_depend>
+  <depend>builtin_interfaces</depend>
+  <depend>gazebo_dev</depend>
+  <depend>rclcpp</depend>
 
-  <exec_depend>rclcpp</exec_depend>
-  <exec_depend>gazebo_dev</exec_depend>
+  <build_export_depend>builtin_interfaces</build_export_depend>
```
tfoote: Adding this is redundant with changing the above build_depend to depend

```
- The generic dependency on 'builtin_interfaces' is redundant with: build_export_depend
```

tfoote (approved):
This looks good to me. I had to update one of the depends. Added in 45137fb before this I couldn't compile in my fresh workspace.

dhood:
thanks!

PR #806, opened 2018-08-26
[ros2] Port diff_drive plugin to ros2

tfoote:
This is compiling but has not been tested. I'm opening the PR for visibility as I'll be on vacation next week.

There are a few TODOs to check and adding some unit tests would be really good.

This builds on top of #803 as well.


chapulina:
I finished most of the port and started the [migration guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Diff-drive). Also:

* Added a demo world with instructions
* Added a test that commands are received and odom is published
* The test doesn't check TF

When visualizing TF on RViz, note that due to https://github.com/ros2/rviz/issues/332 the wheels are not updating correctly, but the TF is being properly published:

tf_diff_drive

---

@tfoote , I haven't addressed the QoS comments you left. If you don't have time for it once you're back, I can take a stab at it.

chapulina:
Motivated by this PR, I also started a [migration guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-gazebo_ros_utils#gazeborosgetparameterboolean) for `gazebo_ros_utils`, more specifically the `get*` functions.

chapulina:
After talking to @tfoote , we will merge this PR and retarget #815 to `ros2`

PR #811, opened 2018-08-30
Initialize gazebo with initial physics configuration

marting87:
Gazebo is **not** using any of the following parameters during initialization:

- time_step
- max_update_rate
- gravity_x
- gravity_y
- gravity_z
- auto_disable_bodies
- sor_pgs_precon_iters
- sor_pgs_iters
- sor_pgs_w
- sor_pgs_rms_error_tol
- cfm
- erp
- contact_surface_layer
- contact_max_correcting_vel
- max_contacts

The default values from _Physics.cfg_ and the values specified in your launch file are overwritten during initialization by the default values used inside Gazebo. 

After initialization it is possible to change the parameters (dynamic reconfigure). 

Example:

When I provide a custom `max_update_rate` of 50.0 and call `rosservice call /gazebo/get_physics_properties` after initialization I can see a different `max_update_rate`
 (1000.0) which does not correspond to the default value or my custom value. In my case this leads to excessive CPU usage. 

This fix ensures **SetPhysicsProperties** is called during initialization to deliver the provided settings in your launch file or default values specified in _Physics.cfg_ to Gazebo.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -2237,31 +2237,9 @@ void GazeboRosApiPlugin::publishModelStates()
 
 void GazeboRosApiPlugin::physicsReconfigureCallback(gazebo_ros::PhysicsConfig &config, uint32_t level)
 {
-  if (!physics_reconfigure_initialized_)
+  bool changed = false;
+  if (physics_reconfigure_initialized_)
   {
-    gazebo_msgs::GetPhysicsProperties srv;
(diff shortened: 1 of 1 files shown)


marting87:
In addition, i would like to set:

1. `sor_pgs_iters` by default to 50 in _Physics.cfg_ to align with the default value of Gazebo. There are various reports on oscillating joints with the default value of 20.
2. `time_step` and `max_update_rate` by default to values so it ensures a real-time factor of 1 (`time_step = 0.001, max_update_rate=1000`)

PR #815, opened 2018-09-06
Switch to use sensor_data qos setting for short queue sizes.

tfoote:
@chapulina I think that this is the best default profile to use. It's not exactly the 1 queue size but it's close at 5.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_diff_drive.cpp
+++ b/gazebo_plugins/src/gazebo_ros_diff_drive.cpp
@@ -268,16 +268,11 @@ void GazeboRosDiffDrive::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr 
   }
   impl_->last_update_time_ = _model->GetWorld()->SimTime();
 
-  // Subscribe to the velocity command topic
-  // TODO(tfoote) equivalent qos settings as below
-  // ros::SubscribeOptions so =
-  //     ros::SubscribeOptions::create<geometry_msgs::Twist>(command_topic_, 1,
-  //             boost::bind(&GazeboRosDiffDrive::cmdVelCallback, this, _1),
-  //             ros::VoidPtr(), &queue_);
-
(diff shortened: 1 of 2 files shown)


chapulina:
Thanks for the PR, @tfoote ! Do you have an idea why the test now fails on [this line](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/806/files#diff-3dd34387f16cc03316857a4793061792R84)? It looks like the odom message is never received. I'm debugging a bit here.

chapulina:
Humm yeah it's not immediately clear to me why the QoS settings break the test. I checked that the plugin receives the command and is publishing odom messages, but the callback on the test is never called. Maybe you have an idea? Otherwise I can try to poke deeper.

chapulina (commented):
As discussed offline, the test was changed to use the same QoS settings as the plugin and now it passes: 315bfde330a81ab073e259809551328b44728e0c. I'll add a note to the migration page.

I'll merge this if CI comes back clean.

dhood:
Changing to the sensor data profile changes the depth but also the reliability (which I think you've spoken about offline). This PR has the following side-effects wrt reliability that we may want to reconsider:
- makes the `odom` messages only subscribable by best effort subscriptions (which is why the test needed to be updated). I would recommend just lowering the depth and leaving reliable reliability offered, so that the subscriber can choose if they want reliable or best effort.
- makes the `cmd_vel` subscription best effort (if it's published at low frequency, missing a message once will mean it's missed forever). Best effort might be appropriate here if we expect it the messages to always come at high frequency and we don't want to bother processing old dropped messages, but reliable seems safer if that isn't always going to be the case (though that would limit it to only reliable publishers, unless the qos is modified externally).

chapulina:
Thanks for taking a look at this, @dhood . I don't understand much about QoS settings yet, but all your comments sound reasonable. Would you be up for making a quick PR with the suggestions?

PR #817, opened 2018-09-07
Lower minimum cmake version

paulbovbel:
This patch lowers the required cmake version so that package builds on xenial. We've been running this build for some time with no adverse effects.


paulbovbel:
Can't really make heads or tails of the ros_gazebo_pkgs-ci-pr_any_melodic-bionic-amd64 CI build.

chapulina (approved):
This looks ok to me, everything in the repo builds fine, and if anyone building on top needs a higher version it's up to them to bump it.

The ROS1 builds on build.osrfoundation.org are broken right now, so don't worry about it.

PR #819, opened 2018-09-11
Diff drive: Reliable QoS with depth of 1

dhood:
Follow up to https://github.com/ros-simulation/gazebo_ros_pkgs/pull/815

Depth size of 1 is used because that was the case in ros1.


chapulina (approved):
Thanks!

Issue #834, opened 2018-11-02
Laser plugins time_increment and scan_time zero


reinzor:
See:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_gpu_laser.cpp#L195

and

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_laser.cpp#L190

Some packages require the time_increment to be non-zero, e.g. https://github.com/googlecartographer/cartographer_ros

skohlbr:
As noted in #835 it is in fact true that simulated scans generated with currently existing plugins are instantaneous and one has to take care not to publish wrong information to consumers by faking a non-zero scan time, when it is in fact zero.

reinzor:
I agree, limited PR to only update the scan_time

PR #835, opened 2018-11-02
Scan time and time increment in LaserScan message

reinzor:
Closes https://github.com/ros-simulation/gazebo_ros_pkgs/issues/834

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
@@ -82,6 +82,12 @@ namespace gazebo
     /// \brief for setting ROS name space
     private: std::string robot_namespace_;
 
+    /// \brief time between measurement points (Laserscan ROS msg property)
+    private: float time_increment_;
+
+    /// \brief time in between scans (Laserscan ROS msg property)
+    private: float scan_time_;
+
     // deferred load in case ros is blocking
     private: sdf::ElementPtr sdf;
     private: void LoadThread();
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
@@ -82,9 +82,6 @@ namespace gazebo
     /// \brief for setting ROS name space
     private: std::string robot_namespace_;
 
-    /// \brief time between measurement points (Laserscan ROS msg property)
-    private: float time_increment_;
-
     /// \brief time in between scans (Laserscan ROS msg property)
     private: float scan_time_;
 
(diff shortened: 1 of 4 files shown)


gazebo_plugins/src/gazebo_ros_laser.cpp
```diff
@@ -98,6 +98,16 @@ void GazeboRosLaser::Load(sensors::SensorPtr _parent, sdf::ElementPtr _sdf)
...
 
+  if (!this->sdf->HasElement("timeIncrement"))
+  {
+    ROS_INFO_NAMED("gpu_laser", "GazeboRosLaser plugin missing <timeIncrement>, defaults to 0");
+    this->time_increment_ = 0;
+  }
+  else
+    this->time_increment_ = this->sdf->Get<float>("timeIncrement");
+
+  this->scan_time_ = static_cast<float>(1. / this->parent_ray_sensor_->UpdateRate());
```
chapulina: Better check if the update rate is zero, I've seen people setting it to zero on startup as a way to "turn off" the sensor.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
```diff
@@ -82,6 +82,12 @@ namespace gazebo
     /// \brief for setting ROS name space
     private: std::string robot_namespace_;
 
+    /// \brief time between measurement points (Laserscan ROS msg property)
+    private: float time_increment_;
+
+    /// \brief time in between scans (Laserscan ROS msg property)
+    private: float scan_time_;
```
chapulina, line 86: Adding new private member breaks ABI. We sometimes let it go through in this repo because certain plugins are not often inherited, but I suspect these may be part of those which are often extended.

I'm not sure what's the best way around this. Maybe get the increment and calculate the scan time at every `OnScan` callback?

Any thoughts, @j-rivero ?

chapulina (commented):
Thanks for the PR.

If this gets in, it would be good to also:

* Add a note about the new SDF parameter to [this tutorial](http://gazebosim.org/tutorials?tut=ros_gzplugins&cat=connect_ros#GPULaser)
* Make a similar PR to the `ros2` branch or ticket an issue about it

gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
reinzor: note: this also holds for the `time_increment_` member

skohlbr:
Given that sensor data acquisition in Gazebo is in fact instantaneous, saying it is not by faking time increment and scan times might make some packages "work", but if they use the data correctly, the actual computation they perform on it (like unwarping points) will be wrong.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
chapulina: Correct, it applies to both of them.

reinzor:
Agreed, limited PR to only scan time

paulwong16:
It seems that ray sensing in gazebo is instantaneous, so the time_increment should be zero. But I am considering about that is it possible to simulate a lidar with non-zero time_increment?

PR #837, opened 2018-11-11
[gazebo_ros_pkgs/gazebo_plugins ]Added gaussian noise to odometry & separate ground truth topic

Lakshadeep:
* Currently odometry published is perfect. Added gaussian noise to simulated wheel odometry to make it more realistic.
* Added separate topic for publishing ground truth.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_skid_steer_drive.cpp
+++ b/gazebo_plugins/src/gazebo_ros_skid_steer_drive.cpp
@@ -64,7 +64,12 @@ namespace gazebo {
     LEFT_REAR=3,
   };
 
-  GazeboRosSkidSteerDrive::GazeboRosSkidSteerDrive() {}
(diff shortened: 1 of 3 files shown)


gazebo_plugins/src/gazebo_ros_skid_steer_drive.cpp
```diff
@@ -232,6 +252,16 @@ namespace gazebo {
     } else {
       this->update_period_ = 0.0;
     }
+
+    // gaussian noise
+    if (!_sdf->HasElement("gaussianNoise"))
+    {
+      ROS_DEBUG_NAMED("skid_steer_drive", "GazeboRosSkidSteerDrive Plugin (ns = %s) <gaussianNoise>, defaults to 0.0");
```
scpeters, line 259: there's a compiler warning here because the `%s` doesn't have a matching argument

Issue #838, opened 2018-11-14
[ros2] - /gazebo/set_model_state service isn't showing up


mkhansenbot:
I'm trying to call the set_model_state service, but I don't see it listed when I do a service list, and when I try to call it I get `waiting for service to become available...`

```
mkhansen@mkhansen-desk:~$ ros2 service list | grep gazebo
/gazebo/describe_parameters
/gazebo/get_parameter_types
/gazebo/get_parameters
/gazebo/list_parameters
/gazebo/set_parameters
/gazebo/set_parameters_atomically

```
Is this supposed to be available in ROS2? I'm running gazebo with this command:
```
gazebo --verbose -s libgazebo_ros_init.so /home/mkhansen/ros2_dev/turtlebot3_ws/src/turtlebot3_simulations/turtlebot3_gazebo/worlds/turtlebot3_ros2_demo.world
```

mkhansenbot:
I should have mentioned:
Ubuntu 18.04
Gazebo 9

chapulina:
No, that service hasn't been ported to ROS2 yet, see #779. If that's blocking you, I can try to make some time to port it, it shouldn't be too hard.

mkhansenbot:
Actually I can live without that service for now, but just realized that also the /gazebo/model_states topic is also missing. That's a higher priority IMO. We're trying to use gazebo for testing and need to get the model state to know if the robot has arrived at the goal pose.

chapulina:
I've assigned the model states services and topics to myself, I'll tackle it asap.

mkhansenbot:
Thanks @chapulina

mhpanah:
> I've assigned the model states services and topics to myself, I'll tackle it asap.

@chapulina Has `set_model_state` service ported yet?

chapulina:
It's currently in a pull request: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/839

Given the time it's been up for review, I'm inclined to merge it if I get a positive review from someone in the community.

nzlz:
We have been using a [merged gazebo_ros_pkgs](https://github.com/AcutronicRobotics/gazebo_ros_pkgs/tree/ros2_gym-gazebo) with ros2_time_cmds, ros2_state and ros2_properties (my PR) for the last 20 days, intensively on a daily basis. Mainly for the upcoming version of gym_gazebo. No errors so far, I'd say ros2_state is ready to merge and others too.

[...]

PR #839, opened 2018-11-16
[ros2] World plugin with get/set entity state services

chapulina:
This is part of breaking `gazebo_ros_api` into several plugins, see #779 and #838.

See the migration documentation: [ROS 2 Migration: Entity states](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Entity-states).

Try the demo world added in this PR (see the world file for more detailed instructions):

`gazebo  gazebo_ros/worlds/gazebo_ros_state_demo.world`

FYI @mkhansen-intel.

### DIFF
--- a/gazebo_msgs/srv/GetEntityState.srv
+++ b/gazebo_msgs/srv/GetEntityState.srv
@@ -1,0 +1,11 @@
+string name                          # Entity's scoped name.
+                                     # An entity can be a model, link, collision, light, etc.
+                                     # Be sure to use gazebo scoped naming notation (e.g. [model_name::link_name])
+string reference_frame               # Return pose and twist relative to this entity.
+                                     # Leaving empty or "world" will use inertial world frame.
+---
+std_msgs/Header header               # Standard metadata for higher-level stamped data types.
+                                     # * header.stamp Timestamp related to the pose.
+                                     # * header.frame_id Filled with the relative_frame.
+gazebo_msgs/EntityState state        # Contains pose and twist.
+bool success                         # Return true if get was successful. If false, the state contains garbage.
--- a/.ros1_unported/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
+++ b/.ros1_unported/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
@@ -58,23 +58,17 @@
 #include "gazebo_msgs/GetWorldProperties.h"
 
 #include "gazebo_msgs/GetModelProperties.h"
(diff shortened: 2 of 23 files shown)


gazebo_msgs/srv/GetEntityState.srv
```diff
@@ -0,0 +1,12 @@
...
+                                     # Be sure to use gazebo scoped naming notation (e.g. [model_name::link_name])
+string reference_frame               # Return pose and twist relative to this entity.
+                                     # Leaving empty or "world" will use inertial world frame.
+---
+std_msgs/Header header               # Standard metadata for higher-level stamped data types.
+                                     # * header.stamp Timestamp related to the pose.
+                                     # * header.frame_id Filled with the relative_frame.
+gazebo_msgs/EntityState state        # Contains pose and twist.
+bool success                         # Return true if get was successful. If false, the state contains garbage.
+string status_message                # Comments if available.
```
tfoote: What sort of "comments" might go here?

tfoote (commented):
Overall it looks good. Adding the status_message is a little odd. In general I think that that sort of output is better to go out on the console logging. Which looking at the implementation is already being done. Because the only thing that we can programmatically do with that information is display it to the user. I'm not entirely against it but it seems to be pushing the limit of what should go back in the message versus go out to console.

gazebo_msgs/srv/GetEntityState.srv
chapulina: I was just keeping it because it was in ROS 1, but I agree it doesn't serve any good purpose, so removed in 6387f4a.

chapulina:

chapulina:

chapulina:
TODO(@chapulina) : rebase against new `crystal` branch

chapulina:
@nzlz's team has been using this branch (see https://github.com/ros-simulation/gazebo_ros_pkgs/issues/838#issuecomment-463455713) and I've already addressed @tfoote's review comment, so I'm merging this.

chapulina:
@mkhansen-intel , see #933. I'm still reviewing it. I was having some difficulties using it which I'm not sure were local problems in my setup.

JWSLAM:
@mkhansenbot  @chapulina  
I am using dashing，also want to use setModelState service. Which commit or branch should I check out. Waiting for reply, Thanks.

i1Cps:
I'm using foxy, and also want to user a service where i can set the state of a model. Have any similar services been migrated to Ros2 yet?

PR #844, opened 2018-11-27
fix: p3d sensor attached to fixed join link

reinzor:
User can now specify the base link as bodyName and overwrite the
childFrameName and add a xyzOffset.


reinzor:
Hmm, this does not work since the xyzOffset is w.r.t. the fixed frame

PR #845, opened 2018-11-28
gazebo_plugins: Port the gazebo_ros_p3d plugin

mjeronimo:
This is a port of the gazebo_ros_p3d plugin to ROS2. Tested with gazebo 9.0.0.

Example plugin XML:

```
<plugin name="p3d_base_controller" filename="libgazebo_ros_p3d.so">
  <ros>
    <namespace></namespace>
    <argument></argument>
  </ros>
  <update_rate>50.0</update_rate>
  <body_name>base_link</body_name>
  <topic_name>base_pose_ground_truth</topic_name>
  <frame_name>world</frame_name>
  <xyz_offsets>0 0 0</xyz_offsets>
  <rpy_offsets>0 0 0</rpy_offsets>
  <gaussian_noise>0.01</gaussian_noise>
</plugin>
```


gazebo_plugins/include/gazebo_plugins/gazebo_ros_p3d.hpp
```diff
@@ -0,0 +1,62 @@
...
+ * Author: Sachin Chitta and John Hsu
+ * Date: 1 June 2008
+ */
+
+#ifndef GAZEBO_PLUGINS__GAZEBO_ROS_P3D_HPP_
+#define GAZEBO_PLUGINS__GAZEBO_ROS_P3D_HPP_
+
+#include <memory>
+#include "gazebo/physics/physics.hh"
+#include "gazebo/common/Plugin.hh"
```
chapulina: Includes from external libraries should use `<>`

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+#include "tf2_ros/transform_listener.h"
+#include "nav_msgs/msg/odometry.hpp"
+
+namespace gazebo_plugins
+{
+
+class GazeboRosP3DPrivate
+{
+public:
+  gazebo::physics::WorldPtr world_;
```
chapulina: We're completing documentation as we port plugins to ROS 2, mind documenting all members?

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+  std::string topic_name_;
+
+  /// \brief frame transform name, should match link name
+  std::string frame_name_;
+
+  /// \brief allow specifying constant xyz and rpy offsets
+  ignition::math::Pose3d offset_;
+
+  /// \brief mutex to lock access to fields used in message callbacks
+  boost::mutex lock;
```
chapulina: I was going to say to use `std::mutex`, but actually I think this mutex can be completely removed. It is only used within the Update callback which always happens in the same thread (gazebo's physics thread), and unlike what the comment says, there are no message callbacks in this plugin.

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+  // Pointer to the update event connection
+  gazebo::event::ConnectionPtr update_connection_;
+
+  unsigned int seed_;
+};
+
+GazeboRosP3D::GazeboRosP3D()
+: impl_(std::make_unique<GazeboRosP3DPrivate>())
+{
+  impl_->seed_ = 0;
```
chapulina: It looks like the seed is always zero and there's no way to set it. I suggest this variable be removed.

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+}
+
+GazeboRosP3D::~GazeboRosP3D()
+{
+  impl_->update_connection_.reset();
```
chapulina: This line shouldn't be necessary

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+// Load the controller
+void GazeboRosP3D::Load(gazebo::physics::ModelPtr _parent, sdf::ElementPtr _sdf)
+{
+  // Create ROS configured from sdf
+  impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
+
+  // Get the world name
+  impl_->world_ = _parent->GetWorld();
+  impl_->model_ = _parent;
```
chapulina: `model_` is only used in this function, we don't need to keep a shared pointer to it

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+  if (!_sdf->HasElement("gaussian_noise")) {
+    RCLCPP_DEBUG(
+      impl_->ros_node_->get_logger(), "p3d plugin missing <gassian_noise>, defaults to 0.0");
+    impl_->gaussian_noise_ = 0;
+  } else {
+    impl_->gaussian_noise_ = _sdf->GetElement("gaussian_noise")->Get<double>();
+  }
+
+#if GAZEBO_MAJOR_VERSION >= 8
```
chapulina: ROS 2 wil only support Gazebo 9+, so you can get rid of all the if defs

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+#else
+    impl_->last_frame_vpos_ = impl_->reference_link_->GetWorldLinearVel().Ign();
+    impl_->last_frame_veul_ = impl_->reference_link_->GetWorldAngularVel().Ign();
+#endif
+  }
+
+  // New Mechanism for Updating every World Cycle
+  // Listen to the update event. This event is broadcast every simulation iteration
+  impl_->update_connection_ = gazebo::event::Events::ConnectWorldUpdateBegin(
+    boost::bind(&GazeboRosP3D::UpdateChild, this));
```
chapulina: You can change this callback to take `gazebo::common::UpdateInfo`, so you'll get the sim time as an input parameter. This also means there's no more need to keep the `world_` pointer. [Here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/src/gazebo_ros_joint_state_publisher.cpp#L140)'s an example.

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  // Rate control
+  if (impl_->update_rate_ > 0 &&
+    (current_time - impl_->last_time_).Double() < (1.0 / impl_->update_rate_))
+  {
+    return;
+  }
+
+  if (impl_->ros_node_->count_subscribers(impl_->topic_name_) > 0) {
```
chapulina: How about returning here if `impl_->ros_node_->count_subscribers(impl_->topic_name_) == 0`? This should make the code a bit more readable.

chapulina: The same goes for `tmp_dt == 0`, doing this would reduce 2 levels of indentation

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+        // Apply xyz offsets and get position and rotation components
+        pose.Pos() = pose.Pos() + impl_->offset_.Pos();
+        // apply rpy offsets
+        pose.Rot() = impl_->offset_.Rot() * pose.Rot();
+        pose.Rot().Normalize();
+
+        // Compute accelerations (not used)
+        impl_->apos_ = (impl_->last_vpos_ - vpos) / tmp_dt;
+        impl_->aeul_ = (impl_->last_veul_ - veul) / tmp_dt;
```
chapulina: Yeah it looks like `apos_`, `aeul_`, `frame_apos_` and `frame_aeul_` can be completely removed.

gazebo_plugins/src/gazebo_ros_p3d.cpp
```diff
@@ -0,0 +1,410 @@
...
+  /// \brief The body of the frame to display pose, twist
+  gazebo::physics::LinkPtr reference_link_;
+
+  /// \brief pointer to ros node
+  gazebo_ros::Node::SharedPtr ros_node_;
+
+  rclcpp::Publisher<nav_msgs::msg::Odometry>::SharedPtr pub_;
+
+  /// \brief ros message
+  nav_msgs::msg::Odometry pose_msg_;
```
chapulina: Is there a reason to keep this as a member variable? Can't a new message be created at every update? It doesn't look like we need past values or anything like that.

chapulina (changes requested):
Thank you again for the PR, @mjeronimo ! I know I left a lot of comments, but that's mainly our fault, not yours. A lot has changed from ROS 1 to ROS 2, and although we've been writing migration guides for downstream developers, we're lacking a guide specific for contributors like you. I'm going to work on this soon.

Meanwhile, besides the comments I left:

- [x] Please remove the files from `.ros1_unported` which you've ported
- [x] Please add a basic test for the plugin. You should be able to start from one of the [tests for plugins already ported](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/ros2/gazebo_plugins/test).
- [x] Also add a demo world with instructions similar to the ones [here](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/ros2/gazebo_plugins/worlds).
- [x] Add usage documentation to the header, see the [other plugins](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/ros2/gazebo_plugins/include/gazebo_plugins) for examples.
- [x] Add a migration guide for downstream users to the [wiki](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki).

I realize this seems like a lot, but hopefully the existing plugins will take you a long way. Don't hesitate to ask questions.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Removed.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Removed.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Removed.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Removed.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Updated to not be a member variable.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_p3d.hpp
mjeronimo: Sure, updated to follow that convention.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Good idea. Previously, I was concerned about a straight port and not improvements. Done.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Removed.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Removed altogether since it was only an alias for _parent.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: I've removed these new members with other changes, so this is no longer applicable.

gazebo_plugins/src/gazebo_ros_p3d.cpp
mjeronimo: Done; world_ pointer removed.

chapulina:
@mjeronimo , I went ahead and tackled the remaining items in the list:

* Rebased with latest `ros2`
* Added a test in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/845/commits/662081ab921601525535bb44630732ecc339a95b
* Added a demo world in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/845/commits/662081ab921601525535bb44630732ecc339a95b
* Added usage documentation in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/845/commits/662081ab921601525535bb44630732ecc339a95b
* Added a migration guide: [ROS 2 Migration: P3D](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-P3D)

Let me know if you have any concerns, I'm ready to merge once you approve the latest changes.

By the way, I also added a contribution guide [here](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-migration-contribution-guide).

chapulina (approved):

mjeronimo:
@chapulina  Thanks! Looks good to me. I was planning to get back to this after vacation (just back today), but glad to see you've taken care of it.

PR #854, opened 2018-12-12
[ros2] ENABLE_DISPLAY_TESTS, and make camera tests more robust

chapulina:
This way, to build camera test one needs to run:

`colcon build --packages-select gazebo_plugins --cmake-args -DENABLE_DISPLAY_TESTS=true`

Camera tests will be build in the OSRF build farm, but not on the ROS build farm.

### DIFF
--- a/gazebo_plugins/test/CMakeLists.txt
+++ b/gazebo_plugins/test/CMakeLists.txt
@@ -1,6 +1,8 @@
 find_package(ament_cmake_gtest REQUIRED)
 find_package(cv_bridge REQUIRED)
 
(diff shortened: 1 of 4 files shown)


mjcarroll (approved):

j-rivero:
I'm stiff fighting with the failing test, unrelated to this PR.

j-rivero (approved):

chapulina:
Last night `test_node` was also failing locally for me when I built rclcpp from source. The executor's thread is never joining in case there's more than one non-static node. I debugged for a long time and I'm still not sure how to fix it, whether this has been recently introduced, and why it passes on the ROS farm...

PR #859, opened 2018-12-14
[ros2] Add check if rclcpp::init() has been called

otamachan:
Hi,

I would like to use both ``gazebo_ros_init`` and ``gazebo_ros_factory`` plugins at the same time. But when I tried that, it failed.

```bash
tamaki@pc:~$ gzserver --verbose -s libgazebo_ros_init.so -s libgazebo_ros_factory.so
Gazebo multi-robot simulator, version 9.0.0
Copyright (C) 2012 Open Source Robotics Foundation.
Released under the Apache 2 License.
http://gazebosim.org

terminate called after throwing an instance of 'rclcpp::ContextAlreadyInitialized'
  what():  context is already initialized
Aborted (core dumped)
```
This PR adds check if ``rclcpp::init()`` has been called before calling it.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_factory.cpp
+++ b/gazebo_ros/src/gazebo_ros_factory.cpp
@@ -106,10 +106,8 @@ GazeboRosFactory::~GazeboRosFactory()
 {
 }
 
-void GazeboRosFactory::Load(int argc, char ** argv)
+void GazeboRosFactory::Load(int /* argc */, char ** /* argv */)
 {
-  rclcpp::init(argc, argv);
-
   // Keep this in the constructor for performance.
   // sdf::initFile causes disk access.
   sdf::initFile("root.sdf", impl_->factory_sdf_);
--- a/gazebo_ros/src/gazebo_ros_init.cpp
+++ b/gazebo_ros/src/gazebo_ros_init.cpp
@@ -50,9 +50,15 @@ GazeboRosInit::~GazeboRosInit()
 void GazeboRosInit::Load(int argc, char ** argv)
 {
   // Initialize ROS with arguments
-  rclcpp::init(argc, argv);
-
-  impl_->ros_node_ = gazebo_ros::Node::Get();
+  if (!rclcpp::is_initialized()) {
+    rclcpp::init(argc, argv);
+    impl_->ros_node_ = gazebo_ros::Node::Get();
+  } else {
+    impl_->ros_node_ = gazebo_ros::Node::Get();
+    RCLCPP_WARN(impl_->ros_node_->get_logger(),
+      "gazebo_ros_init didn't initialize ROS "
+      "because it's already initialized with other arguments");
+  }
 
   // Offer transient local durability on the clock topic so that if publishing is infrequent (e.g.
   // the simulation is paused), late subscribers can receive the previously published message(s).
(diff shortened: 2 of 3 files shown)


gazebo_ros/src/gazebo_ros_factory.cpp
```diff
@@ -108,7 +108,9 @@ GazeboRosFactory::~GazeboRosFactory()
 
 void GazeboRosFactory::Load(int argc, char ** argv)
 {
-  rclcpp::init(argc, argv);
+  if (!rclcpp::is_initialized()) {
+    rclcpp::init(argc, argv);
+  }
```
chapulina: Good call! It may also be worth it adding a check to `gazebo_ros_init`, and in both cases, printing a warning if it's been initialized, so the user knows their arguments won't be loaded.

chapulina (commented):
Thanks for catching the problem and making a PR!

The new test failed on the build farm, it looks like the message is published too late:

~~~
19:54:56 12: [0m/tmp/ws/src/gazebo_ros_pkgs/gazebo_ros/test/test_plugins.cpp:63: Failure
19:54:56 12: Expected: (msg) != (nullptr), actual: 16-byte object <00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00> vs 8-byte object <00-00 00-00 00-00 00-00>
19:54:56 12: test
19:54:57 12: [WARN] [hello_ros_world]: Publishing
~~~

gazebo_ros/src/gazebo_ros_factory.cpp
otamachan: OK. I'll add a check to ``gazebo_ros_init``, too.
I think using both ``gazebo_ros_init`` and ``gazebo_ros_factory`` is a common use case for ROS users. In that case it might be annoying when a warning is always printed. Is that OK?

chapulina: Ok, let's go without the warning, thanks!

otamachan:
Thank you for your reviewing.

1. [Test failure](http://build.ros2.org/job/Cpr__gazebo_ros_pkgs__ubuntu_bionic_amd64/13/testReport/gazebo_ros/Plugins_TestPlugins/TestTopicsReceived_4__48_byte_object__30_6C_DA_E6_5B_55_00_00_58_6C_DA_E6_5B_55_00_00_58_6C_DA_E6_5B_55_00_00_00_6C_DA_E6_5B_55_00_00_20_6C_DA_E6_5B_55_00_00_20_6C_DA_E6_5B_55_00_00_______________________/)
I could reproduce the test failure locally, but after upgrading ``rclcpp`` to master branch it never happens. So it might not be related to this change. (https://github.com/ros2/rclcpp/pull/608 ?)
The test is still [failing](http://build.ros2.org/job/Cpr__gazebo_ros_pkgs__ubuntu_bionic_amd64/16/testReport/junit/(root)/gazebo_ros/) but https://github.com/ros-simulation/gazebo_ros_pkgs/pull/860 will solve the problem.
2. I removed calling ``rclcpp::init()`` from ``gazebo_ros_factory`` rather than checking and calling because ``gazebo_ros::Node::Get()`` will call ``rclcpp::init()`` automatically.
3. I added warning to ``gazebo_ros_init`` if ``rclcpp::init`` is already called as I think ``gazebo_ros_init`` should be the only way to call ``rclcpp::init`` with user defined arguments.

Thanks for your advice in advance.

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -50,9 +50,15 @@ GazeboRosInit::~GazeboRosInit()
...
 {
   // Initialize ROS with arguments
-  rclcpp::init(argc, argv);
-
-  impl_->ros_node_ = gazebo_ros::Node::Get();
+  if (!rclcpp::is_initialized()) {
+    rclcpp::init(argc, argv);
+    impl_->ros_node_ = gazebo_ros::Node::Get();
+  } else {
+    impl_->ros_node_ = gazebo_ros::Node::Get();
```
chapulina, line 57: nitpick: this line can be moved out of the `if`/`else`

chapulina (approved):
Thanks for iterating, this looks good to me.

PR #862, opened 2018-12-24
Add option to publish depth images in 16UC1

awesomebytes:
Given there are sensors in robots that publish depth images natively in the `16UC1` format, it is convenient to be able to simulate it instead of converting manually later on from `32FC1`.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
@@ -181,6 +181,7 @@ namespace gazebo
     private: int depth_info_connect_count_;
     private: void DepthInfoConnect();
     private: void DepthInfoDisconnect();
+    private: bool use_depth_image_16UC1_format_;
 
     // overload with our own
     private: common::Time depth_sensor_update_time_;
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_openni_kinect.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_openni_kinect.h
@@ -134,6 +134,7 @@ namespace gazebo
     private: void DepthInfoConnect();
     private: void DepthInfoDisconnect();
     private: common::Time last_depth_image_camera_info_update_time_;
+    private: bool use_depth_image_16UC1_format_;
     protected: ros::Publisher depth_image_camera_info_pub_;
 
     using GazeboRosCameraUtils::PublishCameraInfo;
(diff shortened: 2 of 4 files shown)


awesomebytes:
I don't know why `ros_gazebo_pkgs-ci-pr_any_kinetic-xenial-amd64` is failing on the tests:

* DistortionTest.barrelDistortion
* DistortionTest.pincushionDistortion
* TriggeredCameraTest.cameraSubscribeTest

Given I haven't changed anything related to them (I think). Let me know if I can do anything to fix it.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
```diff
@@ -138,6 +138,7 @@ namespace gazebo
     private: int depth_info_connect_count_;
     private: void DepthInfoConnect();
     private: void DepthInfoDisconnect();
+    private: bool use_depth_image_16UC1_format_;
```
chapulina, line 184: Adding a private member variable will break ABI, so this would have to target Noetic :confused:

j-rivero: Given the nature of ROS we don't need to keep ABI stable since the buildfarm will recompile all the dependencies on top. For `gazebo_ros_pkgs` we need to keep API, topic/services layout and behavior but I think that we can safely break the ABI.

chapulina: Ok, I misunderstood it then. I thought that we were trying to keep ABI stable for the packages which are not on the ROS build farm, like the ones on http://packages.osrfoundation.org/ and other custom packages that are built downstream on top of ours.

j-rivero: not breaking the ABI is a nice to have but I think that in the ROS1 world it should not be a limitation and external packages needs to be adapted (recompiled) according.

awesomebytes: Soo... Can we merge then?

sloretz:
Closing/reopening to see if github sorts out that only 3 commits need to be added

sloretz:
@awesomebytes I rebased your branch onto `noetic-devel`, hope you don't mind.

awesomebytes:
@sloretz sounds great! I'd love to see this merged :)

chapulina (approved):
Thanks for waiting to get it into Noetic. LGTM.

PR #868, opened 2019-01-08
[ros2] Port get/set properties services

nzlz:
First approach to the migration of get/set properties. 
As I stated in https://github.com/ros-simulation/gazebo_ros_pkgs/issues/779 I am trying to migrate the _properties_ related services to ros2.

This a World plugin. I tried to follow the same structure as `ros2_state` branch.
`get_physics_properties` and `set_physics_properties` are commented as @ironmig stated in https://github.com/ros-simulation/gazebo_ros_pkgs/issues/779:

> Remove the get_physics_properties and set_physics_properties services, which duplicate the functionality of the physics dynamic reconfigure in the current plugin. Replace with parameters for physics properties

Also seems like `get_world_properties` needs some change too:
> Remove get_world_properties. Replace with get_model_list (perhaps under factory?). The other attributes are sim_time (will come from clock) and rendering_enabled (currently unused)

**Note**: This is just a first version, the intend is not to merge but to discuss the structure of these functionalities.
**Note2**: I will be off a few days, so feel free to reuse/update this code in any way.


gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,659 @@
...
+/*  impl_->set_physics_properties_service_ =
+    impl_->ros_node_->create_service<gazebo_msgs::srv::SetPhysicsProperties>(
+    "set_physics_properties", std::bind(&GazeboRosPropertiesPrivate::SetPhysicsProperties, impl_.get(),
+    std::placeholders::_1, std::placeholders::_2));
+*/
+}
+
+void GazeboRosPropertiesPrivate::GetWorldProperties(
+	gazebo_msgs::srv::GetWorldProperties::Request::SharedPtr _req,
+	gazebo_msgs::srv::GetWorldProperties::Response::SharedPtr _res)
```
chapulina: I agree with @ironmig that `GetWorldProperties` has a misleading name and if we want to keep the functionality we should rename it to `GetModelList`.

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,659 @@
...
+
+    //TODO, is this the correct way to do this?
+    auto light_modify_pub_ = ros_node_->create_publisher<gazebo::msgs::Light>("~/light/modify");
+    light_modify_pub_->publish(light);
+
+    _res->success = true;
+  }
+}
+
+/*void GazeboRosPropertiesPrivate::SetPhysicsProperties(
```
chapulina: Yes, I agree that it's good to avoid duplication, and porting just the dynamic reconfigure part should be enough. Note that on ROS 2 that would be implemented with basic parameters, I think you'll need to register a callback with [ros_node_->register_param_change_callback](http://docs.ros2.org/beta3/api/rclcpp/classrclcpp_1_1node_1_1Node.html#a0a49dab3e2d5c42ccab0c5d674c127e5).

chapulina (commented):
Thank you for getting started on this! The general direction looks good to me, I'm happy to review this again once you make more progress.

Also, I think you've already seen this, but just in case, [here](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-migration-contribution-guide)'s a guide that should help a bit. For example, you can remove all references to Gazebo <= 8.

nzlz (commented):
Review.

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,603 @@
...
+    _res->success = true;
+    _res->status_message = "GetJointProperties: got properties";
+  }
+}
+
+void GazeboRosPropertiesPrivate::GetLinkProperties(
+	gazebo_msgs::srv::GetLinkProperties::Request::SharedPtr _req,
+	gazebo_msgs::srv::GetLinkProperties::Response::SharedPtr _res)
+{
+  gazebo::physics::LinkPtr body = boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->EntityByName(_req->link_name));
```
nzlz: Regarding deboosting.. which would be a correct update for this line? 

I guess this is not related to the usage of the new [conversion](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/ros2/gazebo_ros/include/gazebo_ros/conversions) functions.

chapulina: This use of boost is coming from Gazebo, so it can't be removed. I added a [note](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-migration-contribution-guide/_compare/828dbbc519cd1bfcdb34d8d739d50a08c58a7920...6faaeed88f3ab2a70060a2434e1c3c58ee7f0874) to the tutorial.

nzlz:
Regarding the required tests, I guess the services that have a get/set are pretty straight forward to implement. What about the test for `get_model_properties` ?

Is it a good idea to spawn a custom xml file and expect the `get_model_properties` to return equal values? Since this does not seem a very good idea, how could this be tested in another way?

nzlz:
Continuing the test discussion:
Get/Set Link properties and Get/Set Light properties are easy to test since they use the same parameters in the respective services. But in **Get/Set Joint properties**, service request/response differs so much. I think this needs some kind of rework, either in the request or the response part (I guess the Request). 

_GetJointProperties.srv_ [Response]
```
string joint_name                    # name of joint
---
# joint type
uint8 type
uint8 REVOLUTE    = 0                # single DOF
uint8 CONTINUOUS  = 1                # single DOF (revolute w/o joints)
uint8 PRISMATIC   = 2                # single DOF
uint8 FIXED       = 3                # 0 DOF
uint8 BALL        = 4                # 3 DOF
uint8 UNIVERSAL   = 5                # 2 DOF
# dynamics properties
float64[] damping
# joint state
float64[] position
float64[] rate
# service return status
bool success                         # return true if get successful
string status_message                # comments if available
```

_SetJointProperties.srv_ [Request]
```
string joint_name                               # name of joint
gazebo_msgs/ODEJointProperties ode_joint_config # access to ODE joint dynamics properties
---
bool success                                    # return true if get successful
string status_message                           # comments if available

```
_ODEJointProperties_ message
```
# access to low level joint properties, change these at your own risk
float64[] damping             # joint damping
float64[] hi_stop             # joint limit
float64[] lo_stop             # joint limit
float64[] erp                 # set joint erp
float64[] cfm                 # set joint cfm
float64[] stop_erp            # set joint erp for joint limit "contact" joint
float64[] stop_cfm            # set joint cfm for joint limit "contact" joint
float64[] fudge_factor        # joint fudge_factor applied at limits, see ODE manual for info.
float64[] fmax                # ode joint param fmax
float64[] vel                 # ode joint param vel

```

nzlz:
~~Another thing, both set_entity_state and and set_link_properties offer the possibility to change the pose. Should the pose-changing option be available only in the state-related service? Or is it okay is it is duplicated (as it was in ros1).~~

Note: updated PR commit.

Update: Confusion between different pose elements. Leaving as it was in ros1 version.

chapulina:
> Is it a good idea to spawn a custom xml file and expect the get_model_properties to return equal values?

Probably the easiest way would be to have the model already in the world instead of going through the trouble of spawning it. For example, [this test world](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/test/worlds/gazebo_ros_joint_state_publisher.world) has a simple model with a joint.

> both set_entity_state and and set_link_properties offer the possibility to change the pose. Should the pose-changing option be available only in the state-related service? Or is it okay is it is duplicated (as it was in ros1).

I vote for removing duplication and having a clear separation of concerns across the plugins.

> But in Get/Set Joint properties, service request/response differs so much. I think this needs some kind of rework, either in the request or the response part

Very good point. This touches a broader topic about refactoring physics properties so that those which are common across physics engines are in a common message and only the engine specific ones are in separate messages. This subject is definitely larger than the scope of this pull request though, ideally we will eventually have some consistency across `gazebo_ros`, `SDF` and `ign-msgs`.

So I'd propose one of these solutions for now:

1. Create new services only with the bare minimum functionality you need right now, and leave the rest of the properties for later.
1. Just keep the same message structure until this can be revisited.
1. Work on a broader proposal for how to organize joint properties across physics engines.

nzlz:
Regarding `Get/Set Joint properties`, I will go with number 2, which is:

> Just keep the same message structure until this can be revisited.

The reason is that another package I have to migrate is using the ODE-> damping. Will keep this for now . I will spend some time with the new organization of joint properties across physics engines once the ros2 branch gets stable (merges _state, _time_cmds and upcoming _properties). Not vital thing right now.

nzlz:
Updated main commit right now. Current state. **UPDATED**:
- `get_world_properties` migrated to `get_model_list` inside `factory`. Seems like everything is working, including tests.
- `get_joint_properties` added. Working.
- `set_joint_properties` added. Working.
- `get_link_properties` added. Working.
- `set_link_properties` added. Working.
- `get_light_properties` added. Working.
- `set_light_properties` added. Working.

TODO:
- `get/set:physics properties.`

nzlz:
Hi @chapulina,

Pushed ready-to-merge code, ready for a new review, possibly final?. CI errors are fine on my side.

---
Regarding **dynamic physics reconfigure**: 

> Yes, I agree that it's good to avoid duplication, and porting just the dynamic reconfigure part should be enough. Note that on ROS 2 that would be implemented with basic parameters, I think you'll need to register a callback with ros_node_->register_param_change_callback.

`get/set:physics properties` not migrated yet since its not urgent for our needs. I could spend some time working on this in the future if you could provide extra information regarding this topic, but I don't think I it will be possible in the short term.

chapulina:
TODO(@nzlz ) : rebase against new `crystal` branch

PS: Sorry I haven't been able to make time to review the latest changes.

nzlz:
Rebased against `crystal`.

gazebo_msgs/srv/GetModelList.srv
```diff
@@ -0,0 +1,5 @@
+---
+float64 sim_time                     # current sim time
+string[] model_names                 # list of models in the world
+bool success                         # return true if get successful
+string status_message                # comments if available
```
chapulina: Mind removing the `status_message` field? It's something that is present in `gazebo_ros1` and we'd like to stop doing. I also added a note to the contribution guide. No need to remove it from the old messages you're using.

gazebo_ros/src/gazebo_ros_factory.cpp
```diff
@@ -44,6 +46,12 @@ class GazeboRosFactoryPrivate
   /// \param[in] _world_name The world's name
   void OnWorldCreated(const std::string & _world_name);
 
+  /// \brief Function for receiving the model list from a gazebo world.
+  /// \param[out] res Response
```
chapulina, line 52: Please document all fields, even if they're going to be ignored

gazebo_msgs/srv/GetModelList.srv
```diff
@@ -0,0 +1,5 @@
+---
+float64 sim_time                     # current sim time
```
chapulina: How about getting rid of this field and putting sim time in the header instead? Like this:

`_res->header.stamp = Convert<builtin_interfaces::msg::Time>(world_->SimTime());`

gazebo_ros/src/gazebo_ros_factory.cpp
```diff
@@ -19,7 +19,9 @@
 #include <gazebo/physics/Entity.hh>
 #include <gazebo/physics/PhysicsIface.hh>
 #include <gazebo/physics/World.hh>
+#include <gazebo/physics/Model.hh>
```
chapulina: Nitpick: keep includes in alphabetic order

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+#include <gazebo_msgs/srv/get_model_properties.hpp>
+#include <gazebo_msgs/srv/get_joint_properties.hpp>
+#include <gazebo_msgs/srv/get_link_properties.hpp>
+#include <gazebo_msgs/srv/get_light_properties.hpp>
+#include <gazebo_msgs/srv/get_physics_properties.hpp>
+#include <gazebo_msgs/srv/set_joint_properties.hpp>
+#include <gazebo_msgs/srv/set_link_properties.hpp>
+#include <gazebo_msgs/srv/set_light_properties.hpp>
+#include <gazebo_msgs/srv/set_physics_properties.hpp>
+#include <gazebo_ros/node.hpp>
```
chapulina, line 33: Alphabetize includes

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+  /// \brief ROS service to handle requests to set link properties.
+  rclcpp::Service<gazebo_msgs::srv::SetLinkProperties>::SharedPtr set_link_properties_service_;
+
+  /// \brief ROS service to handle requests to set light properties.
+  rclcpp::Service<gazebo_msgs::srv::SetLightProperties>::SharedPtr set_light_properties_service_;
+
+  /// Gazebo node for communication.
+  gazebo::transport::NodePtr gz_node_;
+
+  /// Publishes light factory messages.
```
chapulina, line 124: light factory -> light modify

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+      _req->model_name.c_str());
+    _res->success = false;
+    _res->status_message = "GetModelProperties: model does not exist";
+  } else {
+    // get model parent name
+    gazebo::physics::ModelPtr parent_model =
+      boost::dynamic_pointer_cast<gazebo::physics::Model>(model->GetParent());
+    if (parent_model) {_res->parent_model_name = parent_model->GetName();}
+
+    // get list of child bodies, geoms
```
chapulina, line 202: It's always been a bit confusing on ROS1 how the word `body` can be used to mean `link` and `geom` to mean `collision`. We have a chance to fix this in ROS2 by creating a new message.

Another confusing thing about that message is how links and collisions are listed side-by-side instead of nested.

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+    _res->body_names.clear();
+    _res->geom_names.clear();
+    for (unsigned int i = 0; i < model->GetChildCount(); ++i) {
+      gazebo::physics::LinkPtr body =
+        boost::dynamic_pointer_cast<gazebo::physics::Link>(model->GetChild(i));
+      if (body) {
+        _res->body_names.push_back(body->GetName());
```
chapulina, line 209: The message description suggests this will be the scoped name, which you can get with `GetScopedName`

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+  impl_->gz_node_->Init(_world->Name());
+  impl_->gz_properties_light_pub_ =
+    impl_->gz_node_->Advertise<gazebo::msgs::Light>("~/light/modify");
+}
+
+void GazeboRosPropertiesPrivate::GetModelProperties(
+  gazebo_msgs::srv::GetModelProperties::Request::SharedPtr _req,
+  gazebo_msgs::srv::GetModelProperties::Response::SharedPtr _res)
+{
+  gazebo::physics::ModelPtr model = world_->ModelByName(_req->model_name);
```
chapulina, line 189: We also need to fill the `canonical_body_name` field, or if the message is changed, `canonical_link_name`.

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+      break;
+    }
+  }
+
+  if (!joint) {
+    _res->success = false;
+    _res->status_message = "GetJointProperties: joint not found";
+  } else {
+    /// @todo: FIXME
+    _res->type = _res->REVOLUTE;
```
chapulina, line 264: It would be good to address this TODO

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+    _res->damping.clear();
+    _res->damping.push_back(joint->GetDamping(0));
```
chapulina, line 267: You can check how many axes the joint has and if it has 2, you can add `_res->damping.push_back(joint->GetDamping(1));` after this. The same goes for the other fields below.

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+    _res->ixy = inertia->IXY();
+    _res->ixz = inertia->IXZ();
+    _res->iyz = inertia->IYZ();
+
+    ignition::math::Vector3d com = body->GetInertial()->CoG();
+
+    _res->com.position.x = com.X();
+    _res->com.position.y = com.Y();
+    _res->com.position.z = com.Z();
+    _res->com.orientation.x = 0;  // @todo: gazebo do not support rotated inertia yet
```
chapulina, line 310: This is not true anymore, `auto com = body->GetInertial()->Pose()` should give you the translation and rotation of the CoG.

You can also use the conversion functions to go from ign-math to the ROS msg, something like:

~~~
  _res->com.position = gazebo_ros::Convert<geometry_msgs::msg::Point>(com.Pos());
  _res->com.orientation = gazebo_ros::Convert<geometry_msgs::msg::Quaternion>(com.Rot());
~~~

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+      _res->status_message = "getLightProperties: Requested light " + _req->light_name
+       + " not found!";
+  } else {
+    gazebo::msgs::Light light;
+    phy_light->FillMsg(light);
+
+    _res->diffuse.r = light.diffuse().r();
+    _res->diffuse.g = light.diffuse().g();
+    _res->diffuse.b = light.diffuse().b();
+    _res->diffuse.a = light.diffuse().a();
```
chapulina, line 336: Consider adding a conversion function from `gazebo::msgs::Color` to `std_msgs::ColorRGBA`, not required though

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+    for(unsigned int i = 0;i< _req->ode_joint_config.stop_erp.size();i++)
+      joint->SetParam("stop_erp", i, _req->ode_joint_config.stop_erp[i]);
+    for(unsigned int i = 0;i< _req->ode_joint_config.stop_cfm.size();i++)
+      joint->SetParam("stop_cfm", i, _req->ode_joint_config.stop_cfm[i]);
+    for(unsigned int i = 0;i< _req->ode_joint_config.fudge_factor.size();i++)
+      joint->SetParam("fudge_factor", i, _req->ode_joint_config.fudge_factor[i]);
+    for(unsigned int i = 0;i< _req->ode_joint_config.fmax.size();i++)
+      joint->SetParam("fmax", i, _req->ode_joint_config.fmax[i]);
+    for(unsigned int i = 0;i< _req->ode_joint_config.vel.size();i++)
+      joint->SetParam("vel", i, _req->ode_joint_config.vel[i]);
```
chapulina, line 390: humm `vel` doesn't seem to work for me. Not an issue with this PR though, probably a problem on Gazebo.

gazebo_ros/worlds/gazebo_ros_properties_demo.world
```diff
@@ -0,0 +1,676 @@
...
+
+      ros2 service call /demo/get_light_properties 'gazebo_msgs/GetLightProperties' '{light_name: "sun"}'
+
+  SETTING PROPERTIES
+
+  Try modifying the damping of a joint:
+      ros2 service call /demo/set_joint_properties 'gazebo_msgs/SetJointProperties' '{joint_name: "simple_arm::arm_shoulder_pan_joint", ode_joint_config: {damping: {1.0}}}'
+
+  Or a links properties:
+      ros2 service call /demo/set_link_properties 'gazebo_msgs/SetLinkProperties' '{link_name: "simple_arm::arm_base", mass: 150.0 , izz: 2.0}'
```
chapulina, line 29: This is fun :smile:

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+    // mass.SetInertiaRotation(ignition::math::Quaternion(_req->com.orientation.w,
+    // _res->com.orientation.x,_req->com.orientation.y _req->com.orientation.z));
+    mass->SetCoG(ignition::math::Vector3d(
+      _req->com.position.x,
+      _req->com.position.y,
+      _req->com.position.z));
+    mass->SetInertiaMatrix(_req->ixx, _req->iyy, _req->izz, _req->ixy, _req->ixz, _req->iyz);
+    mass->SetMass(_req->mass);
+    body->SetGravityMode(_req->gravity_mode);
+    // @todo: mass change unverified
```
chapulina: Your demo shows that this works, so this todo can be removed

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,448 @@
...
+    _res->status_message = "SetLinkProperties: properties set";
+  }
+}
+
+void GazeboRosPropertiesPrivate::SetLightProperties(
+  gazebo_msgs::srv::SetLightProperties::Request::SharedPtr _req,
+  gazebo_msgs::srv::SetLightProperties::Response::SharedPtr _res)
+{
+  gazebo::physics::LightPtr phy_light = world_->LightByName(_req->light_name);
+  if (phy_light == NULL) {
```
chapulina: Please use `nullptr` instead of `NULL`

chapulina (changes requested):
Sorry for the delay! I did another pass on the PR.

One thing that stood out to me is how the messages use outdated fields, like `body` and `geom` to refer to `link` and `collision`. It would be really nice to update those, but I understand if this is too much trouble.

I started fixing some linter errors on cb891db90eebac84bb6c45e38f2ad084920bc927, but could you take care of the rest? Make sure there are no failures when you run `colcon test --packages-select gazebo_ros`.

Great demo and tests!

chapulina:
This has been merged into `dashing` on #972 . Closing this PR, reopen it a backport for `crystal` is necessary.

PR #874, opened 2019-01-18
add more light options to service

christian-rauch:
This PR adds `cast_shadows`, `specular`, `direction` and `pose` to the `SetLightProperties` service.

### DIFF
--- a/gazebo_msgs/srv/SetLightProperties.srv
+++ b/gazebo_msgs/srv/SetLightProperties.srv
@@ -1,8 +1,12 @@
 string light_name                    # name of Gazebo Light
+bool cast_shadows
 std_msgs/ColorRGBA diffuse           # diffuse color as red, green, blue, alpha
+std_msgs/ColorRGBA specular          # specular color as red, green, blue, alpha
 float64 attenuation_constant
 float64 attenuation_linear
 float64 attenuation_quadratic
+geometry_msgs/Vector3 direction
+geometry_msgs/Pose pose              # pose in world frame
 ---
 bool success                         # return true if get successful
 string status_message                # comments if available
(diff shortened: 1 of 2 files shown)


christian-rauch:
This PR is just meant for `melodic`. Why does the CI test if this can be merged into `kinetic`? It will obviously fail because the melodic and kinetic branches diverged.

chapulina (changes requested):
This looks like a great improvement to me. The only problem I see is that we can't tell whether the fields are set or not on the ROS message. So if we release this into Melodic, some values would be zeroed for existing users. i.e. specular would be set to black, cast shadows turned off, the light teleported to the world origin...

My recommendation would be to release this into Noetic instead, and make sure the new fields are documented on the commit message so users see them on the changelog.

A possible solution for Melodic would be to treat zero / false as unset, but that would limit the range of values, and doesn't solve the boolean case.

christian-rauch:
Does the protobuf message store if a value has been set? Because otherwise, these values should be default initialised to 0 anyway. We could reproduce this behaviour with NaNs in place of "unset" values. 

But I agree that this should be postponed to the next ROS released then. Shall I rebase this PR to the `noetic-devel` branch then?

chapulina:
> Does the protobuf message store if a value has been set?

Yes, `gazebo::msgs` still uses protobuf 2, so there's support for optional fields. You can check if a field is set with `light.has_cast_shadows()` for example. From protobuf 3 this can only be checked for nested messages though.

>  Shall I rebase this PR to the noetic-devel branch then?

That would be great, thanks!

chapulina:
Thank you for rebasing. We need a way to document this so users can find out why their lights that used to stay in place on Melodic are teleported to the origin on Noetic when they change the light's color. This package doesn't have a migration guide though. Do you mind adding some notes about this to the commit message so it shows up on the changelog? Thanks!

christian-rauch:
> Do you mind adding some notes about this to the commit message so it shows up on the changelog? Thanks!

Done. I mentioned that these newly added ROS message fields will be initialised to 0 by default.

chapulina (approved):

PR #875, opened 2019-01-22
gazebo_plugins/added zero mean gaussian noise to the kinect plugin

chisarie:
### Intro
Hi, I am a developer working with a depth camera that we simulate using the kinect plugin. When simulating the depth camera, gaussian noise is useful in order to have a more realistic representation of the sensor.

### Gaussian noise
The implementation is a zero mean multiplicative Gaussian Noise. The way this is implemented was taken from the gaussian noise representation in the [velodyne_gazebo_plugins](https://bitbucket.org/DataspeedInc/velodyne_simulator/src/master/velodyne_gazebo_plugins/) package, which is under BSD license. The author does not intend to brake any Copyright; if it happened to be the case with this pull request, please decline it.

### Usage
When calling the plugin from your .xacro file, as explained in this [tutorial](http://gazebosim.org/tutorials?tut=ros_depth_camera&cat=connect_ros), simply add the parameter `<gaussianNoise>0.01</gaussianNoise>` (or whatever other value you prefer); this parameter is the standard deviation of the gaussian noise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@
 .DS_Store
 
 CATKIN_IGNORE
+
+# VSCode
+.vscode
+*/build/*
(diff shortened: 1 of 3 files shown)


YoshuaNava:
This is a very useful addition.

@j-rivero, what is your opinion on this PR?

PR #878, opened 2019-02-01
[Windows][melodic] Use portable random generator instead of rand_r

seanyen:
Use ~c++11~ a portable random generator for better cross-platform.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_ft_sensor.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_ft_sensor.h
@@ -88,7 +88,7 @@ class GazeboRosFT : public ModelPlugin
 
   /// \brief Gaussian noise
   private: double gaussian_noise_;
-  private: unsigned int seed;
+
   /// \brief Gaussian noise generator
   private: double GaussianKernel(double mu, double sigma);
 
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gpu_laser.h
@@ -86,7 +86,6 @@ namespace gazebo
     private: sdf::ElementPtr sdf;
     private: void LoadThread();
     private: boost::thread deferred_load_thread_;
-    private: unsigned int seed;
 
     private: gazebo::transport::NodePtr gazebo_node_;
     private: gazebo::transport::SubscriberPtr laser_scan_sub_;
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_imu.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_imu.h
@@ -121,7 +121,6 @@ namespace gazebo
     private: sdf::ElementPtr sdf;
     private: void LoadThread();
     private: boost::thread deferred_load_thread_;
-    private: unsigned int seed;
 
     // ros publish multi queue, prevents publish() blocking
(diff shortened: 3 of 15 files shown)


seanyen:
Looking into why the test failed.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ft_sensor.h
```diff
@@ -88,7 +90,7 @@ class GazeboRosFT : public ModelPlugin
 
   /// \brief Gaussian noise
   private: double gaussian_noise_;
-  private: unsigned int seed;
+  private: std::default_random_engine seed;
```
chapulina: These changes break ABI. Do you think you could make this patch ABI-compatible?

chapulina:
> Looking into why the test failed.

@seanyen , have you figured this out? I'll trigger fresh builds.

chapulina:

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ft_sensor.h
seanyen: Let me see what I can do here.

j-rivero: after speaking with @chapulina we think that although breaking ABI is something that we try to avoid, the ROS policy allows it and we should not stop changes specially if not breaking ABI involves too many effort. We thinks is fine to break it in this case. Thanks anyway Sean.

j-rivero:

j-rivero:

j-rivero:

j-rivero:

j-rivero:

seanyen:
@j-rivero @chapulina I made a chance to use `std::mt19937` instead of `std::default_random_engine` as the pseudo-random generator, which seemed to make the test failure going away. However I am not really comprehensively understanding why it is that.

The information currently I know are:
* It seems to me that the `test_range.py` test failure is not reproducible on Windows build.
* `MSVC runtime library` is using `std::mt19937` for `std::default_random_engine` by default, but `libstdc++` is using `std::minstd_rand0`.

Are you aware of any C++11 random number generator issues on `libstdc++`? And how does the expected range (0.22~0.25) get defined in `test_range.py`? I might need more insights on the test cases so I can dive deeper on this.

chapulina:
> It seems to me that the test_range.py test failure is not reproducible on Windows build.

It looks like all tests are passing now on CI. Are you getting test failures locally?

> MSVC runtime library is using std::mt19937 for std::default_random_engine by default, but libstdc++ is using std::minstd_rand0

We could use [ignition::math::Rand](https://bitbucket.org/ignitionrobotics/ign-math/src/default/include/ignition/math/Rand.hh), which uses `std::mt19937` and supposedly works consistently across all platforms.

seanyen:
@chapulina @j-rivero I changed it to use `ignition::math::Rand` and now all the test failure are clear. Thanks!

chapulina (approved):
LGTM

We could avoid the ABI break by not removing the private seed variables, but as @j-rivero said, we're not being strict about it on ROS 1.

The `math/Rand` includes could also be moved to the `cpp` files since they're not needed in the headers.

But these are just nitpicks. Approving!

seanyen:
@chapulina @j-rivero I rebased the pull request to the latest, cleanup the commit messages and move the new header inclusion into cpp. Let me know anything from me to move this forward. Thanks!

PR #879, opened 2019-02-01
[Windows][melodic] Windows implementation for setenv.

seanyen:
setenv is a POSIX function and which is not provided by MSVC runtime library. Here is to add an implementation for Windows build.


chapulina (approved):
Trusting @seanyen on Windows matters :crossed_fingers:

PR #882, opened 2019-02-05
[ros2] Fix diff_drive error message

romainreignier:
Small fix to add missing `c_str()` method of `std::string` in printf style log.


chapulina (approved):
Nice catch!

Issue #884, opened 2019-02-06
[ros2] Build errors in Windows


jnoyola:
I'm not sure if there's much Windows development going on here, but the ros2 branch _almost_ builds in Windows 10, apart from a few conflicting `#define`s and missing `usleep`. I would open a PR, but I think my fixes might be a little hacky so I wanted to get some opinions.

1. `OUT` [is a macro in Windows](https://stackoverflow.com/questions/35907493/in-and-out-macros-in-minwindef-h), so I simply changed all template tparams from `OUT` to `T`.
2. ROS2 and Gazebo each build fine on their own in Windows, but we see an issue with connecting them here because Gazebo includes `<windows.h>` and ROS uses `NO_ERROR` as an enum value, which is a macro in Windows. I think it should be the consumer's responsibility to make sure they don't clash, so I `#undef NO_ERROR` before including and tf2 headers from ROS.
3. `usleep` does not exist in Windows (luckily gazebo_ros_pkgs was only relying on unistd.h indirectly, so no issue there). In gazebo_ros/utils.hpp I added a simple macro for `usleep` in Windows.
4. I also noticed that ImuSensorPlugin isn't linked to gazebo_ros_imu_sensor, but I don't think that's related to Windows.

If it would help, I can get a branch with these changes.

chapulina:
Thanks for all the pointers, a pull request would be appreciated

jnoyola:
Thanks for the quick review! Building successfully for me on Windows 10 now

PR #885, opened 2019-02-06
Fix Windows conflicting macros and missing usleep

jnoyola:
Addressing the issue mentioned here: #884 
`<windows.h>` is included indirectly via gazebo, which causes conflicts within gazebo_ros_packages and in included ros2 headers.

1. `OUT` [is a macro in Windows](https://stackoverflow.com/questions/35907493/in-and-out-macros-in-minwindef-h) and used in templates in gazebo_ros. Fixed by changing to `T`.
2. `NO_ERROR` [is a macro in Windows](https://github.com/ros/geometry2/issues/172) and used in tf2 in ros2. Fixed with `#undef`.
3. `usleep` does not exist in Windows. Fixed with macro.
4. ImuSensorPlugin unresolved external symbols. Fixed by linking to gazebo_ros_imu_sensor.

### DIFF
--- a/gazebo_plugins/CMakeLists.txt
+++ b/gazebo_plugins/CMakeLists.txt
@@ -115,6 +115,9 @@ ament_target_dependencies(gazebo_ros_imu_sensor
   "sensor_msgs"
   "gazebo_dev"
 )
+target_link_libraries(gazebo_ros_imu_sensor
+  ImuSensorPlugin
+)
 ament_export_libraries(gazebo_ros_imu_sensor)
 
 # gazebo_ros_ray_sensor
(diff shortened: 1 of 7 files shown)


jnoyola:
Could someone help me figure out what's going on with the latest test failures? I'm not super familiar with Jenkins, but I see there's [only one that's a fresh failure](http://build.ros2.org/job/Cpr__gazebo_ros_pkgs__ubuntu_bionic_amd64/35/testReport/junit/(root)/gazebo_ros/test_node_gtest_missing_result/), but I'm not sure where to go from here. Thanks!

chapulina (approved):
`test_node` is a known test failure which #860 is supposed to fix, so don't worry about it.

Thanks for the fixes!

PR #887, opened 2019-02-11
Fix build to account for new NodeOptions interface.

ghost:
[ros2/rclcpp#622](https://github.com/ros2/rclcpp/pull/622) changed the Node class constructor interface, breaking the build. This PR updates the code to build with that latest change.

### DIFF
--- a/gazebo_ros/src/node.cpp
+++ b/gazebo_ros/src/node.cpp
@@ -77,14 +77,12 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
     }
   }
 
-  // Use default context
-  auto context = rclcpp::contexts::default_context::get_global_default_context();
+  rclcpp::NodeOptions node_options;
+  node_options.arguments(arguments);
+  node_options.initial_parameters(initial_parameters);
 
   // Create node with parsed arguments
-  return CreateWithArgs(
-    name, ns,
-    context,
-    arguments, initial_parameters);
+  return CreateWithArgs(name, ns, node_options);
 }
 
 Node::SharedPtr Node::Get()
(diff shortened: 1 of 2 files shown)


chapulina:
Thanks for the PR. This will break compatibility with Crystal, so we'll need to create a separate branch. I can handle that if you just confirm that is the case. Thanks!

ghost:
Yes. This will break the Crystal build. It looks like this interface change comes in with the Dashing release.

chapulina:
I'm trying to get CI running for this. But we'll need this merged first: https://github.com/ros-perception/image_common/pull/120

chapulina:
Got CI to build, but there are a few test failures. I haven't checked if the tests are caused by the move to Dashing or by the different CI setup.

[Build Status](https://ci.ros2.org/job/ci_linux/6511/)

---

For future reference, this is how someone with access to ci.ros2.org would run CI:

* Use this job: https://ci.ros2.org/job/ci_linux/
* Use a `ros2.repos` supplement like [this one](https://bitbucket.org/!api/2.0/snippets/chapulina/AeBk5n/410fc2655b33d8b6a3ce893e5f707967784137ea/files/ros2.repos), which has `gazebo_ros_pkgs`, `image_common` and `vision_opencv`.
* Use the [gazebo_ros_pkgs branch](https://github.com/ros2/ci/compare/master...gazebo_ros_pkgs) from `ros2/ci`so that all dependencies are installed, and other necessary dependencies are not ignored.

chapulina:
New build now that https://github.com/ros-perception/image_common/pull/120 is in: [Build Status](https://ci.ros2.org/job/ci_linux/6519/)

chapulina (approved):
This fixes the build, so we should get it in. At some point before the Dashing release we'll need to fix the new linter errors and check what's causing all the new test failures though.

PR #892, opened 2019-02-19
Add document describing how to release packages using a custom Gazebo version

j-rivero:
The new document describe the manual process needed to be able to produce binaries from the `gazebo_ros_pkgs` repository using a different version of Gazebo. The whole process could probably be fully automated or highly improved but I think is good to document it here so other people can use the information.


docs/RELEASE.md
```diff
@@ -0,0 +1,227 @@
...
+ 3. Run bloom on the modified release repository.
+ 4. Custom infrastructure to create .deb packages
+
+### Upstream versions released using this tutorial
+
+The `gbp -release repository` hosts the latest version released by the maintainers
+of `gazebo_ros_pkgs`. When using these instructions to release a new custom version
+the version of `gazebo_ros_pkgs` released will be the latest one existing in the
+official `gbp -release repository`. The version would be the same but the release number
+will be one above since bloom needs to be run on top of the modified `-release repository`.
```
chapulina, line 41: There's a lot of valuable info in this paragraph, just to be sure I understood it all: when using a custom `gbp` repo, the version released will still come from the official repository, but the debs will be built against a different Gazebo version, correct?

So for example, if the latest `gazebo_ros_pkgs` official release for Kinetic (Gazebo7) are of version N.M-1, following these instructions one would be able to release N.M-2, which supports Kinetic and Gazebo9. If that's correct, mind adding a more concrete example like this to the paragraph so it's a bit easier to parse for beginners like me?

chapulina (approved):
Thanks, @j-rivero! I fixed a few typos in 6ad006ef4f4dde926bdb0d7ad1b0d34316604df5 and just have one request for clarification, but this looks good to me.

PR #893, opened 2019-02-21
Remove extra angle bracket

DLu:
Don't know why no one has cared in the past 6 years that this angle bracket makes invalid XML, but here's the fix anyway.


chapulina (approved):

DLu:
What do you want to do about the other branches?

chapulina:
If you make PRs against them I'd be happy to click merge :sweat_smile:

PR #897, opened 2019-02-26
added noise and outlier for the openni kinect plugin for ros melodic

chisarie:
### Intro

Hi, I am a developer working with a depth camera that we simulate using the kinect plugin. When simulating the depth camera, outliers and gaussian noise are useful in order to have a more realistic representation of the sensor.
### Gaussian noise

The implementation is a zero mean multiplicative Gaussian Noise, where the standard deviation can be set by the user. The way this is implemented was taken from the gaussian noise representation in the [velodyne_gazebo_plugins](https://bitbucket.org/DataspeedInc/velodyne_simulator/src/master/velodyne_gazebo_plugins/) package, which is under BSD license. The author does not intend to brake any Copyright; if it happened to be the case with this pull request, please decline it.

### Outliers

Also the outliers are implemented to act on the depth value of each point. In particular, the outlier magnitude is the percentage ( [-1, +1] ) that we want the outlier points to deviate from the nominal ones, while the outlier index is the position we want the outlier to be in (zero for random).

### Usage

When calling the plugin from your .xacro file, as explained in this [tutorial](http://gazebosim.org/tutorials?tut=ros_depth_camera&cat=connect_ros), simply add the parameters

- `<gaussianNoise>0.01</gaussianNoise>` 
- `<outlierMagnitude>0.4</outlierMagnitude>`
- `<outlierIndex>5000</outlierIndex>`
(or whatever other value you prefer).

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_openni_kinect.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_openni_kinect.h
@@ -123,6 +123,24 @@ namespace gazebo
     /// \brief ROS image topic name
     private: std::string point_cloud_topic_name_;
 
+    /// \brief Gaussian noise standard deviation
+    private: double gaussian_noise_;
+
+    /// \brief Gaussian noise generator
+    private: static double gaussianKernel(double mu, double sigma)
+    {
+      // using Box-Muller transform to generate two independent standard normally distributed normal variables
+      // see wikipedia
+      double U = (double)rand() / (double)RAND_MAX; // normalized uniform random variable
+      double V = (double)rand() / (double)RAND_MAX; // normalized uniform random variable
+      return sigma * (sqrt(-2.0 * ::log(U)) * cos(2.0 * M_PI * V)) + mu;
+    }
+
+    /// \brief outlier distance in meters from nominal point
+    private: double outlier_magnitude_;
+
+    /// \brief the index of the point which will become the outlier
+    private: int outlier_index_;
 
     /// \brief image where each pixel contains the depth data
     private: std::string depth_image_topic_name_;
--- a/gazebo_plugins/src/gazebo_ros_openni_kinect.cpp
+++ b/gazebo_plugins/src/gazebo_ros_openni_kinect.cpp
@@ -103,6 +103,34 @@ void GazeboRosOpenniKinect::Load(sensors::SensorPtr _parent, sdf::ElementPtr _sd
   else
     this->point_cloud_cutoff_max_ = _sdf->GetElement("pointCloudCutoffMax")->Get<double>();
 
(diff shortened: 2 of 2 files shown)


YoshuaNava:
@chapulina @j-rivero We have contributed back this change from our fork. Do you think it makes sense, within the development lines of this one? We don't want to drift away from the work done upstream.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_openni_kinect.h
```diff
@@ -123,6 +123,24 @@ namespace gazebo
     /// \brief ROS image topic name
     private: std::string point_cloud_topic_name_;
 
+    /// \brief Gaussian noise standard deviation
+    private: double gaussian_noise_;
+
+    /// \brief Gaussian noise generator
+    private: static double gaussianKernel(double mu, double sigma)
```
chapulina, line 130: Please move the implementation in the `cc` file.

Nitpick: could you move the function declaration up together with (most) other functions? And follow the `CamelCase` style?

Also, it would be great to document all the params (`mu` /`sigma`). (I know, the rest of the file is barely documented :confused: )

gazebo_plugins/src/gazebo_ros_openni_kinect.cpp
```diff
@@ -103,6 +103,34 @@ void GazeboRosOpenniKinect::Load(sensors::SensorPtr _parent, sdf::ElementPtr _sd
...
+  if (!_sdf->HasElement("outlierIndex")) {
+    this->outlier_index_ = 0;
+  } else {
+    this->outlier_index_ = _sdf->GetElement("outlierIndex")->Get<double>();
+  }
+  int max_index = this->width * this->height;
+  if (outlier_index_ == 0) {
+  // generate a uniformly distributed random number    
+  std::uniform_int_distribution<> uniform_distribution(1, max_index);
+  std::mt19937 generator( std::random_device{}() );
```
chapulina, line 130: We've had cross-platform issues with `std::mt19937` [before](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/878#issuecomment-512132819). Could you use `ignition::math::Rand::DblUniform()`?

Also please be sure to indent the code.

chapulina (changes requested):
Thank you for the PR.

The gaussian kernel implementation looks just like other parts of the code in this repository, such as:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/fc2709384ce7cddff7ecedaaed5da808e8477d0d/gazebo_plugins/src/gazebo_ros_imu_sensor.cpp#L120-L132

I should also bring up that this PR will break ABI, but we haven't been enforcing ABI stability for Melodic, so that should be ok.

Issue #898, opened 2019-02-27
Video Plugin Crashes when create multiple video instance in the same world


talregev:
I am using gazebo 9 with ros melodic.
I was trying the video plugin, with one video it working, and with 2 video instance it crashes.

shiveshkhaitan:
This happens only when the ```<robotNamespace>``` of the two instances of the plugins are same.

This is because of this:
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/e42cc26a473733b05abc648181a6de1286105ed6/gazebo_plugins/src/gazebo_ros_video.cpp#L76-L79

Same namespace creates two instances of the above entity with the same name.

 Keeping different ```<robotNamespace>``` for the plugins can be a temporary solution until its fixed.

talregev:
That help!
you can create a random name for `<robotNamespace>` if there non provide.

PR #899, opened 2019-03-05
[ros2] Port video plugin to ros2

shiveshkhaitan:
Port gazebo_ros_video plugin to ROS2. 

Example:
```
<plugin name="display_video_controller" filename="libgazebo_ros_video.so">
  <ros>
    <namespace>rrbot</namespace>
    <argument>~/image_raw:=/image_raw</argument>
  </ros>
  <height>1000</height>
  <width>1000</width>
</plugin>
```


shiveshkhaitan:
#898 will persist in the current implementation. @chapulina is it a good idea to keep a count of the number of video plugins initialized and append it to the ```entity name```? This way two instances of the plugin will initialize the ```Ogre::MovableObject``` with different names. I have tried this and it fixes the problem. Or is there any other fix?

gazebo_plugins/include/gazebo_plugins/gazebo_ros_video.hpp
```diff
@@ -0,0 +1,109 @@
...
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+// ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+/*
+ * Desc: Video plugin for displaying ROS image topics on Ogre textures
+ * Author: Shivesh Khaitan
+ * Date: 04 March 2019
```
chapulina: Since this is porting the code that already existed, we should keep the original date and author's name. Feel free to add a comment with your name mentioning that you've done the ROS2 port if you wish.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_video.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <memory>
+
+namespace gazebo_plugins
+{
+class GazeboRosVideoPrivate;
+
+/// Video plugin for displaying ROS image topics on Ogre textures
+/*
+ *
+ * \author  Shivesh Khaitan (shiveshkhaitan@gmail.com)
```
chapulina: Same here about not changing the author's name

gazebo_plugins/include/gazebo_plugins/gazebo_ros_video.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+      <!-- Update rate in Hz -->
+      <height>240</height>
+      <weight>320</weight>
+
+    </plugin>
+  \endcode
+*/
+
+class VideoVisual : public gazebo::rendering::Visual
```
chapulina: While migrating code to ROS2, we're trying to make sure all classes, functions and variables are documented. Mind adding documentation here? You can see some examples in the plugins already ported.

chapulina: Since the `VideoVisual` class is not used in the public API (note how it isn't referenced anywhere in the header), it can be moved to the `.cpp` file. This will give us more flexibility to change it after released.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_video.hpp
```diff
@@ -0,0 +1,109 @@
...
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+// FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+// COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+// INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+// BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+// ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
```
chapulina: We should keep the same license as the original file, which was Apache 2. This also applies to the `cpp` file.

gazebo_plugins/src/gazebo_ros_video.cpp
```diff
@@ -0,0 +1,222 @@
...
+  gazebo::rendering::VisualPtr model_;
+
+  /// Connection to event called at every world iteration.
+  gazebo::event::ConnectionPtr update_connection_;
+
+  std::shared_ptr<VideoVisual> video_visual_;
+
+  cv_bridge::CvImagePtr image_;
+  std::mutex m_image_;
+  bool new_image_available_;
```
chapulina: Here too make sure to document all member variables

gazebo_plugins/src/gazebo_ros_video.cpp
```diff
@@ -0,0 +1,222 @@
...
+ public:
+  /// Callback when a image is received.
+  /// \param[in] _msg Image command message.
+  void processImage(const sensor_msgs::msg::Image::ConstSharedPtr msg);
+
+  /// Callback to be called at every simulation iteration.
+  void onUpdate();
+
+  /// Pointer to the model.
+  gazebo::rendering::VisualPtr model_;
```
chapulina: It shouldn't be necessary to keep this pointer as a member variable since it is not used anywhere, this line and `impl_->model_ = _parent;` can be removed.

gazebo_plugins/worlds/gazebo_ros_video.world
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" ?>
```
chapulina: Thanks for including the demo world! Mind renaming it to `gazebo_ros_video_demo.world` to follow the pattern of the rest?

Also, it would be more convenient for quick testing if the video would just start playing as the world is opened. I created an example [here](https://bitbucket.org/snippets/chapulina/be5bdk), feel free to copy it here or propose an alternative. It has a camera that publishes images that are streamed to the video visual, it looks like this:

video_ros2

chapulina (changes requested):
Thank you for the contribution!

I realize that creating a test for this could end up being more complicated than the port itself, but maybe we can add a test that just loads the plugin and publishes an image message to make sure the functions are called? We could use a world similar to the demo one I created with a camera. The idea is to at least make sure there's no crash or other strange behaviour.

Another thing to fix are the linter errors, make sure to run `colcon test` and check the failed results under `log/latest_test/gazebo_plugins/stdout.log`. There are currently some `cpplint`, `lint_cmake` and `uncrustify` errors there.

Let me know if you have any questions, cheers!

shiveshkhaitan:
@chapulina thank you for the review. I have made the required changes in the latest commit. Please review the same. 
The test currently only publishes an image and essentially checks if the plugin executes without an error. Please let me know if I can add anything else to be tested.

gazebo_plugins/package.xml
```diff
@@ -22,6 +22,7 @@
   <depend>camera_info_manager</depend>
   <depend>geometry_msgs</depend>
   <depend>image_transport</depend>
+  <depend>cv_bridge</depend>
```
chapulina: This should be in alphabetical order

gazebo_plugins/include/gazebo_plugins/gazebo_ros_video.hpp
```diff
@@ -0,0 +1,78 @@
...
+/*
+ *
+ * \author Piyush Khandelwal (piyushk@gmail.com)
+ *
+ */
+/**
+  Example Usage:
+  \code{.xml}
+    <plugin name="gazebo_ros_video" filename="libgazebo_ros_video.so">
+      * Images are subscribed from: /custom_ns/custom_img
```
chapulina: Nitpick: This bullet point doesn't fit well inside the XML, maybe it could be wrapped in `<!-- -->`
```suggestion
      <!-- Images are subscribed from: /custom_ns/custom_img -->
```

gazebo_plugins/test/test_gazebo_ros_video.cpp
```diff
@@ -0,0 +1,83 @@
+// Copyright 2018 Open Source Robotics Foundation, Inc.
```
chapulina, line 1: Thanks for writing the test! I put some print statements in the plugin and noticed that it wasn't being loaded while running the test though. Then I remembered that when dealing with visual plugins, we should use the `RenderingFixture` instead of `ServerFixture`. Unfortunately, this is not documented :cry: 

So I propose some changes to the test:

* Use `RenderingFixture`
* Instead of stepping the physics, "step" the rendering by running full `pre/render/post` cycles
* Check that all visuals are properly created. When checking this, I noticed that the video visual was not being tracked by the scene, so I added a call to `Scene::AddVisual` in the plugin. This allows us to get it with `Scene::GetVisual`.

Let me know what you think!

```diff
diff --git a/gazebo_plugins/src/gazebo_ros_video.cpp b/gazebo_plugins/src/gazebo_ros_video.cpp
index f8d8a0f..002ae03 100644
--- a/gazebo_plugins/src/gazebo_ros_video.cpp
+++ b/gazebo_plugins/src/gazebo_ros_video.cpp
@@ -200,8 +200,9 @@ void GazeboRosVideo::Load(
 
   int width = _sdf->Get<int>("width", 320).first;
 
-  impl_->video_visual_ =
-    std::make_shared<VideoVisual>("visual", _parent, height, width);
+  impl_->video_visual_ = std::make_shared<VideoVisual>(_parent->Name() + "::video_visual",
+      _parent, height, width);
+  _parent->GetScene()->AddVisual(impl_->video_visual_);
 
   // Subscribe to the image topic
   impl_->camera_subscriber_ =
@@ -214,7 +215,9 @@ void GazeboRosVideo::Load(
   impl_->update_connection_ = gazebo::event::Events::ConnectPreRender(
     std::bind(&GazeboRosVideoPrivate::onUpdate, impl_.get()));
 
-  RCLCPP_INFO(impl_->rosnode_->get_logger(), "GazeboRosVideo has started");
+  RCLCPP_INFO(impl_->rosnode_->get_logger(),
+    "GazeboRosVideo has started. Subscribed to [%s]",
+    impl_->camera_subscriber_->get_topic_name());
 }
 
 void GazeboRosVideoPrivate::onUpdate()
diff --git a/gazebo_plugins/test/test_gazebo_ros_video.cpp b/gazebo_plugins/test/test_gazebo_ros_video.cpp
index 32d41b7..24eed5a 100644
--- a/gazebo_plugins/test/test_gazebo_ros_video.cpp
+++ b/gazebo_plugins/test/test_gazebo_ros_video.cpp
@@ -21,7 +21,7 @@
 
 using namespace std::literals::chrono_literals; // NOLINT
 
-class GazeboRosVideoTest : public gazebo::ServerFixture
+class GazeboRosVideoTest : public gazebo::RenderingFixture
 {
 public:
   void TearDown() override
@@ -36,16 +36,33 @@ TEST_F(GazeboRosVideoTest, VideoSubscribeTest)
   // Load test world and start paused
   this->Load("worlds/gazebo_ros_video.world", true);
 
-  // World
-  auto world = gazebo::physics::get_world();
-  ASSERT_NE(nullptr, world);
+  // Get the scene
+  auto scene = gazebo::rendering::get_scene();
+  ASSERT_TRUE(scene != nullptr);
 
-  // Model box
-  auto box = world->ModelByName("box_display");
-  ASSERT_NE(nullptr, box);
+  // Trigger render events until model is loaded
+  int sleep = 0;
+  int maxSleep = 30;
+  gazebo::rendering::VisualPtr modelVis;
+  gazebo::rendering::VisualPtr linkVis;
+  gazebo::rendering::VisualPtr visualVis;
+  gazebo::rendering::VisualPtr videoVis;
+  for (; !modelVis && sleep < maxSleep; ++sleep) {
+    gazebo::event::Events::preRender();
+    gazebo::event::Events::render();
+    gazebo::event::Events::postRender();
 
-  // Step a bit for model to settle
-  world->Step(100);
+    modelVis = scene->GetVisual("box_display");
+    linkVis = scene->GetVisual("box_display::base_link");
+    visualVis = scene->GetVisual("box_display::base_link::visual");
+    videoVis = scene->GetVisual("box_display::base_link::visual::video_visual");
+    gazebo::common::Time::MSleep(100);
+  }
+  EXPECT_LT(sleep, maxSleep);
+  EXPECT_NE(nullptr, modelVis);
+  EXPECT_NE(nullptr, linkVis);
+  EXPECT_NE(nullptr, visualVis);
+  EXPECT_NE(nullptr, videoVis);
 
   // Create node and executor
   auto node = std::make_shared<rclcpp::Node>("gazebo_ros_video_test");
@@ -67,10 +84,23 @@ TEST_F(GazeboRosVideoTest, VideoSubscribeTest)
     msg.data.push_back(255);
     msg.data.push_back(255);
   }
-  pub->publish(msg);
-  executor.spin_once(100ms);
-  // Wait for it to be processed
-  world->Step(1000);
+
+  // Give iterations for it to be processed. Right now we don't have a quick way to verify the
+  // output, so we just make sure there's no crash
+  sleep = 0;
+  maxSleep = 10;
+  for (; sleep < maxSleep; ++sleep) {
+    pub->publish(msg);
+    executor.spin_once(100ms);
+
+    gazebo::event::Events::preRender();
+    gazebo::event::Events::render();
+    gazebo::event::Events::postRender();
+
+    EXPECT_NE(nullptr, scene->GetVisual("box_display"));
+    EXPECT_NE(nullptr,
+      scene->GetVisual("box_display::base_link::visual::video_visual"));
+  }
 }
 
 int main(int argc, char ** argv)
diff --git a/gazebo_plugins/test/worlds/gazebo_ros_video.world b/gazebo_plugins/test/worlds/gazebo_ros_video.world
index b30bfc1..293aaa2 100644
--- a/gazebo_plugins/test/worlds/gazebo_ros_video.world
+++ b/gazebo_plugins/test/worlds/gazebo_ros_video.world
@@ -1,6 +1,6 @@
 <?xml version="1.0" ?>
 <!--
-  Gazebo ROS video plugin demo
+  Gazebo ROS video plugin test
 
     The visual on top of the box is displaying the images coming from a camera.
 -->
@@ -28,7 +28,7 @@
             <ros>
               <!-- Remap for image topic to be subscribed -->
               <namespace>/test</namespace>
-              <argument>~/image_raw:=/test/video_test</argument>
+              <argument>~/image_raw:=video_test</argument>
             </ros>
             <height>120</height>
             <width>160</width>
@@ -46,4 +46,4 @@
     </model>
 
   </world>
-</sdf>
\ No newline at end of file
+</sdf>
```

chapulina (changes requested):
Thanks for iterating! I'm ready to approve once the test is updated.

When you address these comments, it would be convenient if you added a commit on top of the existing one so I can review just the new changes and don't need to go over the whole code again. We can squash them together when merging the pull request.

chapulina (approved):
Great work!

chapulina:
@shiveshkhaitan , mind adding a migration page to the wiki? It doesn't need to be too elaborate, you can follow the pattern of existing pages, such as [ROS 2 Migration: Diff-drive](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Diff-drive).

shiveshkhaitan:
Migration guide is available at [ROS 2 Migration: Video](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Video)

chapulina:
> is it a good idea to keep a count of the number of video plugins initialized and append it to the entity name

I just realized I never replied to this question. Counting should be ok. We could also use the visual's scoped name, which is guaranteed to be unique.

PR #901, opened 2019-03-07
[ros2] Disable test_node to clean CI

chapulina:
While the root cause of issue #855 is not figured out, let's disable this test so it's easier to spot regressions on CI.


chapulina:
This is just commenting out a test and fixing a small linter error. Merging so that we have a clean CI baseline for other PRs.

PR #902, opened 2019-03-07
Fix bug in clipping of block laser depth values

sloretz:
This fixes a bug where range values near the max range of a depth camera are incorrectly reported as max range. The range value was being clipped to `(-infinity, maxRange-minRange]`, but it should be clipped to `[minRange, maxRange]`. Range readings past `maxRange - minRange` were reported as `maxRange`.

Before this pull request the range visualization in gazebo is correct, but the published point cloud in RViz shows all max range.

gazebo_ros_depth_max_range_issue

After this pull request the point cloud in RViz shows the correct depths.
gazebo_ros_depth_max_range_issue_with_fix

I'd like to add a test, but I haven't looked through the test code in this repo yet so I'm not yet sure where to add it.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_block_laser.cpp
+++ b/gazebo_plugins/src/gazebo_ros_block_laser.cpp
@@ -304,15 +304,19 @@ void GazeboRosBlockLaser::PutLaserData(common::Time &_updateTime)
       j3 = hja + vjb * rayCount;
       j4 = hjb + vjb * rayCount;
       // range readings of 4 corners
-      r1 = std::min(this->parent_ray_sensor_->LaserShape()->GetRange(j1) , maxRange-minRange);
-      r2 = std::min(this->parent_ray_sensor_->LaserShape()->GetRange(j2) , maxRange-minRange);
-      r3 = std::min(this->parent_ray_sensor_->LaserShape()->GetRange(j3) , maxRange-minRange);
-      r4 = std::min(this->parent_ray_sensor_->LaserShape()->GetRange(j4) , maxRange-minRange);
+      r1 = this->parent_ray_sensor_->LaserShape()->GetRange(j1);
+      r2 = this->parent_ray_sensor_->LaserShape()->GetRange(j2);
+      r3 = this->parent_ray_sensor_->LaserShape()->GetRange(j3);
+      r4 = this->parent_ray_sensor_->LaserShape()->GetRange(j4);
 
       // Range is linear interpolation if values are close,
       // and min if they are very different
       r = (1-vb)*((1 - hb) * r1 + hb * r2)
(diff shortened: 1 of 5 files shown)


sloretz:
Added block laser clipping test in ce8f33c

sloretz:
While writing the test I discovered that the problem is a little worse than I first reported. Without this PR all ranges from `gazebo_ros_block_laser` are wrong; they're all minRange too big.

Test world
gazebo_ros_block_laser_clipping

output without this PR
gazebo_ros_block_laser_clipping_pre_fix

output with this PR
gazebo_ros_block_laser_clipping_post_fix

gazebo_plugins/src/gazebo_ros_block_laser.cpp
```diff
@@ -304,16 +304,20 @@ void GazeboRosBlockLaser::PutLaserData(common::Time &_updateTime)
...
+      r4 = this->parent_ray_sensor_->LaserShape()->GetRange(j4);
 
       // Range is linear interpolation if values are close,
       // and min if they are very different
       r = (1-vb)*((1 - hb) * r1 + hb * r2)
          +   vb *((1 - hb) * r3 + hb * r4);
 
+      // clamp r to the min and max range of the sensor
+      r = std::min(r, maxRange);
+      r = std::max(r, minRange);
```
chapulina: > The range value was being clipped to (-infinity, maxRange-minRange], but it should be clipped to [minRange, maxRange]

Agreed that the previous clipping was weird. But accroding to [REP 117](https://www.ros.org/reps/rep-0117.html), shouldn't the out of range values become +-inf?

chapulina (changes requested):
> Without this PR all ranges from gazebo_ros_block_laser are wrong; they're all minRange too big.

While a fix, this change seems to big for a released distro, since downstream users may already be compensating for this difference. How do you feel about retargeting this to `noetic`?

gazebo_plugins/src/gazebo_ros_block_laser.cpp
sloretz: Good catch. I changed to +- inf in ad7edd2 and adjusted the test world so their should be both -inf and inf values, but it turns out it doesn't matter. The gazebo sensor only casts rays from min to max range anyways because that's taken from the `sdf` xml tags, so values outside [minRange, maxRange] are impossible.

I don't think I can make time to come up for a fix for the block laser plugin never reporting values outside of min/max range, but this PR is still an improvement since it fixes the values being `minRange` too big.

sloretz:
> While a fix, this change seems to big for a released distro, since downstream users may already be compensating for this difference. How do you feel about retargeting this to noetic?

@chapulina rebased and retargeted at `noetic-devel`.

chapulina (approved):

PR #903, opened 2019-03-08
add support in 32FC1 encoding msg images

talregev:
add support in 32FC1 encoding msg images
openni kinect depth msg of camera gazebo plugin.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_video.cpp
+++ b/gazebo_plugins/src/gazebo_ros_video.cpp
@@ -186,8 +186,7 @@ namespace gazebo
     }
 
     std::string name = robot_namespace_ + "_visual";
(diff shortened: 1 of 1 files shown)


chapulina (approved):
This looks ok to me as it doesn't affect existing behaviour and compiles. But I didn't test it since I'd need to setup a world just for this. @talregev , could you point us to an example where this is used so people looking at this PR in the future have a reference?

Also, it would be great to port this to ROS 2. And it would be easier to test it there too, since there's already a [demo world](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/dashing/gazebo_plugins/worlds/gazebo_ros_video_demo.world) for the video plugin which could be extended to also include depth images.

Issue #886, opened 2019-02-11
gazebo_ros_control [ERROR] [...]: No p gain specified for pid.  Namespace: /gazebo_ros_control/pid_gains


matteolucchi:
Hello, I am running a simulation of a mobile manipulator, composed by a mobile base with differential drive and a 6 DOF manipulator. 

For the manipulator I am using a `velocity_controllers/JointGroupVelocityController` controller, whereas for the mobile base I am using `diff_drive_controller/DiffDriveController`. 
After the latest update fa9c99042d1f013398d4fa68398ca772283b85b8 , this error started to pop up for all the joints in my robot model 

`[ERROR] [1549875853.096599133, 0.167000000]: No p gain specified for pid.  Namespace: /gazebo_ros_control/pid_gains/<joint_name>`. 

The controllers still work fine, but I don't understand what causes this error. I tried adding pid gains to the yaml configuration file of the controllers:

```
/gazebo_ros_control:
  pid_gains:
      joint_name:
          p: 100.0
          i: 1.0
          d: 0.1
```
for all the joints. Adding this removes the error when launching the simulation but produces weird behavior for the robot. The robot arm is not standing correctly anymore.
I looked around a lot but I couldn't figure out a way to solve this, this appeared only after the last update. Any hint on how to solve this? 

I am running the simulation on Gazebo 7 and Kinetic.

Thank you in advance for your time

yangtfu:
same issue

francofusco:
I had the same question some weeks ago, and I think I understood what is happening. The point is that Gazebo allows you to use position and velocity control interfaces in a flexible way:
- if you do not specify the gains: when setting the position (velocity) command, the function `SetPosition` (`SetParam`) is used (see [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/fa9c99042d1f013398d4fa68398ca772283b85b8/gazebo_ros_control/src/default_robot_hw_sim.cpp#L322) and [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/fa9c99042d1f013398d4fa68398ca772283b85b8/gazebo_ros_control/src/default_robot_hw_sim.cpp#L363)). This should mean that the command is applied exactly, and therefore I think the dynamics will not be taken into account.
[...]

PR #912, opened 2019-03-25
Make pid init quiet

nlamprian:
With position and velocity controllers, when you don't upload pid gains, you get an error msg

```
No p gain specified for pid...
```

That's leaving you with the impression that you have to specify the pid. It wasn't until I looked inside the default robot hw sim that I realized not using a pid is a valid control method.

I changed the pid init to quiet so the error is not printed anymore.


vincentrou:
This will fix https://github.com/ros-simulation/gazebo_ros_pkgs/issues/886
It should also be ported on melodic.

brschettini:
Hello, any expectations on merging this PR?

Thanks!

chapulina (approved):
LGTM

If someone out there was relying on these error messages and finds this PR in the future, my apologies in advance.

Here's the [init](http://docs.ros.org/kinetic/api/control_toolbox/html/classcontrol__toolbox_1_1Pid.html#aa0ee79554008ea31e0c212187171fcbe) function documentation for future reference.

PR #917, opened 2019-04-06
[ros2] Port tricycle_drive plugin to ros2

shiveshkhaitan:
Port gazebo_ros_tricycle_drive plugin to ROS2. 

Example:
```
<plugin name='tricycle_drive' filename='libgazebo_ros_tricycle_drive.so'>
  <ros>
    <namespace>/demo</namespace>
  </ros>

  <!-- wheels -->
  <steering_joint>wheel_front_steer</steering_joint>
  <actuated_wheel_joint>wheel_front_spin</actuated_wheel_joint>
  <encoder_wheel_left_joint>wheel_rear_left_spin</encoder_wheel_left_joint>
  <encoder_wheel_right_joint>wheel_rear_right_spin</encoder_wheel_right_joint>

  <!-- kinematics -->
  <wheel_separation>1.0</wheel_separation>
  <encoder_wheel_diameter>0.3</encoder_wheel_diameter>
  <actuated_wheel_diameter>0.3</actuated_wheel_diameter>

  <!-- limits -->
  <max_wheel_torque>20</max_wheel_torque>
  <max_wheel_acceleration>5.0</max_wheel_acceleration>
  <max_steering_speed>1.0</max_steering_speed>

  <!-- output -->
  <publish_odom>true</publish_odom>
  <publish_wheel_tf>true</publish_wheel_tf>
  <publish_wheel_joint_state>true</publish_wheel_joint_state>

  <robot_base_frame>fork</robot_base_frame>

</plugin>
```


chapulina:
Attempt at running CI: [Build Status](https://ci.ros2.org/job/ci_linux/7137/)

shiveshkhaitan:
Probably the test failed because of the required angular velocity set in the test which is not feasible given the current constraints of the model. (Not sure how the test passed initially)

chapulina:
It could also be something else, note that the diff drive test is also failing on this branch (along with many others)

shiveshkhaitan:
I tried building ros-simulation/gazebo_ros_pkgs branch ros2 and crystal. The tests for diff_drive, force, imu_sensor and joint_state_publisher failed there too. Can it be not because of tricycle_drive and something else?

chapulina:
@shiveshkhaitan , #926 fixed all tests and warnings on the `ros2` branch. Once you rebase this PR I can trigger new CI.

Also note the changes made in that PR because you may need some of them here too.

shiveshkhaitan:
The tests seem to pass successfully on my system. What could be the problem?

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+      RCLCPP_INFO(impl_->ros_node_->get_logger(), "Publishing wheel transforms");
+    }
+  }
+
+  // listen to the update event (broadcast every simulation iteration)
+  impl_->update_connection_ = gazebo::event::Events::ConnectWorldUpdateBegin(
+    std::bind(&GazeboRosTricycleDrivePrivate::OnUpdate, impl_.get(), std::placeholders::_1));
+}
+
+void GazeboRosTricycleDrive::Reset() {}
```
chapulina: Were you planning on doing something here? Right now, when the world is reset (`Ctrl+R`), the vehicle seems to lose its angular velocity and starts to respond weirdly to commands.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+#include <gazebo/physics/Model.hh>
+#include <gazebo/physics/Link.hh>
+#include <gazebo_plugins/gazebo_ros_tricycle_drive.hpp>
+#include <gazebo_ros/conversions/builtin_interfaces.hpp>
+#include <gazebo_ros/conversions/geometry_msgs.hpp>
+#include <gazebo_ros/node.hpp>
+#include <geometry_msgs/msg/twist.hpp>
+#include <geometry_msgs/msg/pose2_d.hpp>
+#include <nav_msgs/msg/odometry.hpp>
+#include <nav_msgs/msg/occupancy_grid.hpp>
```
chapulina: I don't see `occupancy_grid` being used anywhere

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+#include <sensor_msgs/msg/joint_state.hpp>
+#include <ignition/math/Pose3.hh>
+#include <ignition/math/Vector3.hh>
+#include <sdf/sdf.hh>
+
+#include <tf2_ros/transform_broadcaster.h>
+#include <tf2_ros/transform_listener.h>
+
+#include <assert.h>
```
chapulina: Do you need this header?

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+
+    if (publish_wheel_tf_) {
+      PublishWheelsTf(current_time);
+    }
+
+    if (publish_wheel_joint_state_) {
+      PublishWheelJointState(current_time);
+    }
+
+    double target_wheel_roation_speed = cmd_.linear.x / ( actuated_wheel_diameter / 2.0 );
```
chapulina: Should we also lock the mutex here where we use `cmd_`?

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+  cmd_.angular.z = cmd_msg->angular.z;
+}
+
+void GazeboRosTricycleDrivePrivate::UpdateOdometryEncoder(
+  const gazebo::common::Time & _current_time)
+{
+  double vl = joint_wheel_encoder_left_->GetVelocity(0);
+  double vr = joint_wheel_encoder_right_->GetVelocity(0);
+
+  const gazebo::common::Time & current_time = _current_time;
```
chapulina: Is this copy necessary?

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+
+  double seconds_since_last_update = (current_time - last_odom_update_).Double();
+  last_odom_update_ = current_time;
+
+  double b = wheel_separation_;
+
+  // Book: Sigwart 2011 Autonompus Mobile Robots page:337
```
chapulina: oops, the typo was already there, but I think this is `Autonomous`

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+  vt = tf2::Vector3(pose_encoder_.x, pose_encoder_.y, 0);
+
+  odom_.pose.pose.position.x = vt.x();
+  odom_.pose.pose.position.y = vt.y();
+  odom_.pose.pose.position.z = vt.z();
+
+  odom_.pose.pose.orientation.x = qt.x();
+  odom_.pose.pose.orientation.y = qt.y();
+  odom_.pose.pose.orientation.z = qt.z();
+  odom_.pose.pose.orientation.w = qt.w();
```
chapulina: Suggestion: use [tf2_geometry_msgs](http://wiki.ros.org/tf2_geometry_msgs) to do the conversion here. It's a trivial conversion, but it often helps preventing typos.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+  double max_steering_angle_tol_;
+
+  /// Maximum steering speed
+  double max_steering_speed_;
+
+  /// Distance between the wheels, in meters.
+  double wheel_separation_;
+
+  /// Maximum wheel torque, in Nm.
+  double max_wheel_torque_;
```
chapulina, line 171: This is only used in the `Load` function, so it doesn't need to be kept in a member variable

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+  impl_->max_wheel_torque_ = _sdf->Get<double>("max_wheel_torque", 0.15).first;
+  impl_->max_wheel_accel_ = _sdf->Get<double>("max_wheel_acceleration", 0).first;
+  impl_->max_wheel_decel_ = _sdf->Get<double>(
+    "max_wheel_deceleration", impl_->max_wheel_accel_).first;
+  impl_->max_wheel_speed_tol_ = _sdf->Get<double>("max_wheel_speed_tolerance", 0.01).first;
+  impl_->max_steering_speed_ = _sdf->Get<double>("max_steering_speed", 0).first;
+  impl_->max_steering_angle_tol_ = _sdf->Get<double>("max_steering_angle_tolerance", 0.01).first;
+
+
+  // Get joints
```
chapulina, line 261: If the joint asked by the user is not found, these joints will be `nullptr`. We should check every joint and if any is null, print an error and fail to initialize the plugin.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+  impl_->update_connection_ = gazebo::event::Events::ConnectWorldUpdateBegin(
+    std::bind(&GazeboRosTricycleDrivePrivate::OnUpdate, impl_.get(), std::placeholders::_1));
+}
+
+void GazeboRosTricycleDrive::Reset() {}
+
+void GazeboRosTricycleDrivePrivate::PublishWheelJointState(
+  const gazebo::common::Time & _current_time)
+{
+  std::vector<gazebo::physics::JointPtr> joints;
```
chapulina: This vector will be the same at every iteration. How do you feel about populating it once at `Load` and keeping it as a member variable? It can be used both at `PublishWheelJointState` and `PublishWheelsTf`.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+  joints.push_back(joint_wheel_encoder_left_);
+  joints.push_back(joint_wheel_encoder_right_);
+  for (auto & joint : joints) {
+    std::string frame = joint->GetName();
+    std::string parent_frame = joint->GetParent()->GetName();
+
+    ignition::math::Pose3d pose = joint->GetChild()->RelativePose();
+
+    tf2::Quaternion qt(pose.Rot().X(), pose.Rot().Y(), pose.Rot().Z(), pose.Rot().W());
+    tf2::Vector3 vt(pose.Pos().X(), pose.Pos().Y(), pose.Pos().Z());
```
chapulina: I suggest you add a couple of conversion functions for these types to `gazebo_ros/conversions`. Not a requirement, just something nice to have. Especially when dealing with quaternions, it saves people time looking up if the order is XYZW or WXYZ.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+void GazeboRosTricycleDrivePrivate::OnUpdate(const gazebo::common::UpdateInfo & _info)
+{
+  gazebo::common::Time current_time = _info.simTime;
+  if (odom_source_ == ENCODER) {
+    UpdateOdometryEncoder(current_time);
+  }
+
+  double seconds_since_last_update = ( current_time - last_actuator_update_ ).Double();
+
+  if (seconds_since_last_update > update_period_) {
```
chapulina: A nitpick of mine is to always return as early as possible to make the code less indented and more readable. So I'd recommend doing:

```
if (seconds_since_last_update < update_period_)
  return;
```

Also I believe we want to update if `seconds_since_last_update == update_period_`. So if you keep your logic, I think you may need to change to `>=`.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+}
+
+void GazeboRosTricycleDrivePrivate::MotorController(
+  double target_speed, double target_angle, double dt)
+{
+  double applied_speed = target_speed;
+  double applied_angle = target_angle;
+
+  double current_speed = joint_wheel_actuated_->GetVelocity(0);
+  if (max_wheel_accel_ > 0) {
```
chapulina: Should we also check `max_wheel_decel` here and on line 430?

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
```diff
@@ -0,0 +1,619 @@
...
+  }
+  joint_wheel_actuated_->SetParam("vel", 0, applied_speed);
+
+  double current_angle = joint_steering_->Position(0);
+
+  // truncate target angle
+  if (target_angle > +M_PI / 2.0) {
+    target_angle = +M_PI / 2.0;
+  } else {
+    if (target_angle < -M_PI / 2.0) {
```
chapulina: This `else` and `if` can be combined into an `else if`

chapulina (changes requested):
This is looking great! Mind adding a migration guide to the wiki?

I have several minor comments, most of which have been carried from the ROS 1 implementation. So not really this PR's fault, but this is a good chance to address them nonetheless.

All tests pass locally for me. I only see a couple of deprecation warnings coming from the video plugin, but the tricycle is clean. We should make a new PR to clean up the video warnings. Once you address my comments I can trigger new CI.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
shiveshkhaitan: Can you please suggest what should be the appropriate behaviour when gazebo is reset for tricycle drive?

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
shiveshkhaitan: I realised that `qt` and `vt` here were not being used. Nevertheless I will be adding the required conversions in `gazebo_ros`.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
chapulina: I'd expect the vehicle to go back to the initial position and be stopped. So all the plugin should need to do is zero the velocity commands.

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
chapulina: Oh nice catch, feel free to add that in a separate PR then, no need to block this one on this :wink:

gazebo_plugins/src/gazebo_ros_tricycle_drive.cpp
shiveshkhaitan: This is reintroduced in the latest commit as its required in reset function as well

shiveshkhaitan:
Migration guide is available at [ROS 2 Migration: Tricycle drive](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Tricycle-drive)

chapulina (approved):
Looks great! Merging!

PR #920, opened 2019-04-23
Fix a problem twice named with ros namespace.

shun818:
# Occured problem:
I insert gazebo_ros_imu_sensor plugin my robot.
```xml my_robot.urdf
<plugin filename="libgazebo_ros_imu_sensor.so" name="imu_plugin">
  <robotNamespace>robotNamespace</robotNamespace>
  <topicName>imu</topicName>
  ...
</plugin>
```
and launch file is 
```xml
<node pkg="gazebo_ros" type="spawn_model" name="spawn_robot" args="
  -urdf
  ...
  -robot_namespace robotNamespace
</node>
```
then, gazebo was published ROS topic is named `/robotNamespace/robotNamespace/imu`

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_imu_sensor.cpp
+++ b/gazebo_plugins/src/gazebo_ros_imu_sensor.cpp
@@ -153,12 +153,12 @@ bool gazebo::GazeboRosImuSensor::LoadParameters()
   //TOPIC
   if (sdf->HasElement("topicName"))
   {
-    topic_name =  robot_namespace + sdf->Get<std::string>("topicName");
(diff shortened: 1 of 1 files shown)


chapulina (commented):
The fix looks correct to me, but we shouldn't merge this into `melodic` because it would break existing users. Would you mind targeting this PR at `noetic-devel`?

It would also be nice to add a more verbose commit message about this, which will go into the changelog, so users migrating from melodic to noetic can quickly find out how to migrate their code.

Thanks!

shun818:
No problem. However, I am not familiar with the operation of Github. Can you change this PR target?

chapulina (approved):

PR #921, opened 2019-05-06
[ros2] Port F3d and FTSensor plugin to ros2

shiveshkhaitan:
Port gazebo_ros_f3d plugin to ROS2.

Example:

```
<plugin name="gazebo_ros_f3d" filename="libgazebo_ros_f3d.so">
    <ros>
        <!-- Add a namespace -->
        <namespace>/demo</namespace>
        <!-- Remap the default topic -->
        <argument>wrench:=wrench_demo</argument>
    </ros>

    <!-- Link name -->
    <body_name>link</body_name>
    
    <!-- Frame id of published message -->
    <frame_name>demo_world</frame_name>

</plugin>
```


shiveshkhaitan:
Merged similar plugins gazebo_ros_f3d and gazebo_ros_ft_sensor into gazebo_ros_ft_sensor for ROS2.

Example:

```
<plugin name="gazebo_ros_ft_sensor" filename="libgazebo_ros_ft_sensor.so">
    <ros>
        <!-- Add a namespace -->
        <namespace>/demo</namespace>
        <!-- Remap the default topic -->
        <argument>wrench:=wrench_demo</argument>
    </ros>

    <!-- Link name -->
    <body_name>link</body_name>
    
    <!-- Frame id of published message -->
    <frame_name>demo_world</frame_name>
    
    <!-- Update rate in Hz, defaults to 0.0, which means as fast as possible -->
    <update_rate>1</update_rate>

    <!-- Standard deviation of the noise to be added to the reported wrench messages. -->
    <gaussian_noise>0.01</gaussian_noise>
</plugin>
```
If `<body_name>` is specified, it acts as original f3d_plugin. Otherwise if `<joint_name>` is specified, it acts as original ft_sensor

shiveshkhaitan:
Migration guides are available at [ROS 2 Migration: FT sensor](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-FT-sensor) and [ROS 2 Migration: F3D](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-F3D)

gazebo_plugins/src/gazebo_ros_ft_sensor.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+/*
+ * \file gazebo_ros_ft_sensor.cpp
+ *
+ * \Author Sachin Chitta and John Hsu
+ * \Author Francisco Suarez-Ruiz
+ *
+ */
+
+#include <gazebo/physics/physics.hh>
```
chapulina: Nitpick: instead of including all Gazebo physics, you could include exactly the headers you'll need, such as `Joint.hh`, `Link.hh`, etc.

gazebo_plugins/src/gazebo_ros_ft_sensor.cpp
```diff
@@ -0,0 +1,202 @@
...
+  impl_->rosnode_ = gazebo_ros::Node::Get(_sdf);
+
+  if (!_sdf->HasElement("update_rate")) {
+    RCLCPP_DEBUG(impl_->rosnode_->get_logger(),
+      "ft_sensor plugin missing <update_rate>, defaults to 0.0 (as fast as possible)");
+  }
+  impl_->update_rate_ = _sdf->Get<double>("update_rate", 0.0).first;
+
+  if (!_sdf->HasElement("body_name") && !_sdf->HasElement("joint_name")) {
+    RCLCPP_WARN(impl_->rosnode_->get_logger(),
```
chapulina: This should be an error instead of a warning, because the plugin will not work. And we should reset the node too so we don't have an extra node that's not being used.

The same goes for all the other checks that return in this function.

gazebo_plugins/src/gazebo_ros_ft_sensor.cpp
```diff
@@ -0,0 +1,202 @@
...
+      "ft_sensor plugin missing <body_name> and <joint_name>, cannot proceed");
+    return;
+  }
+
+  if (_sdf->HasElement("body_name")) {
+    auto link_name = _sdf->Get<std::string>("body_name", "link").first;
```
chapulina: Since you already know that the element is present, you don't need to use the `Get` call that takes a default value, and could use this one instead: `auto link_name = _sdf->Get<std::string>("body_name");`

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ft_sensor.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+        <!-- Remap the default topic -->
+        <argument>wrench:=wrench_demo</argument>
+
+      </ros>
+
+      <!-- Link name -->
+      <body_name>link</body_name>
+
+      <!-- Frame id of published message -->
```
chapulina: This can only be used in conjunction with `body_name`, right? If that's correct, mind documenting this here?

gazebo_plugins/test/test_gazebo_ros_ft_sensor.cpp
```diff
@@ -0,0 +1,85 @@
+// Copyright 2018 Open Source Robotics Foundation, Inc.
```
chapulina: 2019 :wink:

gazebo_plugins/test/test_gazebo_ros_ft_sensor.cpp
```diff
@@ -0,0 +1,85 @@
...
+  world->Step(1000);
+  unsigned int max_sleep{30u};
+  for (auto sleep = 0u; !latest_msg && sleep < max_sleep; ++sleep) {
+    executor.spin_once(100ms);
+    gazebo::common::Time::MSleep(1000);
+  }
+
+  // Check message
+  ASSERT_NE(nullptr, latest_msg);
+  EXPECT_EQ("test_world", latest_msg->header.frame_id);
```
chapulina, line 77: How about removing the ground plane on the world so the box free-falls, and we can check that we're correctly getting forces due to gravity? Something like this:

```.cpp
  EXPECT_DOUBLE_EQ(0.0, latest_msg.wrench.force.x);
  EXPECT_DOUBLE_EQ(0.0, latest_msg.wrench.force.y);
  EXPECT_DOUBLE_EQ(-9.8, latest_msg.wrench.force.z);
  EXPECT_DOUBLE_EQ(0.0, latest_msg.wrench.torque.x);
  EXPECT_DOUBLE_EQ(0.0, latest_msg.wrench.torque.y);
  EXPECT_DOUBLE_EQ(0.0, latest_msg.wrench.torque.z);
```

chapulina (changes requested):
Thank you for the migration guides, this works for me!

I just have a few minor comments. Also, don't forget to delete `gazebo_ros_ft_sensor` from `.ros1_unported`

chapulina (approved):

PR #923, opened 2019-05-12
gazebo_plugins: export plugin path in package.xml

lopsided98:
`gazebo_plugins` does not have a `gazebo_ros` export in its package.xml, so it is not picked up by `gazebo_ros_paths_plugin`. This PR adds the necessary export tag and adds `gazebo_ros` as a dependency so that pluginlib processes this tag correctly.

### DIFF
--- a/gazebo_plugins/package.xml
+++ b/gazebo_plugins/package.xml
@@ -25,6 +25,7 @@
   -->
   <exec_depend>gazebo_dev</exec_depend>
 
+  <depend>gazebo_ros</depend>
   <depend>gazebo_msgs</depend>
   <depend>geometry_msgs</depend>
   <depend>sensor_msgs</depend>
@@ -52,4 +53,7 @@
 
   <test_depend>rostest</test_depend>
 
+  <export>
+    <gazebo_ros plugin_path="${prefix}/../../lib" gazebo_media_path="${prefix}" />
+  </export>
 </package>


gazebo_plugins/package.xml
```diff
@@ -52,4 +53,7 @@
 
   <test_depend>rostest</test_depend>
 
+  <export>
+    <gazebo_ros plugin_path="${prefix}/../../lib" gazebo_media_path="${prefix}" />
```
chapulina, line 57: This looks a bit fishy to me. What kind of workspace have you tried it on? The [usual pattern](http://gazebosim.org/tutorials/?tut=ros_plugins#Updatepackage.xml) is:

`plugin_path="${prefix}/lib"`

chapulina (commented):
Thank you for the PR. I'm wondering how this hasn't come up before. What plugin were you trying to use when you ran into this?

gazebo_plugins/package.xml
lopsided98: I'm using [Nix](https://github.com/lopsided98/nix-ros-overlay), which tends to expose edge case bugs like this because it installs each package into its own directory.

AFAICT, this patch is correct and the documentation is wrong. I opened a PR to fix the documentation: https://bitbucket.org/osrf/gazebo_tutorials/pull-requests/539

`${prefix}` refers to the directory containing the `package.xml` file. This is implemented in rospkg (https://github.com/ros-infrastructure/rospkg/blob/master/src/rospkg/manifest.py#L359) and rospack (https://github.com/ros/rospack/blob/melodic-devel/src/rospack.cpp#L2186, https://github.com/ros/rospack/blob/melodic-devel/src/rospack.cpp#L1399).

chapulina: Thank you for digging in, I believe you're right. I checked that a few installation paths match your pattern:

On Ubuntu from debs:

```
/opt/ros/melodic/share/gazebo_plugins/package.xml
/opt/ros/melodic/lib/libgazebo_ros_imu.so
```

On a colcon workspace with `--merge-install`:

```
install/share/gazebo_plugins/package.xml
install/lib/libgazebo_ros_imu.so
```

On a colcon workspace without `--merge-install`:

```
install/gazebo_plugins/lib/libgazebo_ros_imu.so
install/gazebo_plugins/share/gazebo_plugins/package.xml
```

> I'm wondering how this hasn't come up before.

I believe the plugins are picked up because the `lib` path is included in `LD_LIBRARY_PATH` when sourcing `setup.sh`. But you're right that without this PR they're not picked up by `ros::package`.

---

(FYI, a couple of `master` on your links above should be changed to `melodic-devel`)

chapulina (approved):

jonbinney:
Is this change actually needed? That exported path will only be correct in install space anyway, right? In devel space `${prefix}` will (i think) resolve to the package's location in source space, and the path won't lead to the devel/lib folder.

I almost added something similar to a package because it suddenly stopped finding plugins, but the cause turned out to be that cmake 3.0.2 didn't prepend `lib` to library filenames (cmake 2.8.6 did).

chapulina:
@jonbinney , you may be interested in this conversation: https://bitbucket.org/osrf/gazebo_tutorials/pull-requests/552/

I recently started recommending that people use env-hooks instead of exporting paths through `package.xml`.

jonbinney:
@chapulina env hooks do sound better than exporting from package.xml. But in my experience you don't need either for the plugin path (at least in devel space). I think this is because the setup.sh that catkin generates adds devel/lib to your LD_LIBRARY_PATH. I think this is true for install space as well, since `/opt/ros/melodic/setup.sh` adds /opt/ros/melodic/lib to LD_LIBRARY_PATH.

```
$ . /opt/ros/melodic/setup.sh 

$ env|grep LD_LIBRARY_PATH     
LD_LIBRARY_PATH=/opt/ros/melodic/lib

$ ls /opt/ros/melodic/lib|grep "libgazebo.*\.so"
libgazebo_ros_api_plugin.so
libgazebo_ros_block_laser.so
libgazebo_ros_bumper.so
libgazebo_ros_camera.so
libgazebo_ros_camera_utils.so
libgazebo_ros_control.so
libgazebo_ros_depth_camera.so
libgazebo_ros_diff_drive.so
libgazebo_ros_elevator.so
```

And in gazebo it falls back to just calling dlOpen on the library name if it doesn't find anything by that name in the plugin path: https://bitbucket.org/osrf/gazebo/src/84ae5943bf02843b91a10f82ba084db0d753aee7/gazebo/common/Plugin.hh#lines-171

So i don't think this PR needed to add `plugin_path="${prefix}/../../lib"` in the package.xml (or do anything to env hooks). gazebo_media_path is a different story I assume.

chapulina:
Yeah that makes sense, Gazebo will look into `LD_LIBRARY_PATH` too.

> the setup.sh that catkin generates adds devel/lib to your LD_LIBRARY_PATH

Do you know if that's also true for `catkin_tools` and `colcon`? I'm pretty sure we need either the env-hooks or the exported paths for ROS 2. I haven't used ROS 1 much lately though.

jonbinney:
I just did a test with catkin tools, and it does add `devel/lib` to LD_LIBRARY_PATH.  I tried out colcon both with and without `--symlink-install`, and it adds `install/<package_name>/lib` to LD_LIBRARY_PATH for every package in the workspace.

I mean, they have to, right? Otherwise ROS nodes wouldn't be able to find the libraries they want to link against. Unless they used rpath or something to get around LD_LIBRARY_PATH completely.

jonbinney:
Oh! I just realized who @chapulina is :-P

chapulina:
>  I just realized who @chapulina is :-P

Hi there 🙋

Thanks for the quick tests. I'd have to investigate a bit more to find out why the env hooks / export paths were necessary in the first place. I'm 90% sure Gazebo won't find plugins without them.

PR #925, opened 2019-05-21
[ros2] Port gazebo_ros_path plugin to ROS2

shiveshkhaitan:
Port gazebo_ros_paths_plugin plugin to ROS2.

To export a model, plugin or media file, a package needs to export them from package.xml e.g.
```
<export>
    <gazebo_ros gazebo_model_path="${prefix}/models"/>
    <gazebo_ros gazebo_media_path="${prefix}/models"/>
</export>
```
The directory also needs to be installed from CMakeLists.txt e.g.
```
install(DIRECTORY models
          DESTINATION share/${PROJECT_NAME})
```

The plugin is directly included in ```gazebo.launch.py``` launch file.


gazebo_ros/launch/gazebo.launch.py
```diff
@@ -0,0 +1,85 @@
...
+import launch
+
+
+def add_gazebo_ros_paths():
+    gazebo_model_path = []
+    gazebo_plugin_path = []
+    gazebo_media_path = []
+
+    for package_name, package_path in get_packages_with_prefixes().items():
+        package_share_path = os.path.join(package_path, 'share', package_name)
```
sloretz: Recommend replacing with

```python3
from ament_index_python.packages import get_package_share_directory
# ...
package_share_path = get_package_share_directory(package_name)
```

gazebo_ros/launch/gazebo.launch.py
```diff
@@ -0,0 +1,85 @@
...
+        package_file_path = os.path.join(package_share_path, PACKAGE_MANIFEST_FILENAME)
+        if os.path.isfile(package_file_path):
+            # only try to import catkin if a PACKAGE_FILE is found
```
sloretz: Does `catkin` mean `catkin_pkg`? If so, I don't understand the comment since it is imported above. Maybe this comment could be removed if the import is fine as is?

gazebo_ros/launch/gazebo.launch.py
```diff
@@ -0,0 +1,85 @@
...
+                        plugin_xml_path = plugin_xml_path.replace('${prefix}', package_share_path)
+                        gazebo_plugin_path.append(plugin_xml_path)
+                    if 'gazebo_media_path' in export.attributes:
+                        plugin_xml_path = export.attributes['gazebo_media_path']
+                        plugin_xml_path = plugin_xml_path.replace('${prefix}', package_share_path)
+                        gazebo_media_path.appedn(plugin_xml_path)
+
+    os.environ['GAZEBO_MODEL_PATH'] = ':'.join(gazebo_model_path)
+    os.environ['GAZEBO_PLUGIN_PATH'] = ':'.join(gazebo_plugin_path)
+    os.environ['GAZEBO_RESOURCE_PATH'] = ':'.join(gazebo_media_path)
```
sloretz: I think this should return environment variables to set rather than setting `os.environ` directly. These would then be passed [as `env` in `ExecuteProcess`](https://github.com/ros2/launch/blob/0df8154815e9b355b34e4660945b2110da2332ec/launch/launch/actions/execute_process.py#L86). That limits the scope of the environment variables.

gazebo_ros/launch/gazebo.launch.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
```
sloretz: I think the launch file naming convention is `_launch.py` instead of `.launch.py`. The former plays nicer with the python import system (see examples in `launch_ros` https://github.com/ros2/launch_ros/tree/master/launch_ros/examples )

gazebo_ros/launch/gazebo.launch.py
```diff
@@ -0,0 +1,85 @@
...
+    os.environ['GAZEBO_MODEL_PATH'] = ':'.join(gazebo_model_path)
+    os.environ['GAZEBO_PLUGIN_PATH'] = ':'.join(gazebo_plugin_path)
+    os.environ['GAZEBO_RESOURCE_PATH'] = ':'.join(gazebo_media_path)
+
+
+def generate_launch_description():
+    add_gazebo_ros_paths()
+
+    gazebo = launch.actions.ExecuteProcess(
+        cmd=['gazebo', '--verbose', '-s', 'libgazebo_ros_init.so'],
```
sloretz: Strongly recommend adding a launch argument for `world`.

```python3
    world_argument = launch.actions.DeclareLaunchArgument(
        'world',
        default_value='worlds/empty.world',
        description='Path to world file')
    world_substitution = launch.substitutions.LaunchConfiguration('world')

    gazebo = launch.actions.ExecuteProcess(
        cmd=['gazebo', '--verbose', '-s', 'libgazebo_ros_init.so', world_substitution],
        output='screen'
    )

    return launch.LaunchDescription([
        world_argument,
        gazebo
    ])
```

gazebo_ros/launch/gazebo.launch.py
```diff
@@ -0,0 +1,85 @@
...
+                        plugin_xml_path = plugin_xml_path.replace('${prefix}', package_share_path)
+                        gazebo_model_path.append(plugin_xml_path)
+                    if 'plugin_path' in export.attributes:
+                        plugin_xml_path = export.attributes['plugin_path']
+                        plugin_xml_path = plugin_xml_path.replace('${prefix}', package_share_path)
+                        gazebo_plugin_path.append(plugin_xml_path)
+                    if 'gazebo_media_path' in export.attributes:
+                        plugin_xml_path = export.attributes['gazebo_media_path']
+                        plugin_xml_path = plugin_xml_path.replace('${prefix}', package_share_path)
+                        gazebo_media_path.appedn(plugin_xml_path)
```
sloretz: `appedn` -> `append`

sloretz (changes requested):
With a few changes this mostly works for me. I tested with model and world/media path. I did not try adding a plugin path.

gazebo_ros/launch/gazebo.launch.py
sloretz: Also, this should still include the paths set up by `/usr/share/gazebo/setup.sh`. I'm not sure how to get those though

shiveshkhaitan: For `/usr/share/gazebo/setup.sh`, I have done a os.system() call and it seems to work. Is it fine or should I try something else like parsing `/usr/share/gazebo/setup.sh` and setting the env_variable individually?

chapulina: We shouldn't hardcode `/usr/share/gazebo/setup.sh` because that changes according to the user's installation. I think it's fine to leave that out and assume users will source it before launching if they need it. We should work only with the current value of the environment variables.

gazebo_ros/launch/gazebo_launch.py
```diff
@@ -0,0 +1,99 @@
...
+    os.system('. /usr/share/gazebo/setup.sh')
+
+    model, plugin, media = gazebo_ros_paths()
+    model += ':'+os.environ['GAZEBO_MODEL_PATH']
+    plugin += ':'+os.environ['GAZEBO_PLUGIN_PATH']
+    media += ':'+os.environ['GAZEBO_RESOURCE_PATH']
+
+    world_argument = launch.actions.DeclareLaunchArgument(
+        'world',
+        default_value='worlds/empty.world')
```
chapulina: The default value could be an empty string. This way we're sure we're using Gazebo's default value, in case that ever changes.

chapulina:
@shiveshkhaitan , mind rebasing this and adding a migration guide? Thanks!

shiveshkhaitan:
Porting `gazebo_ros` launch scripts might require some changes in this. Can we wait for the scripts before merging this?

chapulina:
> Can we wait for the scripts before merging this?

SGTM

shiveshkhaitan:
I have removed the gazebo launch part as `gazebo_ros_paths` was originally supposed to be a plugin. The plugin is now available as a python script which can be used by other launch files `gzclient`, `gzserver`

chapulina:
@shiveshkhaitan , can you rebase against the latest `ros2` branch? (And we shouldn't forget to backport to `dashing`). This branch doesn't currently build:

```
/home/developer/ros2_ws/src/ros-simulation/gazebo_ros_pkgs/gazebo_ros/src/node.cpp: In static member function 'static gazebo_ros::Node::SharedPtr gazebo_ros::Node::Get(sdf::ElementPtr)':
/home/developer/ros2_ws/src/ros-simulation/gazebo_ros_pkgs/gazebo_ros/src/node.cpp:82:16: error: 'class rclcpp::NodeOptions' has no member named 'initial_parameters'
   node_options.initial_parameters(initial_parameters);
                ^~~~~~~~~~~~~~~~~~
```

chapulina (approved):
LGTM. 

@shiveshkhaitan , we just need a migration guide and a backport to dashing.

PR #927, opened 2019-05-23
[ros2] Port skid_steer_drive to ROS2

shiveshkhaitan:
Differential drive plugin has been modified to integrate skid steer drive into it. Any number of wheel pairs (1: diff drive, 2: skid_steer_drive or greater than 2) can be specified in the sdf and accordingly the joint names, wheel separations, and wheel diameters have to be specified. 
Example:
```
      <plugin name='diff_drive' filename='libgazebo_ros_diff_drive.so'>

        <ros>
          <namespace>/demo</namespace>
        </ros>

        <!-- Number of wheel pairs -->
        <num_wheel_pairs>2</num_wheel_pairs>
        
        <!-- wheels0 -->
        <left_joint0>left_wheel_joint0</left_joint0>
        <right_joint0>right_wheel_joint0</right_joint0>
        
        <!-- wheels1-->
        <left_joint1>left_wheel_joint1</left_joint1>
        <right_joint1>right_wheel_joint1</right_joint1>

        <!-- kinematics -->
        <wheel_separation0>1.25</wheel_separation0>
        <wheel_separation1>1.25</wheel_separation1>

        <wheel_diameter0>1.0</wheel_diameter0>
        <wheel_diameter1>0.6</wheel_diameter1>

        <!-- limits -->
        <max_wheel_torque>20</max_wheel_torque>
        <max_wheel_acceleration>1.0</max_wheel_acceleration>

        <!-- output -->
        <publish_odom>true</publish_odom>
        <publish_odom_tf>true</publish_odom_tf>
        <publish_wheel_tf>true</publish_wheel_tf>

        <robot_base_frame>chassis</robot_base_frame>

      </plugin>
```
Here there are 2 wheel pairs specified. Thus it will behave as skid steer drive. Similarly any number can be specified for num_wheel_pairs. The joint names, wheel separations, and wheel diameter for each pair has to be specified with their index number appended to the base tag. 

For example `<wheel_separation3>` specifies the spacing for `<left_joint3>` and `<right_joint3>` and so. The indexing starts from 0 to (num_wheel_pairs - 1).


gazebo_plugins/src/gazebo_ros_diff_drive.cpp
```diff
@@ -230,41 +233,60 @@ void GazeboRosDiffDrive::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
   // Initialize ROS node
   impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
 
-  // Get joints
-  impl_->joints_.resize(2);
-
-  auto left_joint = _sdf->Get<std::string>("left_joint", "left_joint").first;
```
chapulina, line 236: This change breaks backwards compatibility with previous versions of the plugin, so all existing models that use `left_joint` wouldn't work anymore.

How about instead of using `num_wheel_pairs` + `left_jointN`, we use multiple `left_joint`. The code could iterate over them doing something like this:

```.cpp
  for (auto left_joint_elem = sdf->GetElement("left_joint"); 
        left_joint_elem != nullptr; 
        left_joint_elem = left_joint_elem->GetNextElement("left_joint") ) {
    // Store value
  }
```

And then sanity check that there's the same number of left and right joints.

The same goes for `wheel_separation`, etc.

gazebo_plugins/worlds/gazebo_ros_diff_drive_demo.world
```diff
@@ -14,11 +14,15 @@
...
 
-    ros2 run tf2_ros tf2_echo chassis right_wheel
+    ros2 run tf2_ros tf2_echo chassis right_wheel0
 
-    ros2 run tf2_ros tf2_echo chassis left_wheel
+    ros2 run tf2_ros tf2_echo chassis left_wheel0
+
+    ros2 run tf2_ros tf2_echo chassis right_wheel1
+
+    ros2 run tf2_ros tf2_echo chassis left_wheel1
```
chapulina: How about leaving the diff drive demo as it is, and adding a separate demo for skid steer? Most people will use just a diff drive setup with 2 wheels, so it would be nice if the demo were as simple as possible. Also, people looking for skid steer would find it more easily if it were a separate demo.

chapulina (changes requested):
Works for me, and I love the demo vehicle!

skid

Mind adding a note to the wiki about migrating the skid steer plugin?

Also, don't forget to remove the skid steer files from `.ros1_unported`.

shiveshkhaitan:
Migration guide is available at [ROS 2 Migration: Skid Steer drive](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Skid-Steer-drive)

chapulina (approved):
Works great for me, thanks for iterating :smile:

LihanChen2004:
May I ask if the modified controller supports translational movement similar to the 4WD Mecanum?

I used the description file of [nexus_4wd_mecanum_simulator](https://github.com/RBinsonB/nexus_4wd_mecanum_simulator/tree/master) and replace the original control plugin with the libgazebo_ros_diff_drive.so plugin in ROS2 Humble. However, it seems to have some directional issues.

When I send a clockwise rotation command, it actually moves forward.
Moreover, there is no movement at all when I issue left or right translation commands.

```
<gazebo>
<plugin name='skid_steer_drive' filename='libgazebo_ros_diff_drive.so'>
    <ros>
      <namespace>/</namespace>
      <remapping>cmd_vel:=cmd_vel</remapping>
      <remapping>odom:=odom</remapping>
    </ros>

    <num_wheel_pairs>2</num_wheel_pairs>

    <left_joint>lower_left_wheel_joint</left_joint>
    <right_joint>lower_right_wheel_joint</right_joint>

    <left_joint>upper_left_wheel_joint</left_joint>
    <right_joint>upper_right_wheel_joint</right_joint>

    <wheel_separation>0.6</wheel_separation>
    <wheel_separation>0.6</wheel_separation>

    <wheel_diameter>0.4</wheel_diameter>
    <wheel_diameter>0.4</wheel_diameter>

    <max_wheel_torque>3</max_wheel_torque>
    <max_wheel_acceleration>6.0</max_wheel_acceleration>

    <publish_odom>false</publish_odom>
    <publish_odom_tf>false</publish_odom_tf>
    <publish_wheel_tf>true</publish_wheel_tf>
    <odometry_frame>odom</odometry_frame>
    <robot_base_frame>base</robot_base_frame>

</plugin>
</gazebo>
```

4WD Mecanum

PR #928, opened 2019-05-29
Accept 0 and 1 as booleans to support URDF to SDF boolean conversion;…

ampleh:
… establish function parity with melodic


chapulina (approved):
Thanks for the PR, this is basically backporting #762 to Kinetic.

In the long run, it would be ideal to use `_value = sdf_->Get<bool>(_tag_name);` instead of supporting all this custom logic, because that also supports other corner cases, like `TRUE` or `True`. We're already doing this on the ROS2 branches.

PR #929, opened 2019-05-29
qos dashing api for video plugin

shiveshkhaitan:
Changes for new QoS api in `gazebo_ros_video` plugin


chapulina (approved):
LGTM, I also moved the video test so it isn't run on systems which don't have displays, such as CI: 316ec645e88ed565eb1dcfb97d84c46b87d4d307.

Will wait for the CI results before merging.

PR #932, opened 2019-06-02
[ros2] Port depth camera to ROS2

shiveshkhaitan:
Port gazebo_ros_depth_camera plugin to ROS2. 

Example:
```
<plugin name="camera_controller" filename="libgazebo_ros_depth_camera.so">
  <ros>
    <namespace>demo_cam</namespace>
     <!-- topics need to be prefixed with camera_name for remapping -->
    <argument>mycamera/image_raw:=mycamera/image_demo</argument>
    <argument>mycamera/image_depth:=mycamera/depth_demo</argument>
    <argument>mycamera/camera_info_raw:=mycamera/raw_cam_info_demo</argument>
    <argument>mycamera/camera_info_depth:=mycamera/depth_cam_info_demo</argument>
    <argument>mycamera/points:=mycamera/points_demo</argument>
  </ros>

  <!-- omit camera_name to default to sensor name -->
  <camera_name>mycamera</camera_name>

  <!-- frame_name>omit so it defaults to link name</frameName-->

  <min_depth>0.5</min_depth>
  <max_depth>500</max_depth>
</plugin>
```


chapulina:
Thanks, @shiveshkhaitan ! Did you take a look at https://github.com/ros-simulation/gazebo_ros_pkgs/pull/757? We could include support for `openni_kinect` and also delete the old files for that plugin. A migration guide for both plugins would also be good.

shiveshkhaitan:
Forgot to mention about #757. Openni Kinect has been already integrated into `gazebo_ros_depth_camera`.

Migration guide is available at [ROS 2 Migration: Camera](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Camera)

chapulina:
All the 3 messages being published are working for me.

I haven't looked too deeply into the code yet, but I did notice there's a lot of overlap with `gazebo_ros_camera`. It would be interesting to look into whether we can avoid duplication. Maybe the depth camera can inherit from the normal camera?

I did notice that the camera plugin doesn't have many virtual functions, so I'm not sure how much can be reused without breaking ABI. We could have separate plugins for Dashing, but maybe it would be interesting to consolidate them for Eloquent. What do you think?

shiveshkhaitan:
I tried integrating depth camera into normal camera. Now `<sensor_type>` has to be specified as `depth` in the sdf to make `gazebo_ros_camera` behave as `gazebo_ros_depth_camera` without the need of an extra plugin. `<sensor_type>` is set to `camera` by default, thus there are no breaking changes.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
```diff
@@ -84,15 +87,47 @@ class GazeboRosCamera : public gazebo::CameraPlugin
...
+  * \param[in] _format Image format
+  */
+  void OnNewDepthFrame(
+    const float * _image,
+    unsigned int _width, unsigned int _height,
+    unsigned int _depth, const std::string & _format) override;
+
   /// Callback when camera is triggered.
-  void OnTrigger(const std_msgs::msg::Empty::SharedPtr _dummy);
+  void OnTrigger(std_msgs::msg::Empty::SharedPtr _dummy);
```
chapulina: Can we leave this as `const`? I don't think it needs to be mutable.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
```diff
@@ -55,7 +58,7 @@ class GazeboRosCameraPrivate;
     </plugin>
   \endcode
 */
-class GazeboRosCamera : public gazebo::CameraPlugin
+class GazeboRosCamera : public gazebo::CameraPlugin, gazebo::DepthCameraPlugin
```
chapulina, line 61: Adding a new base class [breaks ABI](https://community.kde.org/Policies/Binary_Compatibility_Examples#Change_the_class_hierarchy), so we can't get this into Dashing.

I suggest we do this:
1. Create a `dashing` branch, and leave `ros2` for Eloquent
1. Open a new PR against `dashing` with an implementation that doesn't break ABI (similar to what you had before)
1. Merge this PR into `ros2`, deprecating the 
1. Add migration notes from to the wiki

What do you think?

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -90,7 +138,14 @@ GazeboRosCamera::~GazeboRosCamera()
 
 void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _sdf)
 {
-  gazebo::CameraPlugin::Load(_sensor, _sdf);
+  impl_->sensor_type_ = static_cast<GazeboRosCameraPrivate::SensorType>(
+    _sdf->Get<int>("sensor_type", GazeboRosCameraPrivate::CAMERA).first);
```
chapulina: We could even skip exposing the `sensor_type` element and instead check the type with code. This way we can prevent user errors in case someone writes the wrong sensor type.

Something like this:

```.diff
diff --git a/gazebo_plugins/src/gazebo_ros_camera.cpp b/gazebo_plugins/src/gazebo_ros_camera.cpp
index 2473036..7a166cf 100644
--- a/gazebo_plugins/src/gazebo_ros_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera.cpp
@@ -138,6 +138,22 @@ GazeboRosCamera::~GazeboRosCamera()
 
 void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _sdf)
 {
+  // Initialize ROS node
+
+  if (nullptr != std::dynamic_pointer_cast<gazebo::sensors::DepthCameraSensor>(_sensor)) {
+    impl_->sensor_type_ = GazeboRosCameraPrivate::DEPTH;
+  }
+  else if (nullptr != std::dynamic_pointer_cast<gazebo::sensors::CameraSensor>(_sensor)) {
+    impl_->sensor_type_ = GazeboRosCameraPrivate::CAMERA;
+  }
+  else {
+    RCLCPP_ERROR(impl_->ros_node_->get_logger(),
+        "Plugin must be attached to sensor of type `camera` or `depth`\n");
+    impl_->ros_node_.reset();
+    return;
+  }
+
   impl_->sensor_type_ = static_cast<GazeboRosCameraPrivate::SensorType>(
     _sdf->Get<int>("sensor_type", GazeboRosCameraPrivate::CAMERA).first);
 
@@ -153,9 +169,6 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
   // Get tf frame for output
   impl_->frame_name_ = gazebo_ros::SensorFrameID(*_sensor, *_sdf);
 
-  // Initialize ROS node
-  impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
-
   // Image publisher
   // TODO(louise) Migrate image_connect logic once SubscriberStatusCallback is ported to ROS2
   impl_->image_pub_ = image_transport::create_publisher(impl_->ros_node_.get(),
```

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -119,6 +174,30 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
   RCLCPP_INFO(impl_->ros_node_->get_logger(), "Publishing camera info to [%s]",
     impl_->camera_info_pub_->get_topic_name());
 
+  if (impl_->sensor_type_ == GazeboRosCameraPrivate::DEPTH) {
+    // Depth image publisher
+    impl_->depth_image_pub_ = image_transport::create_publisher(impl_->ros_node_.get(),
+        impl_->camera_name_ + "/image_depth");
```
chapulina: We should use `depth/image_raw` and `depth/camera_info` like on the old plugins to keep the same API. Also note that tools like RViz expect* `camera_info` and `image_raw` to have the same topic up to the last part, so that pattern is important.

\* Actually RViz2 is behaving differently from RViz1, see https://github.com/ros2/rviz/issues/207 and https://github.com/ros2/rviz/pull/419.

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -291,6 +381,33 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
   impl_->camera_info_manager_ = std::make_shared<camera_info_manager::CameraInfoManager>(
     impl_->ros_node_.get(), impl_->camera_name_);
   impl_->camera_info_manager_->setCameraInfo(camera_info_msg);
+
+  if (impl_->sensor_type_ == GazeboRosCameraPrivate::DEPTH) {
+    impl_->min_depth_ = _sdf->Get<double>("min_depth", 0.4).first;
+    impl_->max_depth_ =
+      _sdf->Get<double>("max_depth", std::numeric_limits<float>::infinity()).first;
+
+    // Initialize point cloud message
```
chapulina, line 395: I'm curious about why you didn't use `sensor_msgs::PointCloud2Modifier`  like on ROS 1 here. That should be less error-prone.

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -303,39 +420,181 @@ void GazeboRosCamera::OnNewFrame(
...
 
   // Copy from src to image_msg
-  sensor_msgs::fillImage(image_msg, impl_->type_, _height, _width,
-    impl_->skip_ * _width, reinterpret_cast<const void *>(_image));
+  for (uint32_t j = 0; j < _height; j++) {
+    for (uint32_t i = 0; i < _width; i++) {
+      index = i + j * _width;
+      float depth = _image[index];
+      std::memset(&image_msg.data[index * FLOAT_SIZE], depth >= impl_->min_depth_ &&
+        depth <= impl_->max_depth_ ? depth : std::numeric_limits<float>::quiet_NaN(), FLOAT_SIZE);
```
chapulina: I recently learned that gazebo-ros1-pkgs has not been following [REP 117](http://www.ros.org/reps/rep-0117.html). Here we should publish +-inf instead of nan for points out of range. Do you mind addressing this for the depth image and the point cloud?

chapulina (changes requested):

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
shiveshkhaitan: Yes, if thats the case then we will have to have it separate for dashing. I will make the PR once the branch is created

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
shiveshkhaitan: I got a suggestion that 'const' declaration has no affect when used in function prototype, it only affects in the definition. Nevertheless I will revert it back

gazebo_plugins/src/gazebo_ros_camera.cpp
shiveshkhaitan: I thought it has not been ported to ROS2. I will make the changes in the next commit

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
chapulina: Yeah it can be different in the two places. Having it in the prototype is good for documentation, so anyone looking at the header knows that the pointer is not going to be changed. For the `const bool` below it's really not doing anything, because the argument is passed by copy anyway.

In any case, I think removing `const` [breaks ABI](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B#The_Do.27s_and_Don.27ts), so we have to be careful about that.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
chapulina: I created a `dashing` branch. We can proceed to merge this one to `ros2` and backport to `dashing` later.

From now on, let's keep targeting all the other PRs to `dashing` and only target `ros2` if it's a breaking change.

gazebo_plugins/src/gazebo_ros_camera.cpp
shiveshkhaitan: For rviz2, we will have to name it `depth/image_raw` and `depth/image_raw/camera_info` because rviz2 is expecting `camera_info` to be appended to the image_topic

shiveshkhaitan: Also normal rgb camera should have `image_raw` and `image_raw/camera_info` then

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
shiveshkhaitan: Should I create a separate PR for dashing?

chapulina: Yeah we should do that, but let's get this one in first so we discuss the code that is common between them only once.

chapulina: https://github.com/ros/rosdistro/pull/21722

chapulina:
@shiveshkhaitan , can you rebase this branch? Thanks!

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -90,34 +139,88 @@ GazeboRosCamera::~GazeboRosCamera()
...
-  impl_->camera_info_pub_ = impl_->ros_node_->create_publisher<sensor_msgs::msg::CameraInfo>(
-    impl_->camera_name_ + "/camera_info", rclcpp::SensorDataQoS());
+    RCLCPP_INFO(impl_->ros_node_->get_logger(), "Publishing camera info to [%s]",
+      impl_->camera_info_pub_->get_topic_name());
+
+  } else {
+    // Image publisher
+    // TODO(shivesh) Migrate image_connect logic once SubscriberStatusCallback is ported to ROS2
+    impl_->image_pub_ = image_transport::create_publisher(impl_->ros_node_.get(),
+        impl_->camera_name_ + "/ir/image_raw");
```
chapulina: On the Kinect sensor, images prefixed with `ir` correspond to the Infrared sensor readings, which [look different](http://wiki.ros.org/openni_launch/Tutorials/IntrinsicCalibration#IR_.28depth.29_camera) from the raw camera image.

I saw that the `gazebo_ros_openni_kinect` plugin had topics with `ir`, but I think they were never used. So I think we can just skip publishing the raw images for depth cameras.

gazebo_plugins/src/gazebo_ros_camera.cpp
chapulina: Let me know if you're still working on this

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -352,5 +653,9 @@ void GazeboRosCamera::OnTrigger(const std_msgs::msg::Empty::SharedPtr)
   impl_->triggered++;
 }
 
-GZ_REGISTER_SENSOR_PLUGIN(GazeboRosCamera)
+extern "C" GZ_PLUGIN_VISIBLE gazebo::SensorPlugin * RegisterPlugin();
+gazebo::SensorPlugin * RegisterPlugin()
+{
+  return (gazebo::CameraPlugin *)(new GazeboRosCamera());
+}
```
chapulina, line 641: Why did you need to change this? Because of the multiple inheritance?

chapulina (changes requested):
It's working for me, I just have some final comments

gazebo_plugins/src/gazebo_ros_camera.cpp
shiveshkhaitan: Yes. Without the change, I was facing the diamond inheritance problem

gazebo_plugins/src/gazebo_ros_camera.cpp
shiveshkhaitan: Raw rgb images from depth cameras did have use cases in ROS1. I don't think we should remove it. Also the pointcloud's rgb colour is assigned from these images. So isn't it almost similar to RGB images?

chapulina: > Raw rgb images from depth cameras did have use cases in ROS1

Wait, were depth cameras publishing RGB images too?

> So isn't it almost similar to RGB images?

Yup, I agree it's almost the same. And the image is available, so we might as well publish it. I just think we don't *need* to, because afaik ROS 1 didn't publish them. But it doesn't hurt to do so. We just need a better topic name than `ir`. Maybe it could publish in the same topic as an RGB camera would (i.e. `<camera name>/image_raw`). What do you think?

shiveshkhaitan: Depth camera did publish RGB images https://github.com/ros-simulation/gazebo_ros_pkgs/blob/3cd100e20ba09245be01ad674248aa31a46ad271/gazebo_plugins/src/gazebo_ros_depth_camera.cpp#L83
Even I would prefer changing the topic name

chapulina: I saw that topic being set, but it doesn't look like it's used anywhere, right?

shiveshkhaitan: https://github.com/ros-simulation/gazebo_ros_pkgs/blob/3cd100e20ba09245be01ad674248aa31a46ad271/gazebo_plugins/src/gazebo_ros_depth_camera.cpp#L312
This calls `PutCameraData` in [GazeboRosCameraUtils](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_camera_utils.cpp) which publishes the image

chapulina: Oh you're totally right! Nevermind, sorry for the noise.

> Even I would prefer changing the topic name

+1, let's do that and point it out on the migration notes

gazebo_plugins/src/gazebo_ros_camera.cpp
chapulina: Let's get this PR in, we can come back to this later if needed.

chapulina (approved):

gazebo_plugins/src/gazebo_ros_camera.cpp
shiveshkhaitan: > Let me know if you're still working on this

Sorry I missed this. 

> Let's get this PR in, we can come back to this later if needed.

I think its already done here

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/e2fd5c51249cc65ee6fa5533b6dba9619442a05b/gazebo_plugins/src/gazebo_ros_camera.cpp#L530-L532

chapulina: Yeah I think we could also use it here, but it's not a big deal

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/e2fd5c51249cc65ee6fa5533b6dba9619442a05b/gazebo_plugins/src/gazebo_ros_camera.cpp#L395-L412

PR #937, opened 2019-06-19
Fix for multiple video plugins (#898)

shiveshkhaitan:
Append the name of the plugin specified in the sdf to avoid crashing when multiple video plugins are used. This connects to #898

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_video.cpp
+++ b/gazebo_plugins/src/gazebo_ros_video.cpp
@@ -200,8 +200,8 @@ void GazeboRosVideo::Load(
 
   int width = _sdf->Get<int>("width", 320).first;
 
-  impl_->video_visual_ = std::make_shared<VideoVisual>(_parent->Name() + "::video_visual",
-      _parent, height, width);
+  impl_->video_visual_ = std::make_shared<VideoVisual>(_parent->Name() + "::video_visual::" +
+      _sdf->Get<std::string>("name"), _parent, height, width);
   _parent->GetScene()->AddVisual(impl_->video_visual_);
 
   // Subscribe to the image topic
--- a/gazebo_plugins/src/gazebo_ros_video.cpp
+++ b/gazebo_plugins/src/gazebo_ros_video.cpp
@@ -186,9 +186,6 @@ GazeboRosVideo::GazeboRosVideo()
 // Destructor
 GazeboRosVideo::~GazeboRosVideo()
 {
-  if (rclcpp::is_initialized()) {
-    rclcpp::shutdown();
-  }
   impl_->update_connection_.reset();
   impl_->rosnode_.reset();
 }
--- a/gazebo_plugins/src/gazebo_ros_video.cpp
+++ b/gazebo_plugins/src/gazebo_ros_video.cpp
@@ -186,6 +186,9 @@ GazeboRosVideo::GazeboRosVideo()
 // Destructor
 GazeboRosVideo::~GazeboRosVideo()
 {
+  if (rclcpp::is_initialized()) {
+    rclcpp::shutdown();
+  }
   impl_->update_connection_.reset();
   impl_->rosnode_.reset();
 }
--- a/gazebo_ros/src/node.cpp
+++ b/gazebo_ros/src/node.cpp
@@ -28,6 +28,7 @@ std::mutex Node::lock_;
 
 Node::~Node()
 {
+  executor_->remove_node(get_node_base_interface());
 }
 
 Node::SharedPtr Node::Get(sdf::ElementPtr sdf)


chapulina:
With this PR, I'm able to add another display to the demo world like this, and both displays work:

```.diff
diff --git a/gazebo_plugins/worlds/gazebo_ros_video_demo.world b/gazebo_plugins/worlds/gazebo_ros_video_$emo.world
index 46766d2..74e05b3 100644
--- a/gazebo_plugins/worlds/gazebo_ros_video_demo.world
+++ b/gazebo_plugins/worlds/gazebo_ros_video_demo.world
@@ -159,5 +159,34 @@
       </link>
     </model>

+    <model name="box_display_2">
+      <pose>3.0 0.0 0.0 0.0 0.0 0.0</pose>
+      <link name="base_link">
+        <visual name="visual">
+          <geometry>
+            <box>
+              <size>0.5 0.5 0.5</size>
+            </box>
+          </geometry>
+          <plugin name="display_video_controller_2" filename="libgazebo_ros_video.so">
+            <ros>
+              <!-- Remap for image topic to be subscribed -->
+              <argument>~/image_raw:=/camera1/image_raw</argument>
+            </ros>
+            <height>120</height>
+            <width>160</width>
+          </plugin>
+        </visual>
+        <collision name="collision">
+          <geometry>
+            <box>
+              <size>0.5 0.5 0.5</size>
+            </box>
+          </geometry>
+          <laser_retro>100.0</laser_retro>
+        </collision>
+      </link>
+    </model>
+
   </world>
```

But Gazebo seems to hang on shutdown and needs to be force-closed. Can you reproduce the issue? It would be good to look into it.

shiveshkhaitan:
Fixed the crashing of gazebo on shutdown

gazebo_plugins/src/gazebo_ros_video.cpp
```diff
@@ -186,6 +186,9 @@ GazeboRosVideo::GazeboRosVideo()
 // Destructor
 GazeboRosVideo::~GazeboRosVideo()
 {
+  if (rclcpp::is_initialized()) {
+    rclcpp::shutdown();
+  }
```
chapulina: This solution is a bit too drastic, because we still want all other plugins to keep working after one plugin is destroyed. You can try for example adding the 2nd video box as on my other comment and then deleting one of the boxes. This line will be called and the other box's video will stop.

Is there another way of fixing the shutdown issue?

shiveshkhaitan: For me the other box's video is stopping even without `rclcpp::shutdown()`. Also I show that this https://github.com/ros-simulation/gazebo_ros_pkgs/blob/120097242c8da6ef4eddbb1932eacfdbd10b7b03/gazebo_ros/include/gazebo_ros/node.hpp#L139-L142 goes into if multiple times when we have multiple video plugins. Is is supposed to happen?

shiveshkhaitan: Not only multiple video plugins, even if I have just one video plugin and a diff drive plugin, diff drive stops working on deleting the box

chapulina: > goes into if multiple times when we have multiple video plugins. Is is supposed to happen?

Ah that got me digging for a while. What's happening is that the video plugins are being loaded twice, once in `gzclient`, for the GUI, and once in `gzserver`, for the camera sensor. That happens for all `VisualPlugin`s.

If you start only `gzserver gazebo_ros_video_demo.world`, you can see that there are only 2 subscribers to camera images:

```
$ ros2 topic info /camera1/image_raw
Topic: /camera1/image_raw
Publisher count: 1
Subscriber count: 2
```

So I think the root of the shutdown problems we're having come from these duplicate nodes being created on both client and server. There are a few ways to tackle this. I think we should load that plugin only on the server scene and not on the client. I can't remember if we have an easy way to tell if we're on the client or server from the plugin though. Since [this PR](https://bitbucket.org/osrf/gazebo/pull-requests/2858/only-initialize-the-marker-service-in-the/diff) the `Scene` keeps a member variable, but that's not exposed. Maybe @iche033 remembers some nice trick to check it?

chapulina: Actually come to think of it, we probably want the plugin on both client and server... So maybe we need to give the nodes unique names.

iche033: if you ever need it, [Scene::EnableVisualizations](https://bitbucket.org/osrf/gazebo/src/a9a3fffb6194be39f6c582ed9b9ef89c019dfae9/gazebo/rendering/Scene.hh#lines-640) returns true if it's the client and false if it's the server

shiveshkhaitan: Setting different names for gzserver and gzclient did not help. But I think I figured out the problem. Adding this line `executor_->remove_node(get_node_base_interface());` in `gazebo_ros::Node` destructor
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/120097242c8da6ef4eddbb1932eacfdbd10b7b03/gazebo_ros/src/node.cpp#L29-L31
fixes it. What I think is that when the plugin is destroyed and the node is deleted, the executor does not remove it and tries to spin but gets stuck as the node is already deleted. The above line explicitly removes the node. Does it make sense @chapulina?

chapulina: Argh it sounds like https://github.com/ros-simulation/gazebo_ros_pkgs/pull/860 is back to haunt us. 

If I remember correctly, when debugging that with @wjwwood , the underlying issue seemed to be coming from the RMW layer. I don't remember if there was any problem adding the `remove_node` call, I think it just didn't fix the issue completely. So if it solves this specific issue without breaking anything else, I think we should get it in. Let's give it a try!

wjwwood: Isn't this essentially https://github.com/ros2/rclcpp/issues/272? I believe it should have been fixed with https://github.com/ros2/rclcpp/pull/741.

If not, then we should make sure there is an open issue on ros2/rclcpp which tracks the bug described in #860.

Unless I misunderstood and it's actually a bug in how rclcpp is being used here, I can't remember :x.

chapulina (approved):

gazebo_plugins/src/gazebo_ros_video.cpp
chapulina: @wjwwood , I just tried uncommenting the test that was failing:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/5e4a989148588272f5c02cd26d27349696823eff/gazebo_ros/test/test_node.cpp#L71-L108

And the situation is still the same:
* Without the `remove_node` call, the test fails every time
* With the `remove_node` call, the test never passes 10 times in a row

Unfortunately, I don't have time right now to create a failing test case using just `rclcpp` :confused:  If I remember correctly, we tried doing that together and it was complicated to reproduce what's happening here.

I'll merge the `remove_node` fix because it solves the issue for the video plugin, but I'll leave #855 open.

PR #938, opened 2019-06-20
Fix default principal point offset

AlexReimann:
There is an offset in the calculation of the principal point for the intrinsic camera calibration.

Assuming the pixels in the image are **zero-indexed** it has to be `- 1.0` instead of `+`.

E.g. image width of `4 px`:
`0 1 2 3` -> center is `1.5 = (width - 1) / 2.0 ` 

`5 px`:
`0 1 2 3 4` -> center is `2 = (width - 1) / 2.0 `

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_camera_utils.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera_utils.cpp
@@ -506,11 +506,11 @@ void GazeboRosCameraUtils::Init()
 
   /// Compute camera_ parameters if set to 0
   if (this->cx_prime_ == 0)
-    this->cx_prime_ = (static_cast<double>(this->width_) + 1.0) /2.0;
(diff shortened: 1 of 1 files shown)


chapulina (commented):
Thank you for the PR. Your math makes sense to me. The only concerns I have are:

* whether consumers of the message expect the principal point to be zero-indexed - I'm not sure if they do
* this would probably break users who are already adapting to the current output in their code, so it would be safer to release in Noetic instead of Melodic

sloretz:
@AlexReimann I rebased your branch onto `noetic-devel`; hope you don't mind.

AlexReimann:
I guess for backward compatibility I can add a flag for this.

chapulina:
> I can add a flag for this.

I think that would be ideal. Even if we get this into Noetic, users could be caught by surprise if they're not checking the changelog.

AlexReimann:
Not going to work on this anymore

PR #939, opened 2019-06-21
[ros2] Port multicamera to ros2

shiveshkhaitan:
Port gazebo_ros_multicamera plugin to ROS2. 

Example:
```
<plugin name="camera_controller" filename="libgazebo_ros_camera.so">
  <ros>
      <namespace>demo_cam</namespace>
      <!-- topics need to be prefixed with camera_name for remapping -->
      <argument>mycamera/left/image_raw:=mycamera/left/image_demo</argument>
      <argument>mycamera/right/image_raw:=mycamera/right/image_demo</argument>
       <argument>mycamera/left/camera_info:=mycamera/left/camera_info_demo</argument>
       <argument>mycamera/right/camera_info:=mycamera/right/camera_info_demo</argument>
    </ros>

  <camera_name>mycamera</camera_name>
  <sensor_type>2</sensor_type>
  <!-- frame_name>omit so it defaults to link name</frameName-->
</plugin>
```
Instead of a new plugin, multicamera has been integrated into `gazebo_ros_camera`. To use it as multicamera, <sensor_type> has to be specified as 2. It also includes `gazebo_ros_triggered_multicamera`


chapulina:
Now that #932 has been merged, this needs rebasing

gazebo_plugins/include/gazebo_plugins/MultiCameraPlugin.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+  // Store image format
+  std::vector<std::string> format;
+
+  // Pointer to multicamera
+  std::vector<rendering::CameraPtr> camera;
+
+private:
+  /// Connects to pre-render events.
+  std::vector<event::ConnectionPtr> newFrameConnection;
```
chapulina: Could you use the PIMPL pattern here? (i.e. `std::unique_ptr<MultiCameraPluginPrivate> dataPtr;`). The [newer](https://bitbucket.org/osrf/gazebo/src/2b59492f96fcf121b33f290fd5cf0b70b42f4e22/plugins/AmbientOcclusionVisualPlugin.cc#lines-97) Gazebo plugins use it.

gazebo_plugins/src/MultiCameraPlugin.cpp
```diff
@@ -0,0 +1,99 @@
...
+    this->camera.push_back(this->parentSensor->Camera(i));
+
+    // save camera attributes
+    this->width.push_back(this->camera[i]->ImageWidth());
+    this->height.push_back(this->camera[i]->ImageHeight());
+    this->depth.push_back(this->camera[i]->ImageDepth());
+    this->format.push_back(this->camera[i]->ImageFormat());
+
+    std::string cameraName = this->parentSensor->Camera(i)->Name();
+    // gzdbg << "camera(" << i << ") name [" << cameraName << "]\n";
```
chapulina: You can remove the commented out debug statements

gazebo_plugins/src/MultiCameraPlugin.cpp
```diff
@@ -0,0 +1,99 @@
...
+  unsigned int /*_width*/,
+  unsigned int /*_height*/,
+  unsigned int /*_depth*/,
+  const std::string & /*_format*/,
+  int /*camera_num*/)
+{
+  /*rendering::Camera::SaveFrame(_image, this->width,
+    this->height, this->depth, this->format,
+    "/tmp/camera/me.jpg");
+    */
```
chapulina: We can remove these comments

gazebo_plugins/include/gazebo_plugins/MultiCameraPlugin.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include <string>
+#include <vector>
+
+#include "gazebo/common/Plugin.hh"
+#include "gazebo/sensors/MultiCameraSensor.hh"
+#include "gazebo/rendering/Camera.hh"
+#include "gazebo/util/system.hh"
+
+namespace gazebo
```
chapulina: Can we put this plugin in the `gazebo_plugins` namespace? Just in case someone creates the same plugin for Gazebo in the future.

In fact, feel free to make a pull request adding this plugin to [Gazebo 9](https://bitbucket.org/osrf/gazebo/src/gazebo9/plugins/) if you have time. It would be just a matter of copying this and the cc file over, and making a couple tweaks to follow [Gazebo's style](http://gazebosim.org/tutorials?tut=contrib_code&cat=development#Style) guide (which is different from ROS 2).

I can help you out with style changes for you since I'm used to it. Some things I can already see:

* rename `.hpp` to `.hh` and `.cpp` to `.cc`
* Use `camelCase` (i.e. `cameraNum`)
* Prefix all arguments with `_` (i.e. `_cameraNum`)
* There should be one `public` / `protected` / `private` per member
* `{` in a new line

The only problem is that even if we release this plugin for Gazebo 9.X, it won't be available to most ROS users, which use Gazebo 9.0.0 from Ubuntu. So we woudn't be able to use it in `gazebo_ros_pkgs`'s debian packages :confused:

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -161,23 +174,48 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
...
+
+  } else {
+    for (uint64_t i = 0; i < impl_->num_cameras_; ++i) {
+      // Image publisher
+      impl_->image_pub_.push_back(image_transport::create_publisher(impl_->ros_node_.get(),
+        impl_->camera_name_ + "/" + gazebo::MultiCameraPlugin::camera[i]->Name() + "/image_raw"));
+
+      // RCLCPP_INFO(impl_->ros_node_->get_logger(), "Publishing %s camera images to [%s]",
+      //   gazebo::MultiCameraPlugin::camera[i]->Name().c_str(),
+      //   impl_->image_pub_.back().getTopic());
```
chapulina, line 206: You already have this below, you can remove the comment

chapulina (changes requested):
Works for me! The requests for changes are minimal, the only critical one is changing the plugin namespace.

gazebo_plugins/include/gazebo_plugins/MultiCameraPlugin.hpp
shiveshkhaitan: If we are using `gazebo_ros_pkgs` conventions, should `MultiCameraPlugin` be renamed to `multi_camera_plugin`?

shiveshkhaitan: And what about the style like should the  `camelCase` variables be changed to `snake_case`. Should I follow `gazebo_ros_pkgs` style or `gazebo` style?

chapulina: All code on `gazebo_ros_pkgs` should be following the [ROS 2 style guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#id1), so yeah could you please change variables and files to use `snake_case`?

gazebo_plugins/src/gazebo_ros_camera.cpp
shiveshkhaitan: The `RCLCPP_INFO` below this https://github.com/ros-simulation/gazebo_ros_pkgs/blob/c377a9632f92a7f5a9fd34ec78399c2b89ffb2e1/gazebo_plugins/src/gazebo_ros_camera.cpp#L214-L216 is for the camera info. This one is for the images.

gazebo_plugins/include/gazebo_plugins/MultiCameraPlugin.hpp
shiveshkhaitan: I will be making the PR soon

chapulina (approved):

chapulina:
LGTM. We could backport this to `dashing`.

PR #940, opened 2019-06-22
[ros2] Dynamic reconfigure for gazebo_ros_camera

shiveshkhaitan:
Update rate for camera is a ros2 parameter which can be dynamically changed. e.g.
`ros2 param set /demo_cam/camera_controller update_rate 5.0` 

This is not allowed in triggered cameras


gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -470,6 +460,66 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
...
+        }
+      }
+      for (const auto & parameter : parameters) {
+        if (nullptr != impl_->trigger_sub_) {
+          RCLCPP_WARN(impl_->ros_node_->get_logger(),
+            "Cannot set update rate for triggered camera");
+          result.successful = false;
+        } else {
+          rclcpp::ParameterType parameter_type = parameter.get_type();
+          if (rclcpp::ParameterType::PARAMETER_DOUBLE == parameter_type) {
```
chapulina: Should we also check `"update_rate" == parameter.get_name()`? Future-proof for when we have more than one param.

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -470,6 +460,66 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
...
+            double rate = parameter.as_double();
+
+            if (impl_->sensor_type_ == GazeboRosCameraPrivate::CAMERA) {
+              gazebo::CameraPlugin::parentSensor->SetUpdateRate(rate);
+            } else if (impl_->sensor_type_ == GazeboRosCameraPrivate::DEPTH) {
+              gazebo::DepthCameraPlugin::parentSensor->SetUpdateRate(rate);
+            } else {
+              MultiCameraPlugin::parent_sensor_->SetUpdateRate(rate);
+            }
+            result.successful = true;
```
chapulina: Since we could be setting multiple parameters, should the result be successful only if all parameters were successfully set (i.e. false if at least one param fails)?

chapulina (approved):
Works for me. I just have a couple of small comments. Also, it would be nice to add some instructions to `gazebo_ros_camera_demo.world` on how to get and set the rate, i.e.

`ros2 param get /demo_cam/camera_controller update_rate`

`ros2 param set /demo_cam/camera_controller update_rate 1.0`

PR #941, opened 2019-06-23
[ros2] Port apply/clear wrench and effort services

shiveshkhaitan:
Port `apply_body_wrench`, `clear_body_wrenches`, `apply_joint_effort` and `clear_joint_forces` services to ROS2. 

**Usage**
*Apply Body Wrench* 
`ros2 service call /apply_body_wrench gazebo_msgs/srv/ApplyBodyWrench '{body_name: "box::link", reference_frame: "", reference_point: { x: 100, y: 0, z: 0 }, wrench: { force: { x: 10, y: 0, z: 0 }, torque: { x: 0, y: 0, z: 0 } }, start_time: {sec: 0, nanosec: 0}, duration: {sec: -1, nanosec: 0} }'`           

*Clear Body Wrenches* 
`ros2 service call /clear_body_wrench gazebo_msgs/srv/BodyRequest '{body_name: "box::link"}' `

*Apply Joint Effort*
`ros2 service call /apply_joint_effort gazebo_msgs/srv/ApplyJointEffort '{joint_name: "joint", effort: -1.0, start_time: {sec: 0, nanosec: 0}, duration: {sec: 2000, nanosec: 0} }' `

*Clear Joint Forces*
`ros2 service call /clear_joint_forces gazebo_msgs/srv/JointRequest '{joint_name: "joint"}'`
`


gazebo_ros/src/gazebo_ros_effort.cpp
```diff
@@ -0,0 +1,451 @@
...
+  /// \param[in] req Request
+  /// \param[out] res Response
+  void ClearBodyWrenches(
+    gazebo_msgs::srv::BodyRequest::Request::SharedPtr _req,
+    gazebo_msgs::srv::BodyRequest::Response::SharedPtr _res);
+
+  /// \brief Callback for clear joint forces service.
+  /// \param[in] req Request
+  /// \param[out] res Response
+  void ClearJointForces(
```
chapulina: Nitpick: change `forces` to `effort` to match the rest of the API

gazebo_ros/src/gazebo_ros_effort.cpp
```diff
@@ -0,0 +1,451 @@
...
+    gazebo_msgs::srv::ApplyJointEffort::Request::SharedPtr _req,
+    gazebo_msgs::srv::ApplyJointEffort::Response::SharedPtr _res);
+
+  /// \bried Helper to shift wrench from reference frame to target frame
+  /// \param[in] reference_force Reference Frame Force
+  /// \param[in] reference_torque Reference Frame Torque
+  /// \param[in] target_to_reference Reference for transform
+  /// \param[out] target_force Shifted Force
+  /// \param[out] target_torque Shifted Torque
+  void transformWrench(
```
chapulina: Nitpick: Use `CamelCase` to be consistent within this file

gazebo_ros/src/gazebo_ros_effort.cpp
```diff
@@ -0,0 +1,451 @@
...
+  rclcpp::Service<gazebo_msgs::srv::JointRequest>::SharedPtr clear_joint_forces_service_;
+
+  /// To be notified once the world is created.
+  gazebo::event::ConnectionPtr world_created_connection_;
+
+  /// Connection to body wrench update event, called at every iteration
+  gazebo::event::ConnectionPtr body_wrench_update_event_;
+
+  /// Connection to joint effort update event, called at every iteration
+  gazebo::event::ConnectionPtr joint_effort_update_event_;
```
chapulina: We could have a single `update_connection_` and perform both body wrench and joint effort updates in the same callback.

gazebo_ros/src/gazebo_ros_effort.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+GazeboRosEffort::~GazeboRosEffort()
+{
+}
+
+void GazeboRosEffort::Load(int /* argc */, char ** /* argv */)
+{
+  // Keep this in the constructor for performance.
+  // sdf::initFile causes disk access.
+  sdf::initFile("root.sdf", impl_->factory_sdf_);
```
chapulina: `factory_sdf_` is not used anywhere, I think it can be removed

gazebo_ros/src/gazebo_ros_effort.cpp
```diff
@@ -0,0 +1,451 @@
...
+#include "gazebo_ros/conversions/builtin_interfaces.hpp"
+#include "gazebo_ros/gazebo_ros_effort.hpp"
+
+namespace gazebo_ros
+{
+class GazeboRosEffortPrivate
+{
+public:
+  /// Apply body wrench task
+  struct BodyWrenchTask
```
chapulina: I forgot to mention that we should be changing references to "body" to say "link" instead, because body is not a proper term in the Gazebo / SDF spec. Since this PR is already implemented, we can leave it as is, but this is something to keep in mind for the future. I've also added a [note](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-migration-contribution-guide/_compare/f7b8745fc45e4c2e68a62c208ad9cbc0ff753c15...6d544c7c24133cab22d7d17dff7f4273cdb5a6b1) about this to the migration guide.

gazebo_ros/include/gazebo_ros/gazebo_ros_effort.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <memory>
+
+namespace gazebo_ros
+{
+
+class GazeboRosEffortPrivate;
+
+/// Provides services and topics to
+/// Apply and clear wrenches on gazebo bodies.
+/// Apply and clear forces on gazebo joints
```
chapulina: Could you document the services provided by this plugin and what they do?

gazebo_ros/src/gazebo_ros_effort.cpp
```diff
@@ -0,0 +1,451 @@
...
+      joint_effort_tasks_.erase(joint_effort_task--);
+      RCLCPP_ERROR(ros_node_->get_logger(),
+        "Joint [%s] does not exist. Deleting task", (*joint_effort_task)->joint->GetName().c_str());
+    }
+
+    if (sim_time < start) {
+      continue;
+    }
+
+    if (sim_time <= end || duration.seconds() < 0.0) {
```
chapulina: We should support durations that are less than 1 second, so we need to check nanoseconds too

chapulina (changes requested):
I did a first pass. Mind adding an example world with a simple model that has a joint, and instructions on how to apply forces to links and joints?

gazebo_ros/src/gazebo_ros_effort.cpp
shiveshkhaitan: I will make the changes in this as well. Also do you have any suggestions for a better name for the plugin itself? `gazebo_ros_effort` seems to be indicating only the `joint_efforts` and not the wrenches. I thought of `gazebo_ros_force` but there already exists one in `gazebo_plugins`

chapulina: Yeah I agree that naming is complicated here. Maybe `gazebo_ros_apply_force`? or `gazebo_ros_force_system`, since it's a system plugin?

gazebo_ros/src/gazebo_ros_effort.cpp
shiveshkhaitan: I don't think nanonseconds would make a difference here. Here we are just checking if the duration is negative. Only `seconds` part of the duration can be negative. `Nanoseconds` cannot be negative as it is `uint32_t` type.

chapulina: Oh you're right. Nanoseconds are only important when comparing `sim_time` and `end`.

gazebo_msgs/CMakeLists.txt
```diff
@@ -33,9 +33,9 @@ set(msg_files
 )
 
 set(srv_files
-  "srv/ApplyBodyWrench.srv"
+  "srv/ApplyLinkWrench.srv"
```
chapulina, line 37: We should keep the `Body` messages for the `ros1_bridge`. Mind keeping the old files, with a `deprecated` note on the top?

gazebo_ros/src/gazebo_ros_force_system.cpp
```diff
@@ -0,0 +1,422 @@
...
+  void ClearJointEfforts(
+    gazebo_msgs::srv::JointRequest::Request::SharedPtr _req,
+    gazebo_msgs::srv::JointRequest::Response::SharedPtr _res);
+
+  /// \bried Helper to shift wrench from reference frame to target frame
+  /// \param[in] reference_force Reference Frame Force
+  /// \param[in] reference_torque Reference Frame Torque
+  /// \param[in] target_to_reference Reference for transform
+  /// \param[out] target_force Shifted Force
+  /// \param[out] target_torque Shifted Torque
```
chapulina, line 99: These params don't match the function

gazebo_ros/worlds/gazebo_ros_force_system_demo.world
```diff
@@ -0,0 +1,150 @@
...
+  Try for example:
+
+    To apply wrench to a link:
+
+      ros2 service call /apply_link_wrench gazebo_msgs/srv/ApplyLinkWrench '{link_name:
+      "single_pendulum_with_base::base", wrench: { force: { x: 100 } }, duration: {sec: -1, nanosec: 0} }'
+
+    To clear wrenches on a link:
+
+      ros2 service call /clear_link_wrenches 'gazebo_msgs/srv/LinkRequest'
```
chapulina: For easier copy-paste:

```suggestion
      ros2 service call /clear_link_wrenches 'gazebo_msgs/srv/LinkRequest' \
```

The same goes for the other commands

gazebo_ros/src/gazebo_ros_force_system.cpp
```diff
@@ -0,0 +1,422 @@
...
+  rclcpp::Time sim_time = Convert<builtin_interfaces::msg::Time>(_info.simTime);
+
+  std::lock_guard<std::mutex> scoped_lock(lock_);
+
+  for (auto link_wrench_task = link_wrench_tasks_.begin();
+    link_wrench_task != link_wrench_tasks_.end(); ++link_wrench_task)
+  {
+    auto start = (*link_wrench_task)->start_time;
+    auto duration = (*link_wrench_task)->duration;
+    auto end = start + duration;
```
chapulina, line 202: The message documentation says that **if start_time is not specified, or start_time < current time, start as soon as possible**. But what's happening here is that if `start + duration < sim_time`, the force is never applied.

I think there are two use cases here:

* User wants to apply force exactly between `start` and `end` time.
* User wants to immediately apply a force for a specific duration. 

In the first case, I think we want the current behaviour.

But for the 2nd case, I think we need to do something like:

```
// Start not specified
if (start == zero) {
  start = max(start, sim_time);
  auto end = start + duration;
}
```

Any thoughts? We don't need to change this now since it's just porting the ROS 1 code, but I thought it was strange when I send a command without a start time and the command never gets executed.

gazebo_ros/src/gazebo_ros_force_system.cpp
shiveshkhaitan: Sorry, I couldn't get the problem here. Is it something related to the description of the parameters?

gazebo_ros/src/gazebo_ros_force_system.cpp
shiveshkhaitan: `start == zero` would be true even if the actual start value is 0. This would lead to the force been applied infinitely. Is there any way to check if the user has passed any value for a particular element?

shiveshkhaitan: Also is there a reason for `start = max(start, sim_time)` and not `start = sim_time`

gazebo_ros/src/gazebo_ros_force_system.cpp
chapulina: Oh sorry, I just looked quickly and didn't realize they were documented in a different order from the order they're declared. We usually keep the same order.

gazebo_ros/src/gazebo_ros_force_system.cpp
chapulina: > start == zero would be true even if the actual start value is 0

Yeah that case wouldn't be well covered

> Is there any way to check if the user has passed any value for a particular element?

Unfortunately I don't think there is. By default the fields are initialized with zero.

> Also is there a reason for start = max(start, sim_time) and not start = sim_time

Yeah you're right, in that case we know start is zero, so it should be `sim_time` anyway

---

Thank you for brainstorming. We can leave the ROS 1 behaviour for now, since there doesn't seem to be a good compromise here.

chapulina (approved):

Pratham-Pandey:
- What do i need to include to use "apply_body_wrench"? '/apply_link_wrench' is available but 'apply_body_wrench' is not.
- Also what is the difference between them?
- Where can i find example usage? 
Thanks

PR #943, opened 2019-07-02
[ros2] Port bumper sensor to ROS2

shiveshkhaitan:
Port gazebo_ros_bumper plugin to ROS2. 

Example:
```
<plugin name="plugin_name" filename="libgazebo_ros_bumper.so">
  <ros>
    <namespace>demo</namespace>
    <argument>bumper_states:=bumper_demo</argument>
  </ros>
  <frame_name>frame</frame_name>
</plugin>
```
Migration guide at [ROS 2 Migration: Bumper](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Bumper)


gazebo_plugins/include/gazebo_plugins/gazebo_ros_bumper.hpp
```diff
@@ -0,0 +1,63 @@
...
+    <plugin name="plugin_name" filename="libgazebo_ros_bumper.so">
+      <!-- Change namespace and topics so:
+           Bumper state is published to: /custom_ns/custom_topic
+      -->
+      <ros>
+        <namespace>custom_ns</namespace>
+        <argument>bumper_states:=custom_topic</argument>
+      </ros>
+
+      <!-- Set TF frame name. If empty, defaults to link name -->
```
chapulina: Defaults to "world"?

gazebo_plugins/src/gazebo_ros_bumper.cpp
```diff
@@ -0,0 +1,226 @@
...
+  contact_state_msg.header.stamp =
+    rclcpp::Time(contacts.time().sec(), static_cast<uint32_t>(contacts.time().nsec()));
+
+  ignition::math::Pose3d pose, frame_pose;
+  ignition::math::Quaterniond rot, frame_rot;
+  ignition::math::Vector3d pos, frame_pos;
+
+  frame_pos = ignition::math::Vector3d(0, 0, 0);
+  frame_rot = ignition::math::Quaterniond(1, 0, 0, 0);  // gazebo u,x,y,z == identity
+  frame_pose = ignition::math::Pose3d(frame_pos, frame_rot);
```
chapulina: It's not necessary to zero these, because the Ignition math types are automatically initialized to zero / identity.

But in any case, these variables are not needed since we always use the world frame. `rot`, `pos`, `pose` and `frame_pose` are never used.  `frame_rot` and `frame_pos` are used, but since they're always the identity, they don't make a difference and can be removed.

gazebo_plugins/src/gazebo_ros_bumper.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+  impl_->parent_sensor_->SetActive(true);
+}
+
+void GazeboRosBumperPrivate::OnUpdate()
+{
+  gazebo::msgs::Contacts contacts;
+  contacts = parent_sensor_->Contacts();
+
+  gazebo_msgs::msg::ContactsState contact_state_msg;
```
chapulina: I suggest the logic in this function to be turned into a `convert` function from `gazebo::msgs::Contacts` to `gazebo_msgs::msg::ContactsState`, so other plugins can use it too.

gazebo_plugins/src/gazebo_ros_bumper.cpp
```diff
@@ -0,0 +1,226 @@
...
+      ")     my geom:" << state.collision1_name <<
+      "   other geom:" << state.collision2_name <<
+      "         time:" << contact.time().sec() << "." << contact.time().nsec() << std::endl;
+
+    state.info = stream.str();
+
+    state.wrenches.clear();
+    state.contact_positions.clear();
+    state.contact_normals.clear();
+    state.depths.clear();
```
chapulina: Are all these `clear()` calls needed? The message is brand-new.

gazebo_plugins/src/gazebo_ros_bumper.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    // sum up all wrenches for each DOF
+    geometry_msgs::msg::Wrench total_wrench;
+    total_wrench.force.x = 0;
+    total_wrench.force.y = 0;
+    total_wrench.force.z = 0;
+    total_wrench.torque.x = 0;
+    total_wrench.torque.y = 0;
+    total_wrench.torque.z = 0;
```
chapulina: All these zeroes should also not be needed

gazebo_plugins/src/gazebo_ros_bumper.cpp
```diff
@@ -0,0 +1,226 @@
...
+      // loop through individual contacts between collision1 and collision2
+      // gzerr << j << "  Position:"
+      //       << contact.position(j).x() << " "
+      //       << contact.position(j).y() << " "
+      //       << contact.position(j).z() << "\n";
+      // gzerr << "   Normal:"
+      //       << contact.normal(j).x() << " "
+      //       << contact.normal(j).y() << " "
+      //       << contact.normal(j).z() << "\n";
+      // gzerr << "   Depth:" << contact.depth(j) << "\n";
```
chapulina: The debug statements can be removed

gazebo_plugins/src/gazebo_ros_bumper.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+#include <memory>
+#include <string>
+
+namespace gazebo_plugins
+{
+class GazeboRosBumperPrivate
+{
+public:
+  /// Callback to be called on contact.
```
chapulina: Actually this is being called whenever the sensor is updated, regardless of whether there has been a contact or not.

gazebo_plugins/src/gazebo_ros_bumper.cpp
```diff
@@ -0,0 +1,226 @@
...
+    // For each collision contact
+    // Create a ContactState
+    gazebo_msgs::msg::ContactState state;
+    const gazebo::msgs::Contact & contact = contacts.contact(i);
+
+    state.collision1_name = contact.collision1();
+    state.collision2_name = contact.collision2();
+    std::ostringstream stream;
+    stream << "Debug:  i:(" << i << "/" << contactsPacketSize <<
+      ")     my geom:" << state.collision1_name <<
```
chapulina: Nitpick: add a space after `:`s. Or to make it even clearer, wrap the collision names on `[]` so they're easy to spot. Right now the message is a bit hard to read: `Debug:  i:(0/1)     my geom:ball::link::collision   other geom:ground_plane::link::collision    
     time:12.862000000`.

Also, `0/1` looks a bit weird if it's the only message. Maybe print `i + 1`?

chapulina (changes requested):
Works for me! I just have some small comments.

gazebo_plugins/src/gazebo_ros_bumper.cpp
shiveshkhaitan: Wont that be the whole of OnUpdate function?

chapulina: Yeah pretty much :sweat_smile: A good parallel is `gazebo_ros_ray_sensor`, which basically takes in Gazebo messages, converts them into ROS messages, and publishes:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/5e4a989148588272f5c02cd26d27349696823eff/gazebo_plugins/src/gazebo_ros_ray_sensor.cpp#L209-L217

gazebo_ros/include/gazebo_ros/conversions/gazebo_msgs.hpp
```diff
@@ -0,0 +1,131 @@
...
+    gazebo_msgs::msg::ContactState state;
+    const gazebo::msgs::Contact & contact = in.contact(i);
+
+    state.collision1_name = contact.collision1();
+    state.collision2_name = contact.collision2();
+    std::ostringstream stream;
+    stream << "Debug:  i: (" << i + 1 << "/" << contacts_packet_size <<
+      ")     my_geom: [" << state.collision1_name <<
+      "]  other_geom: [" << state.collision2_name <<
+      "]        time: [" << contact.time().sec() << "." << contact.time().nsec() << "]\n";
```
chapulina, line 65: Interesting that `in.contact(i).time()` (printed here) is different from `in.time()` (put in the header):

```
header:                                                                                                  
  stamp:                                                                                                 
    sec: 47                                                                                              
    nanosec: 466000000                                                                                   
  frame_id: world                                                                                        
states:                                                                                                  
- info: 'Debug:  i: (1/1)     my_geom: [ball::link::collision]  other_geom: [ground_plane::link::collisio
n]        time: [47.455000000]   
```

I think each contact has its own specific time, but the ROS message doesn't capture that.

This is just an observation, no action needed.

gazebo_ros/test/test_conversions.cpp
```diff
@@ -15,6 +15,7 @@
 #include <gazebo_ros/conversions/builtin_interfaces.hpp>
 #include <gazebo_ros/conversions/sensor_msgs.hpp>
 #include <gazebo_ros/conversions/geometry_msgs.hpp>
+#include <gazebo_ros/conversions/gazebo_msgs.hpp>
```
chapulina: Did you mean to add a test here?

shiveshkhaitan: I couldn't think of any test that may be required. Actually I saw `sensor_msgs` being included without a test so I thought I should include it.

chapulina: humm interesting, the sensors header is not needed. Ideally we'd have unit tests for all conversion functions. Right now they're being indirectly tested through `gazebo_plugins` tests.

shiveshkhaitan: Should I remove this for now?

chapulina: yeah do you mind removing both sensors and gazebo? Thanks!

chapulina (approved):

chapulina (approved):
We shouldn't forget to backport to Dashing

PR #944, opened 2019-07-03
[ros2] Port harness to ROS2

shiveshkhaitan:
Port `gazebo_ros_harness` plugin to ROS2. 

Example:
```
<plugin name="gazebo_ros_harness" filename="libgazebo_ros_harness.so">
  <ros>
    <!-- Add a namespace -->
    <namespace>demo</namespace>
    <argument>box/harness/velocity:=velocity_demo</argument>
    <argument>box/harness/detach:=detach_demo</argument>
  </ros>
  <!-- set initial harness velocity -->
  <init_vel>-0.1</init_vel>
</plugin>
```
Migration guide at [ROS 2 Migration: Harness](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Harness)


chapulina:

chapulina (approved):
Nice, I like the new init vel feature and the change from `Bool` to `Empty`!

The test passes for me and the OSRF farm, but it fails on the ROS farm with the same problem the elevator is having: `libHarnessPlugin.so: cannot open shared object file: No such file or directory`.

I ticketed #970 for this. Since it doesn't seem to be an issue specific to this PR, I'll merge this as soon as it's rebased.

PR #945, opened 2019-07-04
[ros2] Port elevator to ROS2

shiveshkhaitan:
Port `gazebo_ros_elevator` plugin to ROS2. 

Example:
```
 <plugin name="elevator" filename="libgazebo_ros_elevator.so">
  <ros>
    <namespace>demo</namespace>
    <argument>elevator:=elevator_demo</argument>
  </ros>
  <bottom_floor>0</bottom_floor>
  <top_floor>1</top_floor>

  <lift_joint>elevator::lift</lift_joint>
  <door_joint>elevator::door</door_joint>
  <floor_height>3.075</floor_height>

  <!-- Time the elevator door will stay open in seconds -->
  <door_wait_time>10</door_wait_time>
</plugin>
```
Migration guide at [ROS 2 Migration: Elevator](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Elevator)


gazebo_plugins/src/gazebo_ros_elevator.cpp
```diff
@@ -0,0 +1,86 @@
...
+    std::bind(&GazeboRosElevator::OnElevator, this, std::placeholders::_1));
+
+  RCLCPP_INFO(impl_->ros_node_->get_logger(), "Subscribed to [%s]", impl_->sub_->get_topic_name());
+
+  impl_->bottom_ = _sdf->Get<int>("bottom_floor", std::numeric_limits<int>::min()).first;
+
+  impl_->top_ = _sdf->Get<int>("top_floor", std::numeric_limits<int>::max()).first;
+}
+
+void GazeboRosElevator::OnElevator(const std_msgs::msg::String::ConstSharedPtr msg)
```
chapulina, line 67: So strange that the original plugin takes a string when we need an int :thinking: We could easily change this to `Int32` for a cleaner implementation.

I wonder if it was done like this because Gazebo's `ElevatorPlugin` uses a string message. And the reason for that is that `gazebo::msgs` doesn't have an integer message...

chapulina (approved):
Works for me!

I noticed that the elevator sometimes doesn't respond to commands. I needed to send a command a couple of times before it moved. But the problem seems to be in the Gazebo plugin, because I checked that the `MoveToFloor` commands are going through.

Merging!

PR #947, opened 2019-07-05
[ros2] Add ackermann drive plugin

shiveshkhaitan:
Added ackermann drive plugin for ROS2. It subscribe to geometry_msgs/Twist for commands. 

Example usage:
```
<plugin name="gazebo_ros_ackermann_drive" filename="libgazebo_ros_ackermann_drive.so">
  <ros>
    <namespace>demo</namespace>
    <argument>distance:=distance_demo</argument>
  </ros>

  <update_rate>100.0</update_rate>

  <!-- wheels -->
  <front_left_joint>front_left_wheel_joint</front_left_joint>
  <front_right_joint>front_right_wheel_joint</front_right_joint>
  <rear_left_joint>rear_left_wheel_joint</rear_left_joint>
  <rear_right_joint>rear_right_wheel_joint</rear_right_joint>
  <left_steering_joint>front_left_steer_joint</left_steering_joint>
  <right_steering_joint>front_right_steer_joint</right_steering_joint>
  <steering_wheel_joint>steering_joint</steering_wheel_joint>

  <!-- Max absolute steer angle for tyre in radians-->
  <max_steer>0.6458</max_steer>
  <!-- Max absolute steering angle of steering wheel -->
  <max_steering_angle>7.85</max_steering_angle>
  <!-- Max absolute linear speed in m/s -->
  <max_speed>20</max_speed>

  <robot_base_frame>chassis</robot_base_frame>
</plugin>
  ```

### CODE REVIEW

File: gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp [hunk 1]
```diff
@@ -0,0 +1,592 @@
...
+  auto world = impl_->model_->GetWorld();
+  auto physicsEngine = world->Physics();
+  physicsEngine->SetParam("friction_model", std::string("cone_model"));
+
+  // Initialize ROS node
+  impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
+
+  auto chassis_name = _sdf->Get<std::string>("chassis", "chassis").first;
+  auto chassis = impl_->model_->GetLink(chassis_name);
+  if (!chassis) {
```
chapulina: Is this needed? I don't see `chassis` being used anywhere.

gazebo_plugins/worlds/gazebo_ros_ackermann_drive_demo.world
```diff
@@ -0,0 +1,660 @@
+<?xml version="1.0"?>
+<!--
+  Gazebo ROS ackermann drive plugin demo
+
+  Try sending commands:
+
+    ros2 topic pub /demo/cmd_demo geometry_msgs/Twist '{linear: {x: 5.0}}' -1
+
+    ros2 topic pub /demo/cmd_demo geometry_msgs/Twist '{angular: {z: 0.1}}' -1
```
chapulina: Since Ackermann steering can't rotate in-place, this command doesn't do anything. We need to add a linear component for it to move.

gazebo_plugins/worlds/gazebo_ros_ackermann_drive_demo.world
```diff
@@ -0,0 +1,660 @@
...
+  <world name="default">
+    <include>
+      <uri>model://sun</uri>
+    </include>
+
+    <include>
+      <uri>model://ground_plane</uri>
+    </include>
+
+    <model name='vehicle'>
```
chapulina: Right now the model is slightly different from the one on the database. How about including the model from the database instead of describing a new model to avoid duplication? Something like this:

```
    <include>
      <uri>model://prius_hybrid</uri>
      <plugin ....>
        ...
       </plugin>
    </include>
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
```diff
@@ -0,0 +1,106 @@
...
+      <!-- PID tuning -->
+      <left_steering_p_gain>1000</left_steering_p_gain>
+      <left_steering_i_gain>100</left_steering_i_gain>
+      <left_steering_d_gain>1e15</left_steering_d_gain>
+      <right_steering_p_gain>1000</right_steering_p_gain>
+      <right_steering_i_gain>100</right_steering_i_gain>
+      <right_steering_d_gain>1e15</right_steering_d_gain>
+      <linear_velocity_p_gain>1300</linear_velocity_p_gain>
+      <linear_velocity_i_gain>0.1</linear_velocity_i_gain>
+      <linear_velocity_d_gain>10000</linear_velocity_d_gain>
```
chapulina: This is ok as it is, but I'll just leave a suggestion which could be slightly more compact, using vectors:

`<left_steering_pid>1000 100 1e15</left_steering_pid>`

and then

```.cpp
math::Vector3d pid = _sdf->Get<ignition::math::Vector3d>("left_steering_pid", ignition::math::Vector3d::Zero).first;
```

Feel free to ignore this :wink:

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
```diff
@@ -0,0 +1,592 @@
...
+    return;
+  }
+
+  auto steering_wheel_joint =
+    _sdf->Get<std::string>("steering_wheel_joint", "steering_wheel_joint").first;
+  impl_->joints_[GazeboRosAckermannDrivePrivate::STEER_WHEEL] =
+    _model->GetJoint(steering_wheel_joint);
+  if (!impl_->joints_[GazeboRosAckermannDrivePrivate::STEER_WHEEL]) {
+    RCLCPP_ERROR(impl_->ros_node_->get_logger(),
+      "Steering wheel joint [%s] not found, plugin will not work.", steering_wheel_joint.c_str());
```
chapulina: How about making the steering wheel optional? I think most models may not have this level of detail. We'll need to be sure to check if it's not null below.

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
```diff
@@ -0,0 +1,592 @@
...
+
+  // Compute the angle ratio between the steering wheel and the tires
+  impl_->steering_ratio_ = impl_->max_steer_ / max_steering_angle;
+
+  double pgain, dgain, igain, imax, imin;
+
+  pgain = _sdf->Get<double>("left_steering_p_gain", 0.0).first;
+  igain = _sdf->Get<double>("left_steering_i_gain", 0.0).first;
+  dgain = _sdf->Get<double>("left_steering_d_gain", 0.0).first;
+  impl_->pid_[GazeboRosAckermannDrivePrivate::FRONT_LEFT].Init(pgain, dgain, igain);
```
chapulina: Should this be `STEER_LEFT`? Also check the other 2 PID joints below.

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
```diff
@@ -0,0 +1,592 @@
...
+  /// \param[in] _info Updated simulation info.
+  void OnUpdate(const gazebo::common::UpdateInfo & _info);
+
+  /// Callback when a velocity command is received.
+  /// \param[in] _msg Twist command message.
+  void OnCmdVel(geometry_msgs::msg::Twist::SharedPtr _msg);
+
+  /// Extracts radius of a cylinder or sphere collision shape. Otherwise returns zero
+  /// \param[in] _coll Pointer to collision
+  double CollisionRadius(const gazebo::physics::CollisionPtr & _coll);
```
chapulina, line 81: Missing `\return`

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
```diff
@@ -0,0 +1,592 @@
...
+void GazeboRosAckermannDrivePrivate::UpdateOdometryWorld()
+{
+  auto prev_x = odom_.pose.pose.position.x;
+  auto prev_y = odom_.pose.pose.position.y;
+
+  auto pose = model_->WorldPose();
+  odom_.pose.pose.position = gazebo_ros::Convert<geometry_msgs::msg::Point>(pose.Pos());
+  odom_.pose.pose.orientation = gazebo_ros::Convert<geometry_msgs::msg::Quaternion>(pose.Rot());
+
+  distance_.data += hypot(prev_x - odom_.pose.pose.position.x, prev_y - odom_.pose.pose.position.y);
```
chapulina, line 529: This seems to be drifting too fast. I let the Prius sit for 30 sec and this is publishing a distance of 0.19 m. But if I check the model pose on the GUI, it's just -0.09 on X. Maybe we need to double-check the math?

gazebo_plugins/worlds/gazebo_ros_ackermann_drive_demo.world
```diff
@@ -0,0 +1,660 @@
...
+  Try listening to odometry:
+
+    ros2 topic echo /demo/odom_demo
+
+  Try listening to TF:
+
+    ros2 run tf2_ros tf2_echo odom_demo chassis
+
+    and other links
+-->
```
chapulina, line 22: You could also add a line about `ros2 topic echo /demo/distance_demo`.

chapulina (changes requested):
I did a first pass, and the Prius is drivable :+1: 

Besides my other comments, I also noticed that there's a problem with the front wheels TF:

wheel_tf

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
shiveshkhaitan: I don't think there is a problem with the math. Actually the problem is that the model pose seems to be oscillating in the Y direction by very small amount when static. So over time the error accumulates.

gazebo_plugins/worlds/gazebo_ros_ackermann_drive_demo.world
shiveshkhaitan: The model in the database does not have all the joints and links required. I got this model from [osrf/car_demo](https://github.com/osrf/car_demo/blob/master/prius_description/urdf/prius.urdf). I removed the sensors and changed the mesh files to load from those in `prius_hybrid` which the database contains.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
shiveshkhaitan: Can I do it this way for 5 values?

shiveshkhaitan:
> Besides my other comments, I also noticed that there's a problem with the front wheels TF:

Isn't it weird? The car seems to be fine in gazebo. Any idea what could be the problem?

gazebo_plugins/worlds/gazebo_ros_ackermann_drive_demo.world
chapulina: Oh interesting, the models on the database use universal joints on their front axle. On [citysim](https://bitbucket.org/osrf/citysim/src/d258e487ca560159a3b592f77642e5bd125969c0/worlds/simple_city.world#lines-14077) we used `libPriusHybridPlugin` with the models from the database and used the same joint for `front_left_wheel` and `front_left_wheel_steering`. I honestly don't know how that was working though :thinking:

We can leave this the way it is since it works, but now I'm just curious about citysim.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
chapulina: No :confused: `math::Pose3d` takes 6 numbers, but depending on the use case it could be messy to use that, especially because the last 3 are rotation so they get stored as quaternions...

gazebo_plugins/worlds/gazebo_ros_ackermann_drive_demo.world
shiveshkhaitan: I tried `ackermann_drive` the same way (specifying same joint for `front_left` and `front_left_steering` and including model from database) and it works. Also the wrong TF problem and the vibration in `demo/distance_demo` are fixed. I think we will use this then

chapulina: I just noticed these lines are wrong:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/b3426432d64322f57c6b233a5dd6741c1ccb71f0/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp#L471-L472

The argument passed to `Position()` is the axis index. But no joints have 3 axes, so an index of 2 is invalid. The reason why nothing breaks when using a revolute joint is that the [index is ignored](https://bitbucket.org/osrf/gazebo/src/9e18a31cda6328c1d8d4ec08b9d1b6cfcdb93585/gazebo/physics/ode/ODEHingeJoint.cc#lines-128). But for a universal joint, an index of `2` should [print an error](https://bitbucket.org/osrf/gazebo/src/9e18a31cda6328c1d8d4ec08b9d1b6cfcdb93585/gazebo/physics/ode/ODEUniversalJoint.cc#lines-145).

If we want to support both revolute and universal joints, we could use an index `1`, which should work at least for ODE. I haven't checked the other engines.

shiveshkhaitan: Yes,  I had to change that to index 0 to make it work.

shiveshkhaitan: But I am having some problems with the odometry in the database model.  I guess model's X is not the car's front.  When the car moves,  the odom moves in Y direction

chapulina: Ah interesting, the database model is facing -Y for some reason. We don't need to change the model if it will be too much work. It would be nice to document these requirements that we're finding though. That the joints need to be revolute with one axis, the model must be facing +X, etc.

shiveshkhaitan: I have used the database model in the latest commit as it was more stable wrt odometry and had correct TF. I had to change the model to rotate it, so could not avoid duplication of the model.

chapulina:
@shiveshkhaitan , mind fixing conflicts?

chapulina:
> The car seems to be fine in gazebo. Any idea what could be the problem?

With the latest commits, the front wheels look and steering wheel look good to me, but the rear wheels are not showing. 

This is printing some wrong info:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/2484790b90f53fce68dd6f3a55190f8311245c71/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp#L397-L398

More specifically:

```
[INFO] [demo.ackermann_drive]: Publishing wheel transforms between [chassis], [rear_right_wheel_joint] and [rear_right_wheel_joint]
[INFO] [demo.ackermann_drive]: Publishing wheel transforms between [chassis], [rear_left_wheel_joint] and [rear_left_wheel_joint]
```

Inspecting the TF tree on RViz, we can see that the rear wheels are disconnected from the rest of the tree. The parent of the rear wheels is the `rear_axle`, which is the child of `chassis`. We either need:

* To publish transforms from `rear_axle` to `chassis`
* To calculate transform from `rear_*` wheels to `chassis` and publish that.

I think the 2nd option is better to be generalized for other kinematic structures which don't have an intermediate link like `rear_axle`.

shiveshkhaitan:
> To calculate transform from rear_* wheels to `chassis and publish that.

Should this be published only for the demo? Because if we integrate into the code, then it won't generalise again

chapulina:
The user chooses the base link (`robot_base_frame`) and the joints. So if we want to be general, we can provide the transform from that link to each joint. In some cases, that link will be the immediate parent of the joints. But in other cases, like this demo, there will be links in between.

So what we can do is get the parent until we find one which matches the base link name. Here:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/2484790b90f53fce68dd6f3a55190f8311245c71/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp#L557

Instead of getting just the relative pose from the joint to its child link, we need to accumulate the relative poses of all joints and links until we find the base link. I think you'll need to use `Joint::GetParent` and `Link::GetParentJointsLinks`.

Or we can just, for now, not publish the wheel transforms in case the joint parent's name is different from `robot_base_frame`.

chapulina:

chapulina (approved):
Great addition, LGTM!

chapulina:
I just realized that I merged this without checking CI and the test is failing sometimes. Can you take a look, @shiveshkhaitan ?

PR #948, opened 2019-07-06
[ros2] Port spawn model to ROS2

shiveshkhaitan:
Port gazebo_ros node `spawn_model`  to ROS2.

Example usage:
`spawn_model.py -model name -file /path/to/model/file.sdf -x 5.0 -y 5.0`


gazebo_ros/scripts/spawn_model.py
```diff
@@ -0,0 +1,312 @@
...
+            <include>
+                <uri>model://{}</uri>
+            </include>
+        </world>
+    </sdf>"""
+
+    def __init__(self, args):
+        super().__init__('spawn_model')
+        parser = argparse.ArgumentParser(
+            description='Spawn a model in gazebo using the Gazebo ROS Factory')
```
chapulina: We should document somewhere that Gazebo should be started with `libgazebo_ros_factory.so`, `libgazebo_ros_init.so` (and `libgazebo_ros_state.so`?) for all functionality to work

gazebo_ros/scripts/spawn_model.py
```diff
@@ -0,0 +1,312 @@
...
+        """
+        # Wait for model to exist if wait flag is enabled
+        if self.args.wait:
+            self.model_exists = False
+
+            def models_cb(models):
+                self.model_exists = self.args.wait in models.name
+
+            self.subscription = self.create_subscription(
+                ModelStates, '%s/model_states' % self.args.gazebo_namespace, models_cb, 10)
```
chapulina: This topic doesn't exist in ROS2 yet, we could add it to `gazebo_ros_state`

gazebo_ros/scripts/spawn_model.py
```diff
@@ -0,0 +1,312 @@
...
+        # if self.args.bond:
+        #     rospy.on_shutdown(self._delete_entity)
+        #     rospy.loginfo('Waiting for shutdown to delete model {}'.format(self.args.model))
+        #     rospy.spin()
+
+        return 0
+
+    def _spawn_entity(self, model_xml, initial_pose):
+        self.get_logger().info('Waiting for service %s/spawn_entity' % self.args.gazebo_namespace)
+        client = self.create_client(SpawnEntity, 'spawn_entity')
```
chapulina: The messages include `gazebo_namespace`, but the client doesn't actually use it. I think we should use it, and change the default value to empty, because that's the default namespace offered by the plugins on the other end.

gazebo_ros/scripts/spawn_model.py
```diff
@@ -0,0 +1,312 @@
+#!/usr/bin/env python3
```
chapulina: How about renaming this script to `spawn_entity`, like we did with the service? It should work not only for models, but also for lights, as explained in the [migration page](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Spawn-and-delete#summary).

gazebo_ros/scripts/spawn_model.py
```diff
@@ -0,0 +1,312 @@
...
+        initial_pose.position.y = float(self.args.y)
+        initial_pose.position.z = float(self.args.z)
+
+        q = quaternion_from_euler(self.args.R, self.args.P, self.args.Y)
+        initial_pose.orientation.w = q[0]
+        initial_pose.orientation.x = q[1]
+        initial_pose.orientation.y = q[2]
+        initial_pose.orientation.z = q[3]
+
+        # Spawn model using urdf or sdf service based on arguments
```
chapulina: This comment looks outdated

gazebo_ros/scripts/spawn_model.py
```diff
@@ -0,0 +1,312 @@
...
+                            help='yaw angle of initial orientation, radians')
+
+        # TODO(shivesh): Wait for /set_model_configuration
+        # (https://github.com/ros-simulation/gazebo_ros_pkgs/issues/779)
+        # parser.add_argument('-J', dest='joints', default=[], action='append',
+        #     metavar=('JOINT_NAME', 'JOINT_POSITION'), type=str, nargs=2,
+        #     help='initialize the specified joint at the specified position')
+
+        parser.add_argument('-package_to_model', action='store_true', help='convert urdf \
+                            <mesh filename="package://..." to <mesh filename="model://..."')
```
chapulina: Did you give this parameter a try? How do you recommend I try it out?

gazebo_ros/scripts/spawn_model.py
```diff
@@ -0,0 +1,312 @@
...
+        #     joint_names = [joint[0] for joint in self.args.joints]
+        #     joint_positions = [joint[1] for joint in self.args.joints]
+        #     success = _set_model_configuration(joint_names, joint_positions)
+        #     if not success:
+        #         self.get_logger().error('SetModelConfiguration service failed. Exiting.')
+        #         return 1
+
+        # Unpause physics if user requested
+        if self.args.unpause:
+            client = self.create_client(Empty, '/unpause_physics')
```
chapulina: Should we use `gazebo_namespace` here too?

gazebo_ros/scripts/spawn_model.py
```diff
@@ -0,0 +1,312 @@
...
+        self.get_logger().info('Calling service %s/spawn_entity' % self.args.gazebo_namespace)
+        srv_call = client.call_async(req)
+        while rclpy.ok():
+            if srv_call.done():
+                self.get_logger().info('Spawn status: %s' % srv_call.result().status_message)
+                break
+            rclpy.spin_once(self)
+        return srv_call.result().success
+
+    def _delete_entity(self):
```
chapulina: This is currently not being used, so we can comment it out.

chapulina (changes requested):
I did a first pass, but haven't finished reviewing and trying out all parameters.

Mind adding a note about this script to the [spawn migration](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Spawn-and-delete) page?

Also, it would be nice to be able to call the script with `ros2 run gazebo_ros spawn_entity ...`, and to be able to call it from launch files. I think we may need a [setup.cfg](https://github.com/ros2/examples/blob/dashing/rclpy/topics/minimal_publisher/setup.cfg) file to install it correctly under `lib`, but it may also be ok to do it with cmake. Once that's working, we could add an example launch file which spawns a model at startup. That's a very common use case on ROS 1.

gazebo_ros/scripts/spawn_model.py
shiveshkhaitan: "Setting parameters from the command-line is currently only supported in the form of yaml files". This is what the [documentation](https://index.ros.org//doc/ros2/Tutorials/Node-arguments/) says. Should we do it?

chapulina: Oh sorry, I chose the word poorly. I didn't mean we should use parameters, I should have said "argument".

I just meant to ask how can we try this out to check it works. Did you create a dummy model which uses `<mesh filename="package://`?

shiveshkhaitan: Oh. For that I printed just created a random urdf.
```
<?xml version="1.0"?>
<robot name="robot">
  <link name="base_link">
    <inertial>
      <origin xyz="0 0 0" rpy="0 0 0" />
      <mass value="10" />
      <inertia ixx="1" ixy="1" ixz="1" iyy="1" iyz="1" izz="1" />
    </inertial>
    <visual>
      <origin xyz="0 0 0" rpy="0 0 0" />
      <geometry>
        <mesh filename="package://robot_description/meshes/base_link.STL" />
      </geometry>
      <material name="">
        <color rgba="1 1 1 1" />
      </material>
    </visual>
    <collision>
      <origin xyz="0 0 0" rpy="0 0 0" />
      <geometry>
        <mesh filename="package://robot_description/meshes/base_link.STL" />
      </geometry>
    </collision>
  </link>
</robot>
```
And did `print(model_xml)` before line 171 and after 185 to see the change in the xml

gazebo_ros/scripts/spawn_model.py
shiveshkhaitan: I have commented it for now. Will get back to it after porting model states

gazebo_ros/launch/spawn_entity_demo.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
chapulina: roslaunch is not finding this file for me. If I rename it to `spawn_entity_demo.launch.py` the command line autocompletes and the file can be launched.

gazebo_ros/scripts/spawn_entity.py
```diff
@@ -0,0 +1,319 @@
...
+        #     rospy.on_shutdown(self._delete_entity)
+        #     rospy.loginfo('Waiting for shutdown to delete entity {}'.format(self.args.entity))
+        #     rospy.spin()
+
+        return 0
+
+    def _spawn_entity(self, entity_xml, initial_pose):
+        self.get_logger().info('Waiting for service %s/spawn_entity' % self.args.gazebo_namespace)
+        client = self.create_client(SpawnEntity, '%s/spawn_entity' % self.args.gazebo_namespace)
+        client.wait_for_service(timeout_sec=5.0)
```
chapulina: Should we return if this returns false (i.e. wait timed out) and print an error to the user? Otherwise there's no point in making the request.

I think a message could help users that started Gazebo without the factory plugin. Something like "Failed to find /spawn_entity service. Was Gazebo initialized with gazebo_ros_factory?".

gazebo_ros/scripts/spawn_entity.py
```diff
@@ -0,0 +1,319 @@
...
+        self.get_logger().info('Calling service %s/spawn_entity' % self.args.gazebo_namespace)
+        srv_call = client.call_async(req)
+        while rclpy.ok():
+            if srv_call.done():
+                self.get_logger().info('Spawn status: %s' % srv_call.result().status_message)
+                break
+            rclpy.spin_once(self)
+        return srv_call.result().success
+
+    # TODO(shivesh): Wait for https://github.com/ros2/rclpy/issues/244
```
chapulina, line 274: That issue is closed, so I wonder if there's an alternative way of hooking to shutdown, or if there won't ever be a good way of implementing the bond functionality.

shiveshkhaitan: I could not figure out any alternative. Should we remove this functionality then?

chapulina: I left a comment on the issue: https://github.com/ros2/rclpy/issues/244#issuecomment-511832721 Let's see if someone comes back with an alternative.

For now, we can proceed with this PR leaving the code commented out.

chapulina:

gazebo_ros/scripts/spawn_entity.py
```diff
@@ -0,0 +1,333 @@
...
+                            If left empty or specified as "world", gazebo world frame is used')
+        parser.add_argument('-gazebo_namespace', type=str, default='',
+                            help='ROS namespace of gazebo offered ROS interfaces. \
+                            Default is without any namespace')
+        parser.add_argument('-robot_namespace', type=str, default=self.get_namespace(),
+                            help='change ROS namespace of gazebo-plugins')
+        parser.add_argument('-unpause', action='store_true',
+                            help='unpause physics after spawning entity')
+        parser.add_argument('-wait', type=str, metavar='ENTITY_NAME',
+                            help='Wait for entity to exist')
```
chapulina, line 77: We should comment out `wait`, since it doesn't work yet.

gazebo_ros/scripts/spawn_entity.py
```diff
@@ -0,0 +1,333 @@
...
+        parser = argparse.ArgumentParser(
+            description='Spawn an entity in gazebo. Gazebo must be started with gazebo_ros_init,\
+            gazebo_ros_factory and gazebo_ros_state for all functionalities to work')
+        source = parser.add_mutually_exclusive_group(required=True)
+        source.add_argument('-file', type=str, metavar='FILE_NAME',
+                            help='Load entity xml from file')
+        source.add_argument('-param', type=str, metavar='PARAM_NAME',
+                            help='Load entity xml from ROS parameter')
+        source.add_argument('-database', type=str, metavar='ENTITY_NAME',
+                            help='Load entity XML from specified entity in Gazebo Model Database')
```
chapulina: The documentation is not strictly correct. The model doesn't need to be on the Gazebo Model Database. It also works if it's a model in `GAZEBO_MODEL_PATH`.

Mind documenting that? Something like 

'Load entity XML from specified entity in GAZEBO_MODEL_PATH or Gazebo Model Database'

(I put the path before because Gazebo will look there before looking at the database.

gazebo_ros/scripts/spawn_entity.py
```diff
@@ -0,0 +1,333 @@
...
+            except IOError as e:
+                self.get_logger().error('Error reading file {}: {}'.format(self.args.file, e))
+                return 1
+            if entity_xml == '':
+                self.get_logger().error('Error: file %s is empty', self.args.file)
+                return 1
+        # Load entity XML from ROS param
+        elif self.args.param:
+            self.get_logger().info('Loading entity XML from ros parameter %s' % self.args.param)
+            entity_xml = self.get_parameter(self.args.param)
```
chapulina: This is not working for me, undeclared parameters attack again :confused:  I tried:

```
ros2 param set gazebo sdf "<?xml version=\"1.0\" ?><sdf version=\"1.5\"><light name=\"will_be_ignored\" type=\"directional\"></light></sdf>"
```

And then

```
ros2 run gazebo_ros spawn_entity.py -entity TheLight -param sdf
```

And I get this error:

```
rclpy.exceptions.ParameterNotDeclaredException: ('Invalid access to undeclared parameter(s)', 'sdf')
```

I think we'll need to allow undeclared parameters.

chapulina (changes requested):

gazebo_ros/scripts/spawn_entity.py
shiveshkhaitan: Even if we allow undeclared params, setting the parameter before starting the node would not work. And if we start the node first, then it crashes because the parameter is not passed yet. I don't find any use of the param. I think we can change this to a normal argument itself

chapulina: Ah I think you're right.

Getting the robot description from a param was probably the most common use case on ROS 1, because it was common to put the description on a global parameter `robot_description` that many tools used, like RViz. But since ROS2 [doesn't have](https://design.ros2.org/articles/ros_parameters.html) a central parameter server, we can't use that anymore.

It sounds like the new approach being currently used between `robot_state_publisher` and `rviz` is that the former publishes the model description in a latched topic (i.e. transient local) and the latter subscribes to it.

So how about deprecating the `-param` option and having a `-topic` one instead? Note that while the param on ROS 1 held the whole URDF file, the topic on ROS 2 holds the path to the URDF file.

* PR publishing robot description: https://github.com/ros2/robot_state_publisher/pull/9
* PR subscribing to robot description: https://github.com/ros2/rviz/pull/378

I'm also going to get the ball rolling so that this is documented on the [ROS 2 Migration guide](https://index.ros.org/doc/ros2/Contributing/Migration-Guide/).

gazebo_ros/scripts/spawn_entity.py
shiveshkhaitan: Model states should be working now as #968 is merged

gazebo_ros/launch/spawn_entity_demo.launch.py
```diff
@@ -0,0 +1,42 @@
...
+Spawns a model in the gazebo empty world
+"""
+
+from launch import LaunchDescription
+from launch.actions import ExecuteProcess
+from launch_ros.actions import Node
+
+
+def generate_launch_description():
+    gazebo = ExecuteProcess(
```
chapulina: Let's change this to use the launch files from #962 when we can. It will depend on which PR is merged first.

gazebo_ros/scripts/spawn_entity.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def __init__(self, args):
+        super().__init__('spawn_entity')
+        parser = argparse.ArgumentParser(
+            description='Spawn an entity in gazebo. Gazebo must be started with gazebo_ros_init,\
+            gazebo_ros_factory and gazebo_ros_state for all functionalities to work')
+        source = parser.add_mutually_exclusive_group(required=True)
+        source.add_argument('-file', type=str, metavar='FILE_NAME',
+                            help='Load entity xml from file')
+        source.add_argument('-param', type=str, metavar='PARAM_NAME',
```
chapulina: How about renaming this argument to `topic`, since this is no longer a param? The documentation and comments need to be updated too.

gazebo_ros/scripts/spawn_entity.py
```diff
@@ -0,0 +1,354 @@
...
+            entity_xml = ''
+
+            def entity_xml_cb(entity_xml_path):
+                nonlocal entity_xml
+                try:
+                    f = open(entity_xml_path.data, 'r')
+                    entity_xml = f.read()
+                except IOError as e:
+                    self.get_logger().error(
+                        'Error reading file {}: {}'.format(entity_xml_path.data, e))
```
chapulina: Sorry, my bad, I gave you bad guidance on my previous comment. :bowing_woman: 

I read https://github.com/ros2/rviz/pull/378 too quickly and thought that the standard now is to send the file path, but actually that PR reverted that behaviour to work like ROS 1. So in summary, we should treat the incoming data as the xml, not the path.

chapulina (changes requested):
I tried all arguments and everything is working as it should. The only remaining issue is `param -> topic`, sorry again :confused:

mogumbo:
Your spawn_model branch generally works for me. The only problem I have is if I add a `<gazebo>` tag containing a plugin, such as:
```
<gazebo>
    <plugin filename="libgazebo_ros_p3d.so" name="gazebo_ros_p3d">
      <frameName>map</frameName>
      <bodyName>base_link</bodyName>
      <topicName>ground_truth</topicName>
      <updateRate>30.0</updateRate>
    </plugin>
  </gazebo>
```
then gzserver always crashes:
```
[gzserver-1] Error [Element.cc:702] Missing element description for [ros]
[ERROR] [gzserver-1]: process has died [pid 23113, exit code -11, cmd 'gzserver -s libgazebo_ros_init.so -s libgazebo_ros_factory.so -s libgazebo_ros_state.so -e ode /home/me/workspace/mysim/install/mysim/share/mysim/worlds/myworld.world'].
```
Empty `<gazebo>` tags or ones containing materials or lights don't cause problems. Have you encountered anything like that?

shiveshkhaitan:
> [gzserver-1] Error [Element.cc:702] Missing element description for [ros]

This seems to be a problem with the sdf and not the plugin or `spawn_entity`.  Can you please post the problem with link to your complete sdf at http://answers.gazebosim.org/

chapulina (approved):

chapulina:
> Missing element description for [ros]

@mogumbo , I was able to reproduce the error you're getting and opened https://github.com/ros-simulation/gazebo_ros_pkgs/pull/983 to fix it. However, there are other issues with your plugin which will prevent it from working, so I recommend you check the [P3D migration guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-P3D).

mogumbo:
The migration guide helped a lot. Thanks for the tip.

PR #950, opened 2019-07-07
use c_str()

Karsten1987:
To avoid the warning and being able to compile with `-Werror`, this fix is needed.

Signed-off-by: Karsten Knese <karsten@openrobotics.org>


chapulina (approved):

chapulina:
FYI, the `ros2` branch is for Eloquent. This needs to be backported to `dashing`.

Karsten1987:
@jacobperron can we push for having the gazebo_ros_pkgs in the dashing github project?

chapulina:
> can we push for having the gazebo_ros_pkgs in the dashing github project?

I think the problem is that this repo is for both ROS 1 and 2 :confused: But since tickets needs to be manually assigned to boards maybe that's not an issue?

PR #951, opened 2019-07-08
[ros2] Fix test for diff drive

shiveshkhaitan:
The test should check for the linear velocity of the vehicle in the X direction wrt to the vehicle instead of linear velocity in Gazebo's X direction. This is how the odometry is also set.

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/fa09083dc854f275e60884d50b230fd57f4b0db5/gazebo_plugins/src/gazebo_ros_diff_drive.cpp#L574-L575

### DIFF
--- a/gazebo_plugins/test/test_gazebo_ros_diff_drive.cpp
+++ b/gazebo_plugins/test/test_gazebo_ros_diff_drive.cpp
@@ -89,9 +89,17 @@ TEST_P(GazeboRosDiffDriveTest, Publishing)
   // Wait for it to be processed
   int sleep{0};
   int maxSleep{300};
-  for (; sleep < maxSleep && (vehicle->WorldLinearVel().X() < 0.9 ||
+
+  auto yaw = static_cast<float>(vehicle->WorldPose().Rot().Yaw());
+  auto linear_vel = vehicle->WorldLinearVel();
+  double linear_vel_x = cosf(yaw) * linear_vel.X() + sinf(yaw) * linear_vel.Y();
+
+  for (; sleep < maxSleep && (linear_vel_x < 0.9 ||
     vehicle->WorldAngularVel().Z() < 0.09); ++sleep)
   {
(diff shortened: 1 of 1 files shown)


gazebo_plugins/test/test_gazebo_ros_diff_drive.cpp
```diff
@@ -89,9 +89,17 @@ TEST_P(GazeboRosDiffDriveTest, Publishing)
...
+  auto yaw = static_cast<float>(vehicle->WorldPose().Rot().Yaw());
+  auto linear_vel = vehicle->WorldLinearVel();
+  double linear_vel_x = cosf(yaw) * linear_vel.X() + sinf(yaw) * linear_vel.Y();
+
+  for (; sleep < maxSleep && (linear_vel_x < 0.9 ||
     vehicle->WorldAngularVel().Z() < 0.09); ++sleep)
   {
+    yaw = static_cast<float>(vehicle->WorldPose().Rot().Yaw());
+    linear_vel = vehicle->WorldLinearVel();
+    linear_vel_x = cosf(yaw) * linear_vel.X() + sinf(yaw) * linear_vel.Y();
```
chapulina, line 102: Are the values calculated in this loop used anywhere?

chapulina (approved):
Looks reasonable to me, let's try to run CI for it

chapulina:

gazebo_plugins/test/test_gazebo_ros_diff_drive.cpp
shiveshkhaitan: `linear_vel_x` is checked in every iteration of the loop in the condition

chapulina:
> let's try to run CI for it 

I think we need this to get CI working for the `dashing` branch: https://github.com/ros/rosdistro/pull/21722.

For now, I tested this locally and it works for me :+1:

PR #926, opened 2019-05-21
use latest dashing api

Karsten1987:
This PR addresses the latest changes in `rclcpp` and thus let's this package compile wo/ warnings.

a few points to iterate over:
* `rclcpp::SensorDataQoS()` was used for basically all plugins which represent a sensor
* Unused warnings were addressed for the conversions in `gazebo_ros`. 

@chapulina fyi

PR #953, opened 2019-07-11
[ros2] Fix tests on Dashing

chapulina:
Follow up to #926. I think we missed this test because we usually don't compile rendering tests. When building with colcon, this enables the tests:  `--cmake-args -DENABLE_DISPLAY_TESTS=true`

@shiveshkhaitan, mind taking a look? Thanks!

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera.cpp
@@ -82,10 +82,6 @@ GazeboRosCamera::GazeboRosCamera()
 GazeboRosCamera::~GazeboRosCamera()
 {
   impl_->image_pub_.shutdown();
-
-  // TODO(louise) Why does this hang for the 2nd camera plugin?
-  // Commenting it out here just pushes the problem somewhere else.
-  // impl_->ros_node_.reset();
 }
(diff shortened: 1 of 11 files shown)


shiveshkhaitan:
This looks fine. `test_gazebo_ros_camera_distortion` also requires similar change in qos api
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/22d782e56b456744ba47db9711c11651e3933434/gazebo_plugins/test/test_gazebo_ros_camera_distortion.cpp#L103
and two other places. Even this doesn't come up while compiling as the test is commented for now

chapulina:
Nice catch with the distortion tests. I went ahead and backported the fix from #937 and re-enabled the distortion tests, which should now be fixed.

chapulina:

chapulina:

chapulina:
@shiveshkhaitan , it looks like most tests are fixed. The Ackermann is still flaky though. Mind taking another look at this PR? We can leave Ackermann for later.

shiveshkhaitan:
LGTM. All tests (including `ackermann_drive`) drive passed for me 10 times in a row.

PR #954, opened 2019-07-11
use c_str() (#950)

chapulina:
Backport #950 to Dashing

FYI @Karsten1987


chapulina:
This is just a backport and CI is happy: merging.

PR #955, opened 2019-07-15
[ros2] Port joint pose trajectory to ROS2

shiveshkhaitan:
Port `gazebo_ros_joint_pose_trajectory` plugin to ROS2. 

Example:
```
<plugin name="gazebo_ros_joint_pose_trajectory" filename="libgazebo_ros_joint_pose_trajectory.so">
  <ros>
    <namespace>/demo</namespace>
    <argument>set_joint_trajectory:=set_trajectory_demo</argument>
  </ros>
  <update_rate>2</update_rate>
</plugin>
```
Migration guide at [ROS 2 Migration: Joint Pose Trajectory](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Joint-Pose-Trajectory)


JaehyunShim:
Hi @shiveshkhaitan @chapulina 

I tried to build source packages from this PR but it didn't work (In more detail, my laptop froze while building the packages) Could I ask when this PR will be tested and released? 

Thank you in advance,
Ryan

chapulina:
> Could I ask when this PR will be tested and released?

@rjshim , we're currently working on several pull requests for ROS 2. This one should be reviewed and merged in the coming weeks, and released in a Dashing sync after that.

If you need help building from source, you can ask for help at http://answers.gazebosim.org/ .

JaehyunShim:
@chapulina 
Thank you for the prompt response. 

@shiveshkhaitan 
Can I ask if it has been also tested in different development environments? I am asking this as the problem is occurring specifically while building the gazebo-plugin package,

Any advice will be appreciated! 
Ryan


Screenshot from 2019-07-23 20-12-32

chapulina:
> Can I ask if it has been also tested in different development environments?

We're mainly developing and running CI on Ubuntu Bionic, using both source builds and debian packages for the rest of ROS2. There are also users on OSX.

> the problem is occurring specifically while building the gazebo-plugin package

The screenshot you provided doesn't show any problems. Are you having problems with this branch specifically or do you also have trouble building the `dashing` branch? If the problem is not this branch, I ask you to move this discussion to a more appropriate place:

* If you found a **specific bug** in `gazebo_plugins`, please ticket a new issue on this repository with the exact issue.
* If you're having trouble getting a local workspace setup, please ask a question at http://answers.gazebosim.org/ .

In either case, it will be easier to help you if you provide information about your setup, what steps you've followed and what you've already tried to fix the issue. Have you checked your logs under your workspace's `log` folder for the exact errors you're having? Copy-pasting the error into the issue / question is better than screenshots because others can search through it.

rjshim:
@chapulina 

Thank you very much for the advice. I don't know the reason for the problem but I reinstalled Ubuntu and ROS and now building the packages succeeded. But it still often freezes when building that plugin package so I will take more time to figure out a solution for it and ask this questions at the link you attached.

Thank you,
Ryan

JaehyunShim:
@chapulina 

Thank you very much for the advice. I don't know the reason for the problem but I reinstalled Ubuntu and ROS and now building the packages succeeded. But it still often freezes when building that plugin package so I will take more time to figure out a solution for it and ask this question at the link you attached.

Thank you,
Ryan

chapulina:

gazebo_plugins/worlds/gazebo_ros_joint_pose_trajectory_demo.world
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0"?>
+<!--
+  Gazebo ROS joint pose trajectory plugin demo
+
+  Try for example:
+  ros2 topic pub /demo/set_trajectory_demo trajectory_msgs/msg/JointTrajectory
```
chapulina: This should make it easier to copy and paste the command on the terminal:

```suggestion
  ros2 topic pub /demo/set_trajectory_demo trajectory_msgs/msg/JointTrajectory \
```

gazebo_plugins/src/gazebo_ros_joint_pose_trajectory.cpp
```diff
@@ -0,0 +1,262 @@
...
+}
+
+void GazeboRosJointPoseTrajectoryPrivate::SetJointTrajectory(
+  trajectory_msgs::msg::JointTrajectory::SharedPtr msg)
+{
+  std::lock_guard<std::mutex> scoped_lock(lock_);
+
+  std::string reference_link_name = msg->header.frame_id;
+  // do this every time a new joint trajectory is supplied,
+  // use header.frame_id as the reference_link_name_
```
chapulina, line 209: I know this logic was copied from ROS 1, but I'll just point out that the reference links can get mixed up if messages with different `frame_id` are sent one after the other really quickly before the `OnUpdate` call had a chance to process the previous request.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_pose_trajectory.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <gazebo/common/Plugin.hh>
+
+#include <memory>
+
+namespace gazebo_plugins
+{
+class GazeboRosJointPoseTrajectoryPrivate;
+
+/// Set the trajectory of points to be followed by joints in simulation.
```
chapulina, line 26: We should document that only `positions` are being handled. The [message](http://docs.ros.org/melodic/api/trajectory_msgs/html/msg/JointTrajectoryPoint.html) also has velocities, accelerations and effort.

chapulina (changes requested):
Works for me! There are some things which could be improved on the logic, but I think that goes beyond the scope of this PR, so we don't need to address that now.

Ready to merge once we have tests for the new convert functions. Thanks!

gazebo_plugins/src/gazebo_ros_joint_pose_trajectory.cpp
shiveshkhaitan: Any suggestions for this?

chapulina: Actually I think it's ok, because we're also resetting `points_`.

chapulina:
I rebased with `dashing` and it looks good to me. Will merge when CI comes back happy. :+1:

chapulina (approved):
I rebased with `dashing` and it looks good to me. Will merge when CI comes back happy. :+1:

PR #959, opened 2019-07-17
Camera optical frame

MarshallRawson:
This is in response to VRX Issue#140 at https://bitbucket.org/osrf/vrx/issues/140/there-is-no-camera-optical-link . With this pull request Gazebo Ros Cameras (and only Gazebo Ros Camera) now published the transform the the optical frame.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
@@ -27,6 +27,7 @@
 #include <ros/ros.h>
 #include <ros/callback_queue.h>
 #include <ros/advertise_options.h>
+#include <tf2_ros/static_transform_broadcaster.h>
 
 // ros messages stuff
 #include <sensor_msgs/PointCloud.h>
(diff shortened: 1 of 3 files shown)


gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera_utils.h
```diff
@@ -139,8 +140,14 @@ namespace gazebo
     protected: common::Time last_info_update_time_;
 
     /// \brief ROS frame transform name to use in the image message header.
-    ///        This should typically match the link name the sensor is attached.
+    ///        This should typically match the link name the parent sensor + "_optical".
     protected: std::string frame_name_;
```
chapulina: It looks like the documentation for `frame_name_` and `frame_name_optical_` is mixed up

chapulina, line 144: It looks like the documentation for `frame_name_` and `frame_name_optical_` is mixed up

chapulina (commented):
Adding new member variables breaks ABI. Since this is only adding a static transform that has a hardcoded fixed value, I'm not sure if the ABI break would be worth it. This could as easily be published from outside the plugin.

PR #960, opened 2019-07-18
[ros2] Port vacuum gripper to ROS2

shiveshkhaitan:
Port `gazebo_ros_vacuum_gripper` plugin to ROS2. 

Example usage:
```
    <plugin name='vacuum_gripper' filename='libgazebo_ros_vacuum_gripper.so'>
      <ros>
        <namespace>/demo</namespace>
        <argument>switch:=custom_switch</argument>
        <argument>grasping:=custom_grasping</argument>
      </ros>

      <!-- Link associated with gripper -->
      <link_name>link</link_name>

      <!-- Max and min distance to attract entities -->
      <max_distance>10.0</max_distance>
      <min_distance>0.01</min_distance>

      <!-- Max force applied while attracting entities -->
      <max_force>2</max_force>

      <!-- List of entities to be not attracted by the gripper -->
      <fixed>ground_plane</fixed>
      <fixed>wall</fixed>

    </plugin>
```
Migration guide at [ROS 2 Migration: Vacuum Gripper](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Vacuum-Gripper)


chapulina:

gazebo_plugins/include/gazebo_plugins/gazebo_ros_vacuum_gripper.hpp
```diff
@@ -0,0 +1,83 @@
...
+        <argument>switch:=custom_switch</argument>
+        <argument>grasping:=custom_grasping</argument>
+      </ros>
+
+      <!-- Link associated with gripper -->
+      <link_name>link</link_name>
+
+      <!-- Max and min distance to attract entities -->
+      <max_distance>10.0</max_distance>
+      <min_distance>0.01</min_distance>
```
chapulina: I like that you exposed these variables that used to be hardcoded! But I think the name `min_distance` is a bit misleading since the models are still attracted if the distance is below that.

chapulina (changes requested):
I was testing the demo world, and it behaves quite strangely. This is just speeding up the simulation (on left panel, set `Physics -> real time update rate` to 0 to run as fas as possible). See how the gripper pulls the models, but then throws them away. And things start getting real fun as the gripper is moved:

vacuum_gripper

I think the problem comes from the ROS 1 plugin, so it's not introduced by this pull request. Sorry I didn't know of those problems beforehand :confused: 

---

The ARIAC competition wrote their own vacuum gripper plugin to work around the limitations of the plugin in `gazebo_ros_pkgs`. They have a [pure gazebo plugin](https://bitbucket.org/osrf/ariac/src/master/osrf_gear/src/VacuumGripperPlugin.cc) and a [ROS wrapper](https://bitbucket.org/osrf/ariac/src/master/osrf_gear/src/ROSVacuumGripperPlugin.cc) around it. That plugin has been used on the competition successfully many times, so I think we can trust that implementation better than the one here.

How do you feel about porting that plugin to `gazebo_ros_pkgs` instead? We can keep the "pure gazebo + ROS wrapper" approach, similar to what you did for the multicamera. 

I think it's also valid for us to decline this PR, and mark the vacuum gripper as deprecated for now, since it doesn't really work. And in the future, if users come looking for it, someone else could port a better version like ARIAC's to work here.

shiveshkhaitan:
I have updated this plugin itself. Please see if it solves the problem.

chapulina:
> I have updated this plugin itself. Please see if it solves the problem.

Nothing moves in the demo world for me, does it move for you?

I saw you removed the min and max distances. I'm ok removing the min, but without the max, it means that the gripper will be applying tiny forces to objects really far away, which I'm not sure we want. Also, using `SetForce` instead of `AddForce` will probably cancel gravity and other forces.

shiveshkhaitan:
> Nothing moves in the demo world for me, does it move for you?

Did you try switching on the gripper 
`    ros2 service call /demo/switch_demo std_srvs/srv/SetBool "{data: True}"`

chapulina:
> Did you try switching on the gripper

Haha no I didn't. Sorry about that, I just needed some :coffee: . It's working for me! Looks more like a magnet though :smile: 

magnets

Did you give some thought to the max distance? Right now the demo gripper can pull spheres from any distance. I placed one of the spheres 100m away and it was gripped.

It would also be interesting to keep using `AddForce` if possible, so the gripper reacts differently according to other forces in the world.

chapulina (approved):
Works for me, thanks for fixing it! I [removed](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Vacuum-Gripper/_compare/e1ca6faa11fdef54dba72d3ced5851343ba17c82...4918adbf00149b0d530beda5a38ec947fac14ed3) `max_force` and `min_distance` from the wiki page:

PR #962, opened 2019-07-19
[ros2] Port gazebo launch scripts to ROS2

shiveshkhaitan:
Port `gazebo`, `gzserver` and `gzclient` scripts from ROS1 to launch files in ROS2.

Example usage:
`ros2 launch gazebo_ros gazebo.launch.py verbose:=true pause:=true Timer:=true`

`ros2 launch gazebo_ros gzserver.launch.py verbose:=true pause:=true`

`ros2 launch gazebo_ros gzclient.launch.py verbose:=true Timer:=true`

This depends on #925


chapulina:

gazebo_ros/launch/gzclient.launch.py
```diff
@@ -0,0 +1,90 @@
...
+        DeclareLaunchArgument('verbose', default_value='false',
+                              description='Set "true" to increase messages written to terminal'),
+        DeclareLaunchArgument('help', default_value='false',
+                              description='Set "true" to produce gzclient help message'),
+        DeclareLaunchArgument('Timer', default_value='false',
+                              description='Set "true" to start gzclient with Timer GUI plugin'),
+        DeclareLaunchArgument('Cessna', default_value='false',
+                              description='Set "true" to start gzclient with Cessna GUI plugin'),
+        DeclareLaunchArgument('Keyboard', default_value='false',
+                              description='Set "true" to start gzclient with Keyboard GUI plugin'),
```
chapulina: Instead of exposing these 3 GUI plugins as boolean options, it would be more flexible to let the user pass the full plugin name as a string. This way, they can also choose custom plugins. What do you think?

gazebo_ros/launch/gzserver.launch.py
```diff
@@ -0,0 +1,126 @@
...
+        plugin += ':'+environ['GAZEBO_PLUGIN_PATH']
+    if 'GAZEBO_RESOURCE_PATH' in environ:
+        media += ':'+environ['GAZEBO_RESOURCE_PATH']
+
+    env = {'GAZEBO_MODEL_PATH': model,
+           'GAZEBO_PLUGIN_PATH': plugin,
+           'GAZEBO_RESOURCE_PATH': media}
+
+    return LaunchDescription([
+        DeclareLaunchArgument('world', default_value='worlds/empty.world',
```
chapulina: How about leaving the default empty so Gazebo decides on the default value?

gazebo_ros/launch/gzclient.launch.py
```diff
@@ -0,0 +1,90 @@
...
+                 _plugin_command('Cessna'), _plugin_value('Cessna'),
+                 _plugin_command('Keyboard'), _plugin_value('Keyboard'),
+                 '-s', 'libgazebo_ros_init.so',
+                 '-s', 'libgazebo_ros_factory.so',
+                 # Wait for (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/941)
+                 # '-s', 'libgazebo_ros_effort.so',
+                 ],
+            output='screen',
+            additional_env=env,
+        )
```
chapulina, line 74: We could also expose `gdb` and `valgrind` for `gzclient`

gazebo_ros/launch/gzserver.launch.py
```diff
@@ -0,0 +1,126 @@
...
+                 _command('record'),
+                 _arg_command('record_encoding'), LaunchConfiguration('record_encoding'),
+                 _arg_command('record_path'), LaunchConfiguration('record_path'),
+                 _arg_command('record_period'), LaunchConfiguration('record_period'),
+                 _arg_command('record_filter'), LaunchConfiguration('record_filter'),
+                 _arg_command('seed'), LaunchConfiguration('seed'),
+                 _arg_command('iters'), LaunchConfiguration('iters'),
+                 _command('minimal_comms'),
+                 '-s', 'libgazebo_ros_init.so',
+                 '-s', 'libgazebo_ros_factory.so',
```
chapulina: You could expose a parameter like `server-plugin` for the user to pass extra system plugins

chapulina (changes requested):
#925 is merged and backported to `dashing`. I rebased this branch so it can use `GazeboRosPaths`.

It's working for me, I just have some suggestions. We also need a migration guide.

gazebo_ros/launch/gzclient.launch.py
shiveshkhaitan: The problem is that one argument only takes one value.  Users cannot assign a list of plugins to an argument

gazebo_ros/launch/gzserver.launch.py
shiveshkhaitan: The problem with leaving it empty is that after an empty string, `gazebo` tries to load a world named '' and it falls back to default but shows a warning world name '' not found.

gazebo_ros/launch/gzserver.launch.py
shiveshkhaitan: Same problem as in `gzclient` that users cannot assign a list to an argument

gazebo_ros/launch/gzclient.launch.py
chapulina: Together with @sloretz, we found out about the `shell=True` option, which lets us pass the whole command as a single string. Using that, I have some suggestions on https://github.com/ros-simulation/gazebo_ros_pkgs/commit/57216468803f5f9eba3cd585c50c9e9b75c537a6 .

The last thing I'd like to add would be `gui` and `server` arguments to `gazebo.launch.py` so that users can conditionally include `gzclient` and `gzserver`, i.e.:

`ros2 launch gazebo_ros gazebo.launch.py gui:=false`


I found documentation implying that `IncludeLaunchDescription` can be conditionally included but I didn't figure out exactly how.

* https://answers.ros.org/question/314903/ros2-optionally-launch-node-with-no-arguments/
* https://github.com/ros2/launch/blob/master/launch/doc/source/architecture.rst#base-action
* https://github.com/ros2/launch/pull/121/files

chapulina: It turns out conditional includes are not possible yet: https://github.com/ros2/launch/issues/303. We could add a note to the code pointing to that issue.

This PR looks good to me once the changes to use `shell=True` are in.

chapulina (approved):
Merging! We just need a migration guide

gazebo_ros/launch/gzserver.launch.py
```diff
@@ -0,0 +1,150 @@
...
+    return LaunchDescription([
+        DeclareLaunchArgument('world', default_value='',
+                              description='Specify world file name'),
+        DeclareLaunchArgument('version', default_value='false',
+                              description='Set "true" to output version information.'),
+        DeclareLaunchArgument('verbose', default_value='false',
+                              description='Set "true" to increase messages written to terminal.'),
+        DeclareLaunchArgument('help', default_value='false',
+                              description='Set "true" to produce gzserver help message.'),
+        DeclareLaunchArgument('pause', default_value='false',
```
jacobperron, line 51: Note, in ROS 1 the launch argument name is "pause**d**". I'm not sure why it changed in ROS 2, but maybe we can document this difference somewhere on a migration page.

jacobperron: I'm referring to the [empty_world.launch](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/0606994ee2cd42afed75b32c494cacc6fe8adf16/gazebo_ros/launch/empty_world.launch) file, but I guess this isn't really an exact equivalent to `gzserver.launch.py`, they just happen to both offer launch args for pausing.

PR #967, opened 2019-07-26
[ros2] Backport depth camera to dashing

shiveshkhaitan:
Back port depth camera and openni kinnect to dashing https://github.com/ros-simulation/gazebo_ros_pkgs/pull/932


chapulina (approved):
LGTM! The usage remains exactly the same as on `eloquent`, great job!

I just moved the depth camera header from `include` to `src` so it isn't installed: 8719c30. This way, we don't need to go through a deprecation cycle for it (unless someone comes asking us to install it in the future).

Will merge once we get clean CI.

PR #972, opened 2019-08-06
Address reviews on #868

shiveshkhaitan:
This addresses the requested changes on #868.


gazebo_msgs/srv/GetModelProperties.srv
```diff
@@ -1,9 +1,9 @@
 string model_name                    # name of Gazebo Model
```
chapulina, line 1: A couple of problems here:

* This file is needed for the `ros1_bridge`. But this could be fixed adding the remapping rules [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/dashing/gazebo_msgs/gazebo_msgs_mapping_rules.yaml).
* This message has already been released on Dashing, so we'd be breaking API if we changed it. Even if we added new fields side-by-side with the old ones, we'd keep API compatible, but break ABI.

I think the best solution is to copy the message and rename it. Unfortunately anything other than `GetModelProperties` would break the pattern with the other services, but it's either that or leaving the current fields. I'll leave it up to you.

gazebo_ros/include/gazebo_ros/gazebo_ros_properties.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <gazebo/common/Plugin.hh>
+
+#include <memory>
+
+namespace gazebo_ros
+{
+
+class GazeboRosPropertiesPrivate;
+
+class GazeboRosProperties : public gazebo::WorldPlugin
```
chapulina, line 56: Mind documenting the services offered by this plugin?

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,457 @@
...
+      break;
+    }
+  }
+
+  if (!joint) {
+    _res->success = false;
+    _res->status_message = "GetJointProperties: joint not found";
+  } else {
+    /// @todo: FIXME
+    _res->type = _res->REVOLUTE;
```
chapulina: I know this was like this on ROS 1, but we could fix it now :innocent: 

We could also fix the fields below to correctly get information for both axes, in case the joint has 2.

gazebo_ros/src/gazebo_ros_properties.cpp
```diff
@@ -0,0 +1,457 @@
...
+    boost::dynamic_pointer_cast<gazebo::physics::Link>(world_->EntityByName(_req->link_name));
+  if (!link) {
+    _res->success = false;
+    _res->status_message =
+      "SetLinkProperties: link not found, did you forget to scope the link by model name?";
+  } else {
+    gazebo::physics::InertialPtr mass = link->GetInertial();
+    // @todo: FIXME: add inertia matrix rotation to Gazebo
+    // mass.SetInertiaRotation(ignition::math::Quaternion(_req->com.orientation.w,
+    // _res->com.orientation.x,_req->com.orientation.y _req->com.orientation.z));
```
chapulina: There's a [SetCoG](https://bitbucket.org/osrf/gazebo/src/9e18a31cda6328c1d8d4ec08b9d1b6cfcdb93585/gazebo/physics/Inertial.hh#lines-115) which takes position and orientation. Do you know off the top of your head if that function is what we want here, @scpeters ?

chapulina (changes requested):
I pushed some changes on 5293a5a, let me know if anything doesn't look good to you.

gazebo_ros/src/gazebo_ros_properties.cpp
scpeters: there's an overload that takes an `ignition::math::Pose3d`:

* https://bitbucket.org/osrf/gazebo/src/2b59492f96fcf121b33f290fd5cf0b70b42f4e22/gazebo/physics/Inertial.hh#lines-126:128

the one at line 115 takes roll, pitch, yaw, but it's easier to construct a Pose with the position xyz and quaternion wxyz

gazebo_msgs/srv/GetModelProperties.srv
shiveshkhaitan: Can we change the message for `eloquent`? We will keep the same message for dashing.

gazebo_ros/src/gazebo_ros_properties.cpp
shiveshkhaitan: I guess you fixed this in 5293a5a

gazebo_msgs/srv/GetModelProperties.srv
chapulina: We still need the bridge to work for Eloquent. But here's what we can do:

* On Eloquent, have the fields side-by-side, accept both, but print warnings if the user is setting the old ones
* We probably don't want to change this for F turtle, since people will still be using the bridge for a long time. So we can keep this deprecated state for a few distros.

What do you think?

gazebo_ros/src/gazebo_ros_properties.cpp
chapulina: ha, I did, and forgot to delete the comment :grin:

gazebo_msgs/srv/GetModelProperties.srv
shiveshkhaitan: SGTM. I have reverted the changes to message for now aac7178

gazebo_ros/src/gazebo_ros_properties.cpp
chapulina: Setting the whole CoG pose on f0e9796

chapulina (approved):
Thank you @nzlz for starting the migration and @shiveshkhaitan for wrapping it up. LGTM, just waiting on CI.

PR #976, opened 2019-08-10
diff_drive: handle negative velocities in encoder mode, related to #7…

phcapde:
diff_drive: handle negative velocities in encoder mode, related to #717 and #719


chapulina (commented):
This looks reasonable to me. I just worry about merging this into a released distribution because downstream users running into this problem may already be compensating from the reversed signal in their code. If we release this fix, they'd be broken when updating without much indication of why.

PR #978, opened 2019-08-14
[ros2] Adding option to select the frame where the force will be applied

alexfneves:
I ended up doing this change for the simulation my project, and felt that it's a valid option when using the force plugin.

A new parameter was added on the plugin with the options world and body frame.
Internally the AddRelativeForce() and torque functions are used instead of the AddForce() when the body option is selected.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_force.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_force.hpp
@@ -69,10 +69,6 @@ protected:
   /// Optional callback to be called at every simulation iteration.
   virtual void OnUpdate();
 
-  /// Optional callback to be called at every simulation iteration when the force
-  /// is applied on the body frame.
-  virtual void OnUpdateRelative();
-
 private:
   /// Callback when a ROS Wrench message is received
   /// \param[in] msg The Incoming ROS message representing the new force to
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_force.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_force.hpp
@@ -46,6 +46,9 @@ class GazeboRosForcePrivate;
       <!-- Name of link within model which will receive the force -->
       <link_name>link</link_name>
 
+      <!-- Frame where the force/torque will be applied (options: world; body)-->
+      <force_frame>link</force_frame>
+
     </plugin>
   \endcode
 */
@@ -65,6 +68,9 @@ protected:
 
   /// Optional callback to be called at every simulation iteration.
   virtual void OnUpdate();
+
+  /// Optional callback to be called at every simulation iteration when the force is applied on the body frame.
+  virtual void OnUpdateRelative();
 
 private:
   /// Callback when a ROS Wrench message is received
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_force.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_force.hpp
@@ -69,7 +69,8 @@ protected:
   /// Optional callback to be called at every simulation iteration.
   virtual void OnUpdate();
 
-  /// Optional callback to be called at every simulation iteration when the force is applied on the body frame.
+  /// Optional callback to be called at every simulation iteration when the force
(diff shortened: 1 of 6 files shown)


gazebo_plugins/src/gazebo_ros_force.cpp
```diff
@@ -72,6 +75,22 @@ void GazeboRosForce::Load(gazebo::physics::ModelPtr model, sdf::ElementPtr sdf)
     return;
   }
 
+  // Frame 
+  if (!sdf->HasElement("force_frame")) {
+    RCLCPP_ERROR(logger, "Force plugin missing <force_frame>, cannot proceed");
```
chapulina: This would break backwards compatibility with the current plugin. We should keep the current behaviour if `force_frame` is not defined, i.e. world.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_force.hpp
```diff
@@ -46,6 +46,9 @@ class GazeboRosForcePrivate;
       <!-- Name of link within model which will receive the force -->
       <link_name>link</link_name>
 
+      <!-- Frame where the force/torque will be applied (options: world; body)-->
```
chapulina: `body` -> `link`

gazebo_plugins/src/gazebo_ros_force.cpp
```diff
@@ -43,6 +43,9 @@ class GazeboRosForcePrivate
 
   // Pointer to the update event connection
   gazebo::event::ConnectionPtr update_connection_;
+
+  /// Indicates that the force should be applied on the world frame instead of the body frame
+  bool force_world_frame_;
```
chapulina: This variable is only used within the `Load` function, so we don't need to store it as a member variable.

gazebo_plugins/worlds/gazebo_ros_force_demo.world
```diff
@@ -2,6 +2,11 @@
 <!--
   Gazebo ROS force plugin demo
 
+  If you want the force to be applied on the world frame, use
+    <force_frame>world</force_frame>
+  If you want the force to be applied on the body frame, use
+    <force_frame>body</force_frame>
```
chapulina: `body`->`link`

chapulina (changes requested):
Thank you for the PR. I just have a couple of comments regarding keeping backwards compatibility. It would also be good to change `body` to `link`, since body is not a well defined term within Gazebo or SDF/URDF.

A nice bonus would be to add a test case which uses the link frame, making sure the link is not aligned with the world in the test.

alexfneves:
I did the changes you asked me to do. I also changed the force test little bit so that it would do the normal test with force on the world frame and a new with the force on the link frame.

alexfneves:
I think the test failing is on the GazeboRosAckermannDriveTest, which has nothing to do with the code I wrote. Is the last commit alright then?

gazebo_plugins/include/gazebo_plugins/gazebo_ros_force.hpp
```diff
@@ -66,6 +69,10 @@ class GazeboRosForce : public gazebo::ModelPlugin
   /// Optional callback to be called at every simulation iteration.
   virtual void OnUpdate();
 
+  /// Optional callback to be called at every simulation iteration when the force
+  /// is applied on the body frame.
+  virtual void OnUpdateRelative();
```
chapulina: Sorry @alexfneves , I just realized that this can't go into dashing because adding `virtual` functions is not guaranteed not to break ABI. Can you change the implementation to use `OnUpdate` on both cases? I think you may need to bring back the `force_on_world_frame_` member variable.

chapulina (changes requested):

chapulina (approved):
LGTM, thank you again for the contribution!

I pushed some changes in 95fbd6c:

* remove some remaining `body`
* made the warning into info, because it's not wrong to leave out the new element
* added another box to the demo so it's easier to test both frames at the same time

I also added a [migration page](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Force) for the plugin, because it didn't exist.

alexfneves:
Fair enough, thanks!

PR #981, opened 2019-08-17
[ros2] CUDA accelerate depth camera

shiveshkhaitan:
GPU accelerated construction of pointcloud from rgb and depth images in `gazebo_ros_camera`.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/fill_depth.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/fill_depth.hpp
@@ -1,0 +1,53 @@
+// Copyright 2019 Open Source Robotics Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GAZEBO_PLUGINS__FILL_DEPTH_HPP_
+#define GAZEBO_PLUGINS__FILL_DEPTH_HPP_
+
+#include <cuda.h>
+
+class FillDepth
+{
+public:
+  struct parameters
+  {
+    int height;
+    int width;
+    int size;
+    double fl;
+    double min_depth;
+    double max_depth;
+    float infinity;
+  } params;
+
+  FillDepth();
+
+  ~FillDepth();
+
+  void initialize(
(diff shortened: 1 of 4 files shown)


iche033:
what's the performance gain is with cuda acceleration? Just wondering if you have done some profiling to see what the speedup is?

shiveshkhaitan:
It was 10Hz vs 50Hz

gazebo_plugins/include/gazebo_plugins/fill_depth.hpp
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GAZEBO_PLUGINS__FILL_DEPTH_HPP_
+#define GAZEBO_PLUGINS__FILL_DEPTH_HPP_
+
+#include <cuda.h>
+
+class FillDepth
```
chapulina, line 22: Could you document everything in this header? This is especially important since it will be installed.

gazebo_plugins/include/gazebo_plugins/fill_depth.hpp
```diff
@@ -0,0 +1,53 @@
...
+  FillDepth();
+
+  ~FillDepth();
+
+  void initialize(
+    unsigned int height, unsigned int width, double fl, double min_depth, double max_depth);
+
+  void compute(const float * image_depth, uint8_t * image_rgb, float * depth, float * cloud);
+
+private:
```
chapulina, line 53: Can we use the PIMPL pattern here and leave only the public functions in the header?

chapulina (commented):
I just have a couple of comments for now. I haven't been able to test this with CUDA yet, but I verified that it doesn't break users who don't have CUDA.

It would also be nice to have a concrete benchmarking example. @shiveshkhaitan, how large were the point clouds you were publishing to get the improvement from 10 Hz to 50 Hz? Did you see improvements for both the point cloud and the depth images?

shiveshkhaitan:
Sorry for the delay. 

I tested the performance in the [demo file](https://github.com/shiveshkhaitan/gazebo_ros_pkgs/blob/depth_camera_gpu/gazebo_plugins/worlds/gazebo_ros_depth_camera_demo.world) for `gazebo_ros_depth_camera`.
-  The size of the image is 480 x 640.
- Yes the performance improvement was for both, pointcloud and depth images
- It was tested on Nvidia GeForce GTX 1070

ruffsl:
> It would also be nice to have a concrete benchmarking example.

@shiveshkhaitan where you able to add a benchmarking test so we can track future performance.

shiveshkhaitan:
@ruffsl https://github.com/ros-simulation/gazebo_ros_pkgs/pull/981#issuecomment-549015029 this is the only test I had done. I am not sure what further tests and benchmarks would be helpful. I am willing to perform more tests and benchmarks if required.

PR #982, opened 2019-08-19
[ros2] Adding GPS plugin

alexfneves:
I created the GPS sensor plugin that outputs a NavSatFix message in ros2. I used the IMU plugin as reference for the implementation.

I also added a test. First the box position acquired, then its position is set on the world (positive x, y and z), then check if the latitude, longitude and altitude are positive (positive East, North and Up).

Please take a look if it seems fine.

I have some notes:
1. P.S. Gazebo seems to be outputting ENU (the default world orientation) wrong. My take is that the C++ code shouldn't change, what should be fixed is the Gazebo code. I added a 180 deg on the heading on the .world file to make it output what it should. Later I'll take a look at the gazebo code to check the reason for this behavior.

2. The SDF has <position_sensing> and <velocity_sensing> tags, but I only implemented the <position_sensing> because the GpsSensor in Gazebo API for Gazebo 9.0 doesn't have the functions VelocityEast(), VelocityNorth() and VelocityUp(), which are available in Gazebo 9.1 and forth.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gps_sensor.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gps_sensor.hpp
@@ -1,4 +1,4 @@
-// Copyright 2018 Open Source Robotics Foundation, Inc.
+// Copyright 2019 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gps_sensor.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_gps_sensor.hpp
@@ -1,0 +1,66 @@
+// Copyright 2018 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+#ifndef GAZEBO_PLUGINS__GAZEBO_ROS_GPS_SENSOR_HPP_
+#define GAZEBO_PLUGINS__GAZEBO_ROS_GPS_SENSOR_HPP_
+
+#include <gazebo/common/Plugin.hh>
+#include <gazebo/sensors/GpsSensor.hh>
+#include <gazebo/common/Events.hh>
+
+#include <memory>
+
+namespace gazebo_plugins
+{
+
+class GazeboRosGpsSensorPrivate;
+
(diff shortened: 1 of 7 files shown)


alexfneves:
The test that is failing is the GazeboRosAckermannDriveTest Publishing, which is not related to the GPS.
http://build.ros2.org/job/Dpr__gazebo_ros_pkgs__ubuntu_bionic_amd64/105/testReport/

gazebo_plugins/include/gazebo_plugins/gazebo_ros_gps_sensor.hpp
```diff
@@ -0,0 +1,66 @@
+// Copyright 2018 Open Source Robotics Foundation, Inc.
```
chapulina: Nitpick: 2019. Also on the other files.

gazebo_plugins/src/gazebo_ros_gps_sensor.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+}
+
+void GazeboRosGpsSensor::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _sdf)
+{
+  impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
+
+  impl_->sensor_ = std::dynamic_pointer_cast<gazebo::sensors::GpsSensor>(_sensor);
+  if (!impl_->sensor_) {
+    RCLCPP_ERROR(impl_->ros_node_->get_logger(), "Parent is not an gps sensor. Exiting.");
```
chapulina: Nitpick: "an gps" -> "a GPS"

gazebo_plugins/test/worlds/gazebo_ros_gps_sensor.world
```diff
@@ -0,0 +1,62 @@
...
+    <include>
+      <uri>model://ground_plane</uri>
+    </include>
+    <include>
+      <uri>model://sun</uri>
+    </include>
+    <spherical_coordinates>
+        <!-- currently gazebo has a bug: instead of outputing lat, long, altitude in ENU as the
+        default configurations, it's outputting (-E)(-N)U, therefore we rotate the default frame
+        180 so that it would go back to 180 -->
```
chapulina: Mind adding a link to this issue here so we don't lose track of it?

https://bitbucket.org/osrf/gazebo/issues/2022/

chapulina (changes requested):
Thanks for the PR! It's looking good to me! Besides the nitpick comments, all I'd like to ask is that you create a demo world in the `gazebo_plugins/worlds` folder. It can be similar to the test world, but with a few instructions on the top. Thanks!

chapulina (approved):
Thanks for iterating, LGTM! :+1:

gazebo_plugins/test/worlds/gazebo_ros_gps_sensor.world
alexfneves: I didn't search about it before. Good to know there is already an issue on this!

alexfneves: Sorry @chapulina, at the time I was making the changes I didn't understand that you wanted me to add this link https://bitbucket.org/osrf/gazebo/issues/2022/ on the .world comment. Now that it has been merged into master, maybe you can add it directly.

chapulina: No worries, it can be added later :wink:

PR #986, opened 2019-08-25
[ros2] Uncommenting bond option on spawn_entity (wait Ctrl+C then remove entity)

alexfneves:
This is connected to this issue https://github.com/ros2/rclpy/issues/244. This feature is almost implemented by @shiveshkhaitan, it is just waiting for a shutdown callback to be implemented in rclpy.

Instead of waiting for that, we can use the try/except to get the SIGINT signal, then delete the entity. I only did minor changes his code. Is this approach alright?


chapulina (approved):
Nice solution! I tweaked some of the console messages on e6a00bdba344cdd463cf0a6b2f5223d492052cd5, but it works great for me!

PR #987, opened 2019-08-28
Reorder fields initialization to match initialization order in .h file

aeneev:
This change fixes -Wreorder warnings. More on -Wreorder:
https://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder


chapulina (approved):
Thanks!

PR #975, opened 2019-08-08
[Windows][melodic-devel] more Windows build break fix

seanyen:
More Windows build break fix.

PR #993, opened 2019-09-11
[Windows][scripts] Search plugin DLLs from bin folder

seanyen:
A recent change https://github.com/ros-simulation/gazebo_ros_pkgs/pull/975 (well...made by me) started placing the plugin shared libraries to `bin` folder, but I forgot to update the launch scripts to include `bin` as the search folder. (Unfortunately, I didn't find a way to do it from `catkin_find` CLI tool so I am using `catkin.find_in_workspaces` instead.)


chapulina (approved):
Seems reasonable to me :+1:

PR #994, opened 2019-09-11
Remove ROS-specific arguments before passing to argparse

jacobperron:
This resolves argparse errors when trying to launch the spawn_entity script as a ROS node.

For example, a launch file like the following wouldn't work without this change:

    <launch>
      <arg name="model_urdf" default="$(find-pkg-share mymodels)/urdf/ball.urdf" />
      <node
        pkg="gazebo_ros"
        exec="spawn_entity.py"
        name="spawner"
        args="-entity foo -file /path/to/my/model/foo.urdf" />
    </launch>

### DIFF
--- a/gazebo_ros/scripts/spawn_entity.py
+++ b/gazebo_ros/scripts/spawn_entity.py
@@ -340,7 +340,8 @@ def quaternion_from_euler(roll, pitch, yaw):
 
 def main(args=sys.argv):
     rclpy.init(args=args)
-    spawn_entity_node = SpawnEntityNode(args)
+    args_without_ros = rclpy.utilities.remove_ros_args(args)
+    spawn_entity_node = SpawnEntityNode(args_without_ros)
     spawn_entity_node.get_logger().info('Spawn Entity started')
     exit_code = spawn_entity_node.run()
     sys.exit(exit_code)


jacobperron:
I'm not sure what's causing the test failure, my change looks innocent enough :innocent:

chapulina (approved):
The ackermann test has been flaky.

This looks good to me :+1:

mogumbo:
Unfortunately, I still can't run spawn_entity.py as a node. If I print out args and args_without_ros, they are the same:

[spawn_entity.py-5]             args =  ['/home/mogumbo/workspace/install/gazebo_ros/lib/gazebo_ros/spawn_entity.py', '-entity', 'rp1a', '-file', '/tmp/rp1a_robot.urdf', '-package_to_model', '--ros-args']
[spawn_entity.py-5] args_without_ros =  ['/home/mogumbo/workspace/install/gazebo_ros/lib/gazebo_ros/spawn_entity.py', '-entity', 'rp1a', '-file', '/tmp/rp1a_robot.urdf', '-package_to_model', '--ros-args']

Is something broken in rclpy.utilities.remove_ros_args?

jacobperron:
@mogumbo :thinking: I've successfully launched spawn_entity.py as a node building the gazebo_ros_pkgs `dashing` branch against both ROS 2 Dashing binaries as well as the latest source. Double check that your source checkout of gazebo_ros_pkgs is up-to-date.

mogumbo:
Thanks for testing. I have up-to-date gazebo_ros_pkgs:dashing as well. I also tried ros-dashing-rclpy from the repo and building rclpy:dashing in my workspace. Just rebuilt my whole workspace and the problem remains. I have no idea why it's behaving differently here.

jacobperron:
Just to clarify, avoid mixing `rclpy:dashing` with latest versions of `launch_ros` or other packages.

`--ros-args` is new and should not appear anywhere in Dashing (ie. the remove args utility in Dashing doesn't know about `--ros-args`). Since we don't have XML launch files in Dashing, I would expect a launch file like this example to work (replacing the entity and path to URDF with valid values):

``` python
from launch import LaunchDescription
from launch_ros.actions import Node


def generate_launch_description():
    return LaunchDescription([
            Node(
                package='gazebo_ros',
                node_executable='spawn_entity.py',
                arguments=['-entity', 'foo', '-file', '/full/path/to/model.urdf']
            ),
        ])
```

I'm guessing you're experiencing some workspace shenanigans.

mogumbo:
I don't totally understand your comment. I have ros-dashing-rclpy installed. Should I expect this to cause problems?

I have rebuilt my workspace from scratch and recompiled and I still have a --ros-args error when I launch.

jacobperron:
> I don't totally understand your comment. I have ros-dashing-rclpy installed. Should I expect this to cause problems?

It shouldn't cause problems if you're rebuilding rclpy in your workspace. To be sure, I would avoid sourcing `/opt/ros/dashing/setup.*` before you build your workspace.

mogumbo:
I don't understand. How can you build a ROS2 workspace without first sourcing /opt/ros/dashing/setup.* ? The build fails immediately.

jacobperron:
You can build everything from source. Here's [instructions for building the latest from source](https://index.ros.org/doc/ros2/Installation/Eloquent/Linux-Development-Setup/#linux-latest).

PR #999, opened 2019-09-16
[Windows][ros2] fix multi_camera_plugin on windows

jnoyola:
Another update so the latest changes are compatible with Windows. These changes should have no effect on other systems.

- Link `multi_camera_plugin` with the `gazebo_sensors` import library.
- Set `BUILDING_DLL` to correctly [assign the `GAZEBO_VISIBLE` attribute for export](https://bitbucket.org/osrf/gazebo/src/c1acd8bfc1bdd390380bdb33025de4bde0c9f458/gazebo/util/system.hh#lines-40)
  - Note this is also [used by Gazebo](https://bitbucket.org/osrf/gazebo/src/c1acd8bfc1bdd390380bdb33025de4bde0c9f458/plugins/CMakeLists.txt#lines-2) when building the built-in plugins

### DIFF
--- a/gazebo_plugins/CMakeLists.txt
+++ b/gazebo_plugins/CMakeLists.txt
@@ -97,6 +97,15 @@ ament_export_libraries(gazebo_ros_template)
 add_library(multi_camera_plugin SHARED
   src/multi_camera_plugin.cpp
 )
+ament_target_dependencies(multi_camera_plugin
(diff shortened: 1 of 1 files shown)


chapulina (approved):
Thanks for the PR. CI is happy and the changes look reasonable.

PR #1000, opened 2019-09-19
[ros2] Add support for the orientation_reference_frame SDF option for IMU sensors

jcmonteiro:
It is now possible to change the IMU reference frame instead of always acquiring data with respect to the initial IMU pose.

I wrote unit tests for all supported (based on the current SDF spec) use cases.

Besides, it looks like this feature should be added to the Gazebo IMU sensor class, but I found it easier and faster to do it here.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_imu_sensor.cpp
+++ b/gazebo_plugins/src/gazebo_ros_imu_sensor.cpp
@@ -13,8 +13,11 @@
 // limitations under the License.
 
 
-#include <builtin_interfaces/msg/time.hpp>
 #include <gazebo_plugins/gazebo_ros_imu_sensor.hpp>
+
+#include <gazebo/physics/World.hh>
+
+#include <builtin_interfaces/msg/time.hpp>
 #include <gazebo_ros/conversions/builtin_interfaces.hpp>
 #include <gazebo_ros/conversions/geometry_msgs.hpp>
 #include <gazebo_ros/node.hpp>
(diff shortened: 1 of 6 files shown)


chapulina:
Thanks for the PR!

[Here](https://bitbucket.org/osrf/gazebo/branches/compare/issue_1959_gazebo7%0Dgazebo7#chg-gazebo/sensors/ImuSensor.cc)'s an initial implementation for Gazebo and [here](https://bitbucket.org/osrf/gazebo/issues/1959)'s an issue tracking it.

As you mentioned, ideally this would be handled on the Gazebo side. And if this ever gets implemented there, some assumptions on this PR should be changed.

I'm curious to know why you "found it easier and faster to do it here."

Finally, note the test failures on CI, and check the [contribution guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-contribution-guide) for other tips.

jcmonteiro:
> As you mentioned, ideally this would be handled on the Gazebo side. And if this ever gets implemented there, some assumptions on this PR should be changed.

I believe the changes added to this PR would become obsolete as soon as they implement them on the Gazebo side. I might submit them there if I have the time.

> I'm curious to know why you "found it easier and faster to do it here."

Mainly because recompiling Gazebo and installing it from source for everybody in my workplace is kind of a dealbreaker. Also, changes here tend to reach upstream really fast.

> Finally, note the test failures on CI, and check the [contribution guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-contribution-guide) for other tips.

My bad... I did solve these issues but forgot to force push them before making the pull request. I will do this tomorrow morning.

chapulina:
> Mainly because recompiling Gazebo and installing it from source for everybody in my workplace is kind of a dealbreaker.

Thank you for the feedback, that's helpful. I think there's a misconception there though, because there wouldn't be a need to build from source. We periodically release minor versions of Gazebo 9 which they could use.

> Also, changes here tend to reach upstream really fast.

I think that's also a misconception. We did a recent push to port `gazebo_ros_pkgs` to ROS 2, but as most of the work is done, the level of maintenance of this package will be closer to that of Gazebo's. :wink: 

---

That all said, I'd really recommend you target this at Gazebo :grimacing:

jcmonteiro:
I mean compile from source while the changes do not reach the upstream. But you're right, this feature should really be targeted at Gazebo.

I might have the time to do it in a couple of weeks. Should this PR be considered anyway (after I force push to fix the CI errors, naturally, and assuming all changes are in fact correct)?

chapulina:
> Should this PR be considered anyway

How about waiting for it to get into Gazebo? Your team should be able to use your branch until then. My worry is that we'll need to come back and revert this once it's in Gazebo.

jcmonteiro:
Agreed. Feel free to close the PR then. I'll implement the changes in Gazebo if I have the time.

chapulina:
Thanks!

Closing as the changes will be ported to Gazebo.

PR #1008, opened 2019-10-17
diffdrive - apply acceleration until both speeds reach setpoint

scgroot:
Currently acceleration is no longer applied as soon as one of the wheel reach the target velocity. The other wheel is then also directly set to the set point, causing weird behavior.

By checking if both wheel speeds reached the target velocity this behavior is fixed.


scgroot:
I'll replace it with a pull request for the ros2 branch.

PR #1009, opened 2019-10-17
Apply acceleration until both left and right reach targetspeed

scgroot:
Currently acceleration is no longer applied as soon as one of the wheel reach the target velocity. The other wheel is then also directly set to the set point, causing weird behavior.

By checking if both wheel speeds reached the target velocity this behavior is fixed.


topalbu (approved):

chapulina (approved):
This looks reasonable to me. I'd usually ask for a test to make sure we don't break this by mistake again, but this pull request has been waiting long enough. Please consider opening a pull request with a test later :slightly_smiling_face:

PR #1010, opened 2019-10-17
Remove TODO in package.xml

jacobperron:
As I understand, listing execution dependencies is the correct way to form a metapackage.


chapulina (approved):
Thanks. Also saw this PR: https://github.com/ros2/ros2_documentation/pull/325/

Issue #991, opened 2019-09-06
[ros2] gazebo doesn't pick up the "out:=/scan" from turtlebot3 world file


mkhansenbot:
Sometime this week a change has merged somewhere in ROS2 source (using ros2.repos master file), that seems to have broken Gazebo ability to rename the turtlebot3 topics like "/scan" and "/imu" (see here: https://github.com/ROBOTIS-GIT/turtlebot3_simulations/blob/ros2/turtlebot3_gazebo/models/turtlebot3_waffle/model-1_4.sdf#L160)

Now we don't see those topics but we see default ones like "/turtlebot3_laserscan/out" instead of "/scan"

Maybe this is related to the "--ros-args" changes that merged upstream? Just a guess.

To reproduce:
1. build ros2 from source using ros2.repos file. 
2. build latest ros2 branch of gazebo_ros_pkgs
3. `git clone https://github.com/ROBOTIS-GIT/turtlebot3_simulations.git`
4. `export GAZEBO_MODEL_PATH=~/turtlebot3_simulations/turtlebot3_gazebo/models`
5. `gazebo --verbose -s libgazebo_ros_init.so ~/turtlebot3_simulations/turtlebot3_gazebo/worlds/turtlebot3_worlds/waffle.model`
6. Wait for simulation to load
7. In another window run `ros2 topic echo /scan` (nothing will echo)
8. `ros2 topic list` will not show the /scan topic, but will show /turtlebot3_laserscan/out instead
```

**Comment by ghost** [comment]
It looks like it is related to https://github.com/ros2/rcl/pull/491

**Comment by chapulina** [comment]
I think from Eloquent we may need to use this in the SDF: `<argument>-r ~/out:=scan</argument>`. If that works, we should update all our examples and docs.

**Comment by ghost** [comment]
That doesn't seem to be sufficient. I'm still seeing the output on /laserscan/out instead of scan

**Comment by chapulina** [comment]
Oh sorry, maybe this?

```.xml
<argument>--ros-args -r ~/out:=scan</argument>
```

I hope we don't need this :sweat: 

```.xml
<argument>--ros-args</argument>
<argument>-r</argument>
<argument>~/out:=scan</argument>
```

mkhansenbot:
Only that last method works :(

chapulina:
Ok, thanks for testing. I can add a new `<remapping>` element which prepends `--ros-args` and `-r` under the hood for convenience.

[...]

PR #1011, opened 2019-10-22
[ros2] Add remapping tag

mikaelarguedas:
Fix for https://github.com/ros-simulation/gazebo_ros_pkgs/issues/991

I originally added the `--ros-args` for any argument in the `ros` tag, but this breaks the current deprecated use of specifying remapping rules in the `argument` tag.

So this PR just adds `--ros-args` in front of the remapping rules and `-r` in front of each remap rule. This allows to get rid of warnings by modifying users sdf as below

```xml
<argument>from:=to</argument>
[gazebo-1] [WARN] [rcl]: Found remap rule '~/from:=to'. This syntax is deprecated. Use '--ros-args --remap ~/from:=to' instead.

```
becomes
```xml
<remapping>~/from:=to</remapping>
```

Note: tests worlds have been updated accordingly but I'm not sure there is coverage for topics being remapped

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
@@ -32,9 +32,9 @@ class GazeboRosAckermannDrivePrivate;
 
       <ros>
         <namespace>demo</namespace>
-        <argument>cmd_vel:=cmd_demo</argument>
-        <argument>odom:=odom_demo</argument>
-        <argument>distance:=distance_demo</argument>
+        <remapping>cmd_vel:=cmd_demo</remapping>
+        <remapping>odom:=odom_demo</remapping>
+        <remapping>distance:=distance_demo</remapping>
       </ros>
 
       <update_rate>100.0</update_rate>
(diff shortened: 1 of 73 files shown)


chapulina (approved):
Thank you for the PR, @mikaelarguedas . It works for me. I pushed more changes on 114adee8262a271e207ffc12d1f5b8644204cf3c so we don't have any demos or tests using deprecated functionality. Let me know if you're ok with that commit.

---

I also noticed several tests are failing locally. Didn't do much probing, but I'm almost certain it's not this PR's fault. Queued a CI build: [Build Status](https://ci.ros2.org/job/ci_linux/8581/)

mikaelarguedas:
> pushed more changes on 114adee so we don't have any demos or tests using deprecated functionality. Let me know if you're ok with that commit.

:+1:

---

> I also noticed several tests are failing locally. Didn't do much probing, but I'm almost certain it's not this PR's fault

All tests were passing for me locally (I ran only tests on `gazebo_ros` and `gazebo_plugins `packages)

chapulina:
> All tests were passing for me locally

Thanks for letting me know, it could be a problem with my local setup. I just tried it against the latest `eloquent` debs. I'll just wait for CI to come back then :wink:

PR #1012, opened 2019-10-22
[ros2] catch const ref to fix -Wcatch-value warnings

mikaelarguedas:
This warning appears on GCC8 and higher.

```
/tmp/src/gazebo_ros_pkgs/gazebo_ros/test/plugins/sdf_node.cpp:98:32: warning: catching polymorphic type ‘class rclcpp::exceptions::ParameterNotDeclaredException’ by value [-Wcatch-value=]
   98 |   } catch (rclcpp::exceptions::ParameterNotDeclaredException) {
```

Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>


chapulina (approved):
Merci!

PR #1014, opened 2019-11-01
[ROS2] add joint_effort plugin

igricart:
send effort to joint via sensor_msgs/msg/joint_state instead of geometry_msgs/msg/wrench

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_effort.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_effort.hpp
@@ -1,4 +1,4 @@
-// Copyright 2012 Open Source Robotics Foundation, Inc.
+// Copyright 2019 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -31,7 +31,7 @@ class GazeboRosJointEffortPrivate;
 
   Example Usage:
   \code{.xml}
-    <plugin name="gazebo_ros_effort" filename="libgazebo_ros_effort.so">
+    <plugin name="gazebo_ros_joint_effort" filename="libgazebo_ros_joint_effort.so">
 
       <ros>
 
@@ -39,7 +39,7 @@ class GazeboRosJointEffortPrivate;
         <namespace>/test</namespace>
 
         <!-- Remap the default topic -->
-        <argument>gazebo_ros_effort:=effort_test</argument>
+        <argument>joint_efforts:=effort_test</argument>
 
       </ros>
 
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_effort.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_effort.hpp
@@ -1,0 +1,79 @@
+// Copyright 2012 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GAZEBO_PLUGINS__GAZEBO_ROS_JOINT_EFFORT_HPP_
+#define GAZEBO_PLUGINS__GAZEBO_ROS_JOINT_EFFORT_HPP_
+
+#include <gazebo/common/Plugin.hh>
+#include <sensor_msgs/msg/joint_state.hpp>
+
+#include <memory>
+
+namespace gazebo_plugins
+{
+class GazeboRosJointEffortPrivate;
+
+/// This plugin collects data from a ROS topic and applies effort to a joint accordingly.
+/**
+  \details The last received effort will be continuously added to the joint at every simulation iteration.
+  Send an empty / zero message to stop applying a effort.
+
+  Example Usage:
+  \code{.xml}
+    <plugin name="gazebo_ros_effort" filename="libgazebo_ros_effort.so">
+
+      <ros>
+
+        <!-- Add a namespace -->
+        <namespace>/test</namespace>
+
+        <!-- Remap the default topic -->
+        <argument>gazebo_ros_effort:=effort_test</argument>
+
+      </ros>
+
+      <!-- Name of joint within model which will receive the effort -->
+      <joint_name>joint</joint_name>
+    </plugin>
+  \endcode
+*/
+
+class GazeboRosJointEffort : public gazebo::ModelPlugin
+{
+public:
+  /// Constructor
+  GazeboRosJointEffort();
+
+  /// Destructor
+  virtual ~GazeboRosJointEffort();
+
+protected:
+  // Documentation inherited
+  void Load(gazebo::physics::ModelPtr model, sdf::ElementPtr sdf) override;
+
+  /// Optional callback to be called at every simulation iteration.
+  virtual void OnUpdate();
+
+private:
+  /// Callback when a ROS Wrench message is received
+  /// \param[in] msg The Incoming ROS message representing the new effort to
+  /// exert.
+  void OnRosJointStateMsg(const sensor_msgs::msg::JointState::SharedPtr msg);
+
+  /// Private data pointer
+  std::unique_ptr<GazeboRosJointEffortPrivate> impl_;
+};
+}  // namespace gazebo_plugins
+
+#endif  // GAZEBO_PLUGINS__GAZEBO_ROS_JOINT_EFFORT_HPP_
--- a/gazebo_plugins/src/gazebo_ros_joint_effort.cpp
+++ b/gazebo_plugins/src/gazebo_ros_joint_effort.cpp
@@ -1,0 +1,128 @@
+// Copyright 2013 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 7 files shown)


gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_effort.hpp
```diff
@@ -0,0 +1,79 @@
+// Copyright 2012 Open Source Robotics Foundation, Inc.
```
chapulina:
```suggestion
// Copyright 2019 Open Source Robotics Foundation, Inc.
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_effort.hpp
```diff
@@ -0,0 +1,79 @@
...
+class GazeboRosJointEffortPrivate;
+
+/// This plugin collects data from a ROS topic and applies effort to a joint accordingly.
+/**
+  \details The last received effort will be continuously added to the joint at every simulation iteration.
+  Send an empty / zero message to stop applying a effort.
+
+  Example Usage:
+  \code{.xml}
+    <plugin name="gazebo_ros_effort" filename="libgazebo_ros_effort.so">
```
chapulina:
```suggestion
    <plugin name="gazebo_ros_joint_effort" filename="libgazebo_ros_joint_effort.so">
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_joint_effort.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+      <ros>
+
+        <!-- Add a namespace -->
+        <namespace>/test</namespace>
+
+        <!-- Remap the default topic -->
+        <argument>gazebo_ros_effort:=effort_test</argument>
```
chapulina:
```suggestion
        <argument>joint_efforts:=effort_test</argument>
```

I think it needs to be like this for proper remapping

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
```diff
@@ -0,0 +1,128 @@
...
+  sdf::ElementPtr joint_elem = sdf->GetElement("joint_name");
+  while (joint_elem) {
+    auto joint_name = joint_elem->Get<std::string>();
+
+    auto joint = model->GetJoint(joint_name);
+    if (!joint) {
+      RCLCPP_ERROR(impl_->ros_node_->get_logger(), "Joint %s does not exist!", joint_name.c_str());
+    } else {
+      impl_->joints_.push_back(joint);
+      RCLCPP_INFO(impl_->ros_node_->get_logger(), "Going to publish joint [%s]",
```
chapulina: This message doesn't correspond to the plugin's behaviour

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
```diff
@@ -0,0 +1,128 @@
...
+    "joint_efforts", rclcpp::SystemDefaultsQoS(),
+    std::bind(&GazeboRosJointEffort::OnRosJointStateMsg, this, std::placeholders::_1));
+
+  // Callback on every iteration
+  impl_->update_connection_ = gazebo::event::Events::ConnectWorldUpdateBegin(
+    std::bind(&GazeboRosJointEffort::OnUpdate, this));
+
+  impl_->joint_state_msg_.effort.resize(impl_->joints_.size());
+  impl_->joint_state_msg_.position.resize(impl_->joints_.size());
+  impl_->joint_state_msg_.velocity.resize(impl_->joints_.size());
```
chapulina: are `position` and `velocity` ever used?

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+void GazeboRosJointEffort::OnRosJointStateMsg(const sensor_msgs::msg::JointState::SharedPtr msg)
+{
+  if (msg->effort.size() != impl_->joints_.size()) {
+    RCLCPP_ERROR(
+      impl_->ros_node_->get_logger(),
+      "Joint input size differs from msg");
+  }
```
chapulina, line 114: I'm curious, why pass the joint names through the SDF if they could just be sent with the [JointState](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/JointState.html) message?

With the current implementation, the order of `<joint_name>`s has to match the order of joint efforts, which could be error prone.

chapulina (changes requested):
Thank you for the new plugin! I checked the existing plugins, and even though this one offers functionality similar to `gazebo_ros_joint_pose_trajectory` and `gazebo_ros_effort` (added to `eloquent` in #941) , I think it's different enough that it deserves to be its own plugin.

Besides the comments below, could I also ask you to add a [test](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-contribution-guide#testing) for the new plugin?

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
igricart: not really. I thought about using position and velocity as feedback measures. In my application I am using both `joint_state_publisher` (as feedback for a control loop) and `joint_effort` for control input signal. Any suggestions?

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
igricart: Do you mean that the joint names should be used as validation instead of the size ? If yes, that's a good idea and I will implement that way. If no, could you please rephrase?

igricart (commented):
I will make the stated modifications and add a test. However, the test will be kind of dumb (just checking whether the joint velocity is over a threshold after a given time). Sounds good ?

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
chapulina: I was mainly wondering if it was an oversight and something was missing from the implementation. Since these won't be used, I think it should be ok to remove these 2 lines and leave the vectors empty.

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
chapulina: Sorry for not being clear. Yes, my suggestion is to remove `<joint_name>` completely and just use `msg->name[i]` instead. This would require calling `auto joint = model->GetJoint(msg->name[i]);` whenever a new message is received though. It's more flexible, but potentially less performant.

igricart:
Just made the changes we have talked about. There is one observation ,though. The `jointStates` msg should have the name from all joints, otherwise it will find the joint but won't assign the input signal to the right one. For example, if we consider double pendulum with base model , there are two joints: upper and lower. If I publish a msg containing only `lower_joint` and the torque, the signal will be assigned to `upper_joint`.

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+void GazeboRosJointEffort::OnRosJointStateMsg(const sensor_msgs::msg::JointState::SharedPtr msg)
+{
+  if(msg->name.empty()) {
+      RCLCPP_ERROR(
+        impl_->ros_node_->get_logger(),
+        "No joint names in msg");
+        return;
+  }
+  for (size_t i = 0; i < msg->effort.size(); ++i){
```
chapulina: Couldn't you just copy the entire message instead of iterating and checking each field?

`impl_->joint_state_msg_ = msg;`

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
```diff
@@ -0,0 +1,144 @@
...
+        msg->name[i].c_str());
+        return;
+    }
+    impl_->joint_state_msg_.effort[i] = msg->effort[i];
+  }
+  impl_->joint_state_msg_.header = msg->header;
+}
+
+void GazeboRosJointEffort::OnUpdate()
+{
```
chapulina, line 127: If you copy the whole message as I mentioned before, you could do the name checking here instead, so there's no need to keep track of indices at all. You'd do something like:

```.cpp
void GazeboRosJointEffort::OnUpdate()
{
  if (impl_->joint_state_msg_.name.size() != impl_->joint_state_msg_.effort.size()) {
    // print error
    return;
  }

  for (size_t i = 0; i < impl_->joint_state_msg_.name.size(); ++i){
    auto joint = impl_->model_->GetJoint(impl_->joint_state_msg_.name[i]);
    if (joint)
      joint->SetForce(0, impl_->joint_state_msg_.effort[i]);
  }
}
```

chapulina (changes requested):
It looks like you mixed some commits changing the IMU which are unrelated to this PR. Mind removing them?

> it will find the joint but won't assign the input signal to the right one

This would make the plugin quite error-prone and it would be difficult for developers to find out what they're doing wrong. See my proposal about checking the names at `OnUpdate` and let me know what you think.

gazebo_plugins/src/gazebo_ros_joint_effort.cpp
igricart: I've tried to implement this way but `effort.size()` is different than `name.size()`. Because of that, I've also added a case to check whether `name` is empty.  Thank you for your suggestions and your heads up concerning the IMU commit

chapulina: > effort.size() is different than name.size()

It sounds like the message may be malformed? The [spec](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/JointState.html) says that _"All arrays in this message should have the same size, or be empty"_. I believe the name is the only one which shouldn't be empty, because _"Each joint is uniquely identified by its name"_.

igricart: It makes sense. In your opinion, which conditions of malformed msgs should be verified?

igricart (commented):
I think all the required changes have already been made. What do you think? Is it ok?

chapulina:
Sorry, @igricart , I haven't had time to review this again. I'll try to do it as soon as possible.

igricart:
> Sorry, @igricart , I haven't had time to review this again. I'll try to do it as soon as possible.

No worries!

PR #1017, opened 2019-11-13
Made the MultiCameraPlugin really multi-camera.

peci1:
Backwards compatibility on SDF/functional level should be retained. One virtual method was removed and one was removed in gazebo_ros_multicamera.cpp, gazebo_ros_triggered_multicamera.cpp and MultiCameraPlugin.cpp.

SDF parameter <frameName> became optionally a comma-separated list specifying a different frame for each camera.

Optional SDF parameter <cameraSuffixes> was added. If empty or non-present, "/<sdf_camera_name>" is used as the suffix. If nonempty, it is treated as a comma-separated list of camera suffixes. This list has to have the same lengths as the number of cameras. Any separators (like "/") should be part of the specified suffixes.


GrigorijSkult:
See #13.

Issue #123, opened 2013-10-01
gazebo_ros_diff_drive crashes on delete


maxbader:
Hi

It looks like that the gazebo_ros_diff_drive plug-in cases the gzserver to crash when a robot is removed (deleted) from the world

Ubuntu 12.04 (32bit)
ROS Hydro

Greetings

scpeters:
Did you manage to get a backtrace?

maxbader:
deleting the robot with the diff_drive plugin terminates the gzserver with the following msgs:

```
[ INFO] [1380650883.039866268, 0.047000000]: waitForService: Service [/gazebo/set_physics_properties] is now available.
[ INFO] [1380650883.119690601, 0.092000000]: Physics dynamic reconfigure ready.
[ INFO] [1380650893.319212552, 10.184000000]: Starting GazeboRosDiffDrive Plugin (ns = /pioneer2dx/)!
pure virtual method called
terminate called without an active exception
Aborted (core dumped)
[gazebo-2] process has died [pid 4626, exit code 134, cmd /opt/ros/hydro/lib/gazebo_ros/gzserver /home/max/projects/catkin_ws/src/v4r_gazebo/worlds/4walls.world __name:=gazebo __log:=/home/max/.ros/log/68978a4c-2ac4-11e3-a3f0-0024e8b0560b/gazebo-2.log].
log file: /home/max/.ros/log/68978a4c-2ac4-11e3-a3f0-0024e8b0560b/gazebo-2*.log
```

I am trying now to compile the diff_drive by my own and to add some debug outputs.
By the way the gazebo-2.log files do not exist

maxbader:
Hi

I figured it out, the problem is within the destructor of plug-in.
The current code just deletes the rosnode_ poniter to without calling shutdown first.

```
GazeboRosDiffDrive::~GazeboRosDiffDrive() {
  delete rosnode_;
  delete transform_broadcaster_;
  }
```

This destructor solves the problem

```
GazeboRosDiffDrive::~GazeboRosDiffDrive() {
  delete transform_broadcaster_;  // I am not sure but I think it is better do delete the boradcaster first
  rosnode_->shutdown();
  delete rosnode_;
}
```

Greetings 
Markus

j-rivero:
I haven't been able to reproduce this problem with ROS Jade. Please reopen if it still present on your system.

Abhinav93:
[...]

PR #1019, opened 2019-12-06
Fix destructor of gazebo_ros_diff_drive.cpp

RemiRigal:
Fix issue referenced in #123 where the destructor of ROS DiffDrive plugin causes `gzserver` to crash on model deletion.


chapulina (approved):
Looks reasonable to me, thanks!

RemiRigal:
Glad to contribute !

Might be reasonable to apply the PR to the other branches as well (Melodic at least).

chapulina:
> Might be reasonable to apply the PR to the other branches as well (Melodic at least).

Agreed. If you could open a PR against Melodic, I can merge it once CI comes back clean. Thanks for the help!

PR #1021, opened 2019-12-07
Fix destructor of gazebo_ros_diff_drive.cpp

RemiRigal:
Add fix for `melodic-devel` branch: #1019.

Fix issue referenced in #123 where the destructor of ROS DiffDrive plugin causes gzserver to crash on model deletion.


chapulina (approved):
Thanks for the forward-port!

PR #1024, opened 2019-12-17
gazebo_ros: Wait spawning Gazebo Model until ros::Time::now() is not 0.

marcin-sielski:
Why:

* Plugin incorrectly sends failure status message.

This change addresses the need by:

* Looping over ros::Time::now() until non-zero is returned.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -2660,6 +2660,8 @@ bool GazeboRosApiPlugin::spawnAndConform(TiXmlDocument &gazebo_model_xml, const 
 
   /// \brief poll and wait, verify that the model is spawned within Hardcoded 10 seconds
   ros::Duration model_spawn_timeout(10.0);
+  /// Wait until time can be read.
+  while (ros::Time::now().is_zero());
   ros::Time timeout = ros::Time::now() + model_spawn_timeout;
 
   while (ros::ok())


gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -2660,6 +2660,8 @@ bool GazeboRosApiPlugin::spawnAndConform(TiXmlDocument &gazebo_model_xml, const
 
   /// \brief poll and wait, verify that the model is spawned within Hardcoded 10 seconds
   ros::Duration model_spawn_timeout(10.0);
+  /// Wait until time can be read.
+  while (ros::Time::now().is_zero());
```
chapulina, line 2664: If simulation is started paused, this line will hang until simulation starts running. I'm not sure if this is desired.

tfoote: It might make more sense to switch this to be wall time not ROS time. I'm assuming this is resource contention based and not simulation time passing. Then we won't need to worry about whether ROS time is initialized.

tomniac:
I encountered this problem while spawning models as gazebo starts using worlds that had the `sim_time`, `real_time` and `wall_time` tags in the world definition set to positive values. This meant the `timeout` variable was getting set when `ros::Time::now()` was returning 0 but as soon as the world actually started, the clock would skip forward to the defined `sim_time` and the spawn would immediately time out.

It's not a solution to this issue as the spawn model service should handle this case, but if you're reading this, check the `sim_time` of your world.

grassjelly:
> I encountered this problem while spawning models as gazebo starts using worlds that had the `sim_time`, `real_time` and `wall_time` tags in the world definition set to positive values. This meant the `timeout` variable was getting set when `ros::Time::now()` was returning 0 but as soon as the world actually started, the clock would skip forward to the defined `sim_time` and the spawn would immediately time out.
> 
> It's not a solution to this issue as the spawn model service should handle this case, but if you're reading this, check the `sim_time` of your world.

This worked for me

gazebo_ros/src/gazebo_ros_api_plugin.cpp
marcin-sielski: As far as I understand the comment form @spark4160 it is related to simulation time passing.

kvnptl:
> I encountered this problem while spawning models as gazebo starts using worlds that had the `sim_time`, `real_time` and `wall_time` tags in the world definition set to positive values. This meant the `timeout` variable was getting set when `ros::Time::now()` was returning 0 but as soon as the world actually started, the clock would skip forward to the defined `sim_time` and the spawn would immediately time out.
> 
> It's not a solution to this issue as the spawn model service should handle this case, but if you're reading this, check the `sim_time` of your world.

Setting `sim_time` to 0, fixed my problem

roboticlemon:
Changing `sim_time` in my .world file also fixed this issue for me.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
HHendy69: Where would I find the file gazebo_ros_api_plugin?

marcin-sielski: https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_ros/src/gazebo_ros_api_plugin.cpp

HHendy69: I installed gazebo ros packages using sudo apt-get install ros-melodic-gazebo-ros-pkgs ros-melodic-gazebo-ros-control and I can not find this directory!
image

Where would I find the file on melodic distribution?

rk800506:
> Changing `sim_time` in my .world file also fixed this issue for me.

can u just show where did you set the 'sim_time' .. image of part of the world file

kvnptl:
> 
> 
> > Changing `sim_time` in my .world file also fixed this issue for me.
> 
> can u just show where did you set the 'sim_time' .. image of part of the world file

check your world file, 
image

rk800506:
> 
> 
> > > Changing `sim_time` in my .world file also fixed this issue for me.
> > 
> > 
> > can u just show where did you set the 'sim_time' .. image of part of the world file
> 
> check your world file,
> image

thanks a lot.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
kjyothiswaroop: @hichamhendy did you find a solution for this

HHendy69: No, I skipped it!!

rainer-d: I checked the terminal that launched my gazebo world and found the errors that I needed to fix there. So if the URDF file has any issues, it doesn't load and times out.

SwastikNandan:
This was both helpful and inciteful.

SwastikNandan:
> I encountered this problem while spawning models as gazebo starts using worlds that had the `sim_time`, `real_time` and `wall_time` tags in the world definition set to positive values. This meant the `timeout` variable was getting set when `ros::Time::now()` was returning 0 but as soon as the world actually started, the clock would skip forward to the defined `sim_time` and the spawn would immediately time out.
> 
> It's not a solution to this issue as the spawn model service should handle this case, but if you're reading this, check the `sim_time` of your world.

This was helpful. It worked like magic!

Issue #243, opened 2014-09-12
gazebo_ros_openni_kinect plugin with roll + pitch orientation generates wrong depth image


efernandez:
The problem is probably because of a wrong depth image coming from gazebo, as explained here:
http://answers.gazebosim.org/question/7005/gazebo_ros_openni_kinect-plugin-with-roll-pitch/

I'm asking here because it's related with a plugin tough.

j-rivero:
There is a good explanation of the problem and a possible workaround in [ros answers](http://answers.ros.org/question/53784/simulated-kinect-rotation-around-x-bug/?answer=55968#post-id-55968)

annamannucci:
Hi, I have the same problem. I'm using gazebo 4, ROS Indigo with Ubuntu 14.04. I would like to know if someone has tried to correct this bug.

felixvd:
This is still a bug in Kinetic. I get incorrectly rotated point clouds. It seems like something is wrong with the coordinate systems, because moving the robot results in behavior like shown below.

![camprob2](https://user-images.githubusercontent.com/4535737/42564381-8d433518-853b-11e8-9cfb-ce806f1265e9.png)
![camprob3](https://user-images.githubusercontent.com/4535737/42564382-8d69839e-853b-11e8-9716-bb0045d142ab.png)

In these pictures, the camera is looking down at the bins. The point cloud is not displayed at the correct place. When the robot moves to the goal pose of the second picture, the point cloud rotates in more than one axis:

![camprob4](https://user-images.githubusercontent.com/4535737/42564385-8d97e7e8-853b-11e8-8a59-ec502f92a26d.png)

Another shot with an extra movement:
![camprob5](https://user-images.githubusercontent.com/4535737/42564386-8dbd8d2c-853b-11e8-8b49-4e3be994395e.png)

And a close-up of the frames:
![camprob1](https://user-images.githubusercontent.com/4535737/42564380-8d1aa918-853b-11e8-9cdd-4b577ef8c594.png)

I assumed that the realsense camera URDF I am using is rotating the camera frames, and gazebo might not like that. The question posted by @j-rivero seems to imply as much. But removing all rotations from the camera's xacro so that the frames are oriented like this does not change the behavior:

![camprob6](https://user-images.githubusercontent.com/4535737/42565101-7ada625a-853d-11e8-891a-2b1c19c06833.png)

[...]

PR #1027, opened 2020-01-03
[gazebo_plugins] fix Issue #243: Use x-forward z-up frame for pointcloud

sachazyto:
I've observed issue #243 on a melodic-based robot simulation I ran yesterday. Using the ROS body frame convention (x-forward, z-up), solved the issue for me (cf attached screenshot)
Screenshot from 2020-01-03 15-18-13


lunaluxie:
I had the same problem also on melodic, this commit fixed it.

vib2810:
Could you please share the .so file "libgazebo_ros_openni_kinect.so" so we can use this fix until the commit is merged?

AlexKaravaev:
Why it's not merged yet? It's been already 6 years since issue first appeared... Unless there are some not so obvious reasons why should point cloud be oriented like z-forward, but I can not think such use-case.

PR #1028, opened 2020-01-06
[Windows][ros2] Fix gazebo launch for windows

jnoyola:
`gazebo.launch.py` currently fails on Windows because the path env vars use colons to separate paths. This will fix it to use the platform-specific path separator (semicolons for Windows).


chapulina (approved):
Thanks, LGTM

PR #1031, opened 2020-01-11
Image publishers use SensorDataQoSProfile

jacobperron:
All other sensor publishers were updated previously to use the same profile (#926).
I'm not sure if the image publishers were overlooked or the image_transport API didn't
support setting the QoS profile at the time.


chapulina (approved):
LGTM, just waiting for CI

PR #1033, opened 2020-01-17
[ros2] make transient local reliable

Karsten1987:
When using the robot state publisher topic (`robot_description`) to spawn an entity into Gazebo, the topics don't match because the subscription reliability QoS don't match the publisher.

I am wondering whether there is a mismatch between Python and C++ that by default a Python subscription is best effort whereas in C++ it's reliable. 

Both, in Dashing as well as Eloquent, we weren't able to spawn objects correctly with the `spawn_entity.py` script.

/cc @clalancette @relffok

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/gazebo_ros/CMakeLists.txt
+++ b/gazebo_ros/CMakeLists.txt
@@ -13,11 +13,11 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(builtin_interfaces REQUIRED)
-find_package(rcl REQUIRED)
-find_package(rclcpp REQUIRED)
 find_package(gazebo_dev REQUIRED)
 find_package(gazebo_msgs REQUIRED)
 find_package(geometry_msgs REQUIRED)
+find_package(rcl REQUIRED)
+find_package(rclcpp REQUIRED)
 find_package(std_srvs REQUIRED)
 find_package(tinyxml_vendor REQUIRED)
 
--- a/gazebo_ros/CMakeLists.txt
+++ b/gazebo_ros/CMakeLists.txt
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
 project(gazebo_ros)
 
-
 # Default to C++14
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
@@ -22,10 +21,6 @@ find_package(geometry_msgs REQUIRED)
 find_package(std_srvs REQUIRED)
 find_package(tinyxml_vendor REQUIRED)
 
-include_directories(
-  include
-  ${gazebo_dev_INCLUDE_DIRS}
-)
 link_directories(
   ${gazebo_dev_LIBRARY_DIRS}
 )
@@ -37,6 +32,13 @@ add_library(gazebo_ros_node SHARED
   src/executor.cpp
   src/node.cpp
 )
+target_include_directories(gazebo_ros_node
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>
+    ${gazebo_dev_INCLUDE_DIRS}
+)
+
 ament_target_dependencies(gazebo_ros_node
   "gazebo_dev"
   "rclcpp"
@@ -47,6 +49,11 @@ ament_export_libraries(gazebo_ros_node)
 # gazebo_ros_utils
 add_library(gazebo_ros_utils SHARED
   src/utils.cpp
+)
+target_include_directories(gazebo_ros_utils
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>
 )
 ament_target_dependencies(gazebo_ros_utils
   "gazebo_dev"
--- a/gazebo_ros/src/gazebo_ros_force_system.cpp
+++ b/gazebo_ros/src/gazebo_ros_force_system.cpp
@@ -22,6 +22,7 @@
 #include <gazebo_msgs/srv/link_request.hpp>
 #include <gazebo_ros/node.hpp>
 
+#include <algorithm>
 #include <memory>
 #include <string>
 #include <vector>
@@ -355,14 +356,16 @@ void GazeboRosForceSystemPrivate::ClearLinkWrenches(
   gazebo_msgs::srv::LinkRequest::Response::SharedPtr /*_res*/)
 {
   std::lock_guard<std::mutex> scoped_lock(lock_);
-  auto it = std::find_if(link_wrench_tasks_.begin(), link_wrench_tasks_.end(),
-      [_req](auto & link_wrench_task) {
-        return link_wrench_task->link->GetScopedName() == _req->link_name;
-      });
-  if (it != link_wrench_tasks_.end()) {
-    link_wrench_tasks_.erase(it);
-    RCLCPP_INFO(ros_node_->get_logger(), "Deleted wrench on [%s]", _req->link_name.c_str());
-  } else {
+  auto prev_end = link_wrench_tasks_.end();
+  link_wrench_tasks_.erase(std::remove_if(link_wrench_tasks_.begin(), link_wrench_tasks_.end(),
+    [_req, this](auto & link_wrench_task) {
+      if (link_wrench_task->link->GetScopedName() == _req->link_name) {
+        RCLCPP_INFO(ros_node_->get_logger(), "Deleted wrench on [%s]", _req->link_name.c_str());
+        return true;
+      }
+      return false;
+    }), link_wrench_tasks_.end());
+  if (prev_end == link_wrench_tasks_.end()) {
     RCLCPP_WARN(ros_node_->get_logger(), "No applied wrenches on [%s]", _req->link_name.c_str());
   }
 }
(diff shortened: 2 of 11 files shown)


chapulina:
Interesting that this hasn't come up before. Do you think you'd be able to add a test for spawning from a topic? If a test would be too much, maybe at least add an example to `spawn_entity_demo.launch.py`?

Karsten1987:
@chapulina as mentioned offline, https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1033/commits/00bf7ff4377bcaf215641e217d760cd04e4ceaa1 fixes the actual compilation/tests for master. 

regarding the tests, I would propose to utilize launch_test to provide a node which publishes some xml model with the same QOS as the robot state publisher and run the entity spawner script after that and see if the scripts returns successfully. However, I don't have a clue about launch testing yet. So any help would be appreciated here.

gazebo_ros/CMakeLists.txt
```diff
@@ -37,6 +32,13 @@ add_library(gazebo_ros_node SHARED
   src/executor.cpp
   src/node.cpp
 )
+target_include_directories(gazebo_ros_node
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>
+    ${gazebo_dev_INCLUDE_DIRS}
+)
+
```
Karsten1987, line 41: This fix is similar in behavior as https://github.com/ros/console_bridge/pull/63. `gazebo_dev` exports GTest headers which are different to what `ament_cmake_gtest` provides and ultimately clashes - at least on my OSX machine.

gazebo_ros/src/gazebo_ros_force_system.cpp
```diff
@@ -354,18 +354,15 @@ void GazeboRosForceSystemPrivate::ClearLinkWrenches(
...
-  }
-  if (!found) {
+  auto it = std::find_if(link_wrench_tasks_.begin(), link_wrench_tasks_.end(),
+      [_req](auto & link_wrench_task) {
+        return link_wrench_task->link->GetScopedName() == _req->link_name;
+      });
+  if (it != link_wrench_tasks_.end()) {
+    link_wrench_tasks_.erase(it);
+    RCLCPP_INFO(ros_node_->get_logger(), "Deleted wrench on [%s]", _req->link_name.c_str());
+  } else {
```
Karsten1987: `cppcheck` complained about potential pointer invalidation. Looking at the code, it can be conveniently rewritten with a simple `std::find_if`.

Karsten1987:
https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1033/commits/b9a2ba2b76c23e489c8d97bc3dc42b5b222309b9 adds a launch test which starts a mock robot state publisher, gzserver and finally checks if the spawn entity script exits correctly; meaning the entity is spawned correctly.

@hidmic could you please have a look at the provided launch test? I've ran into quite some issues when changing `launch.actions.OpaqueFunction(function=lambda context: ready_fn())` to the actual advised `launch_testing.actions.ReadyToTest()`

@chapulina Can you please run CI on this?

Karsten1987:
It took a bit to figure out what's happening, but I believe I've understood now that the `entity_spawner` never used an actual latching QoS. 
https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1033/commits/d45abffb704a9c612a14e2458f473ef209b2ba16 fixes that. The default values are correctly set, so there is no need to additionally set the reliability to `reliable`.

chapulina:
> Can you please run CI on this?

Triggered here, hopefully I didn't mess it up: [Build Status](https://ci.ros2.org/job/ci_linux/9102/)

Also documented the CI triggering process here: https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-CI

Karsten1987:
@chapulina your CI branch has to be rebased on the latest master.

Specifically you're missing this commit: https://github.com/ros2/ci/pull/364

chapulina:
Rebased and retriggered: [Build Status](https://ci.ros2.org/job/ci_linux/9103/) (after I triggered I noticed that `CI_TEST_ARGS` will test everything. I can't cancel the build, so it will just take a while...)

Karsten1987:
Re-triggered testing only the related `gazebo_ros` pkg: [Build Status](https://ci.ros2.org/job/ci_linux/9104/)

Karsten1987:
and once more: [Build Status](https://ci.ros2.org/job/ci_linux/9108/)

Karsten1987:
not giving up on this one: [Build Status](https://ci.ros2.org/job/ci_linux/9110/)

Karsten1987:
boy, I start to love this PR: [Build Status](https://ci.ros2.org/job/ci_linux/9115/)

Karsten1987:
Exciting times. Build has passed.

Test ran successfully:
```
15:02:55 19: [INFO] [ros2-3]: process started with pid [1701]
15:02:55 19: [ros2-3] [INFO] [1579734175.644794957] [spawn_entity]: Spawn Entity started
15:02:55 19: [ros2-3] [INFO] [1579734175.645249507] [spawn_entity]: Loading entity published on topic /mock_robot_description
15:02:55 19: [ros2-3] [INFO] [1579734175.646847036] [spawn_entity]: Waiting for entity xml on /mock_robot_description
15:02:55 19: [ros2-3] [INFO] [1579734175.744446788] [spawn_entity]: Waiting for service /spawn_entity
15:02:55 19: [ros2-3] [INFO] [1579734175.748032487] [spawn_entity]: Calling service /spawn_entity
15:02:55 19: [ros2-3] [INFO] [1579734175.878373861] [spawn_entity]: Spawn status: SpawnEntity: Successfully spawned entity [mock_robot_state_entity]
15:02:55 19: [INFO] [ros2-3]: process has finished cleanly [pid 1701]
```

So this is now up for review and should definitely be backported to eloquent.

gazebo_ros/src/gazebo_ros_force_system.cpp
chapulina: I think the new code is slightly different though, because it won't erase multiple occurrences of the same name.

gazebo_ros/test/entity_spawner.test.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
chapulina:
```suggestion
# Copyright 2020 Open Source Robotics Foundation, Inc.
```

:grimacing:

gazebo_ros/test/entity_spawner.test.py
```diff
@@ -0,0 +1,67 @@
...
+            output='screen'
+        ),
+        IncludeLaunchDescription(
+            PythonLaunchDescriptionSource([
+                ament_index_python.get_package_share_directory('gazebo_ros'),
+                '/launch', '/gzserver.launch.py'
+                ])
+        ),
+        launch.actions.OpaqueFunction(function=lambda context: ready_fn())
+        # launch_testing.actions.ReadyToTest()
```
chapulina: Can the comment be removed?

chapulina (changes requested):
Thank you for the tests and for addressing all those other issues :muscle:

gazebo_ros/test/entity_spawner.test.py
Karsten1987: I am waiting on @hidmic to comment on this. `ReadyToTest` is apparently the preferred way of handling this, but it yields to problems on my machine. So I hope to get this solved.

Karsten1987:
@chapulina https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1033/commits/fee09f64c62fa2a273d743ac7b39a53cd99d80c0 should address your concerns about removing multiple instances.

CI: [Build Status](https://ci.ros2.org/job/ci_linux/9116/)

gazebo_ros/test/entity_spawner.test.py
hidmic: It is the preferred way indeed, and thus I'm curious why it doesn't work for you.

Karsten1987: The tests fails with output similar to this:

```
19: [INFO] [sh-1]: process started with pid [74368]
19: Timed out waiting for processes to start up
19: [INFO] [sh-1]: sending signal 'SIGINT' to process[sh-1]
```
```
19: [INFO] [gzserver-2]: process has finished cleanly [pid 73832]
19: Processes under test stopped before tests completed
19: -- run_test.py: return code 1
```

The test run correctly when using the _old_ method.

hidmic: @Karsten1987 did you remove `ready_fn` from the arguments' list, as in [examples/ready_action_test.py](https://github.com/ros2/launch/blob/a2f3255aa7a9823fbf330269f170d1350d8a0378/launch_testing/test/launch_testing/examples/ready_action_test.py) when you tried `launch_testing.actions.ReadyToTest`? On a probably unpleasant note, trying to use both old and new methods simultaneously may leave you in a state where neither works (see [here](https://github.com/ros2/launch/blob/f73bc56caa9c7b3c53b264475728392d0e1ce9cf/launch_testing/launch_testing/loader.py#L37-L69)).

Karsten1987: That seems to work. See https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1033/commits/880cf8bb9f174c95d0aa5626a43bfebeb40105be

Is there a way to warn the user of this UB? Something like issuing a warning if `ready_fn` is available but the `OpaqueFunction` isn't used? I don't know the implications of this in the first place, just thinking out loud.

Karsten1987:
what a time to be alive. Next round of CI hopefully sorting out the launch testing behavior: 
[Build Status](https://ci.ros2.org/job/ci_linux/9122/)

chapulina (approved):
Thank you for all the extra fixes. It all looks good to me.

I tested the changes on `gazebo_ros_force_system` by using #1035 like this:

* Launch the demo world: `ros2 launch gazebo_ros gazebo.launch.py world:="src/gazebo_ros_pkgs/gazebo_ros/worlds/gazebo_ros_force_system_demo.world"`
* Insert a box from the top toolbar
* Apply 1st wrench: `ros2 service call /apply_link_wrench gazebo_msgs/srv/ApplyLinkWrench '{link_name: "unit_box::link", wrench: { force: { x: 10 } }, duration: {sec: -1, nanosec: 0} }'`
* Apply 2nd wrench: `ros2 service call /apply_link_wrench gazebo_msgs/srv/ApplyLinkWrench '{link_name: "unit_box::link", wrench: { force: { y: 10 } }, duration: {sec: -1, nanosec: 0} }'`
* Clear wrenches: `ros2 service call /clear_link_wrenches 'gazebo_msgs/srv/LinkRequest' '{link_name: "unit_box::link"}'`
* Reset world (`Ctrl+R`) and check that the box doesn't move, because both wrenches have been correctly deleted.

(While testing this I noticed that the wrenches are not actually accumulating and only the last wrench is applied. I think we may want to use `AddForce` instead of `SetForce` but I can also see a use case for `SetForce`, so I won't touch the plugin for now.)

PR #1035, opened 2020-01-24
[ros2] Enable the force system on launch files

chapulina:
The force system was ported on #941 , but we hadn't enabled it on the launch files.

With this change, it is loaded by default, and can be disabled with `force_system:=false`.


ahcorde (approved):
LGTM, tested locally too

PR #1039, opened 2020-02-04
Add parameter to spawn_entity.py.

clalancette:
ROS 1 had the ability to specify the entity XML via a parameter
on the command-line.  This re-introduces that to ROS 2, as we
don't always want to have a topic published by another entity
to do this.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

This has been useful in one of the projects I'm currently working on.  If this PR is not desirable, I'm looking for another way to pass the entity XML on the command-line.

### DIFF
--- a/gazebo_ros/scripts/spawn_entity.py
+++ b/gazebo_ros/scripts/spawn_entity.py
@@ -57,6 +57,8 @@ class SpawnEntityNode(Node):
                             help='Load entity xml from file')
         source.add_argument('-topic', type=str, metavar='TOPIC_NAME',
                             help='Load entity xml published on topic')
+        source.add_argument('-param', type=str, metavar='PARAM_NAME',
+                            help='Load entity xml contained in parameter')
         source.add_argument('-database', type=str, metavar='ENTITY_NAME',
                             help='Load entity XML from specified entity in GAZEBO_MODEL_PATH \
                             or Gazebo Model Database')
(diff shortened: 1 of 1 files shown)


IanTheEngineer:
> I'm looking for another way to pass the entity XML on the command-line.

Specifically, we needed a mechanism to pass a fully constructed urdf string (from xacro) in via ros2 launchfile syntax

chapulina:
Supporting a parameter besides the topic sounds fine to me. My only concern is about what the recommended pattern will be moving forward. On #948 we chose to use a topic because that's what RViz and `robot_state_publisher` were doing:

* https://github.com/ros-visualization/rviz/issues/1125
* https://github.com/ros2/robot_state_publisher/issues/14#issuecomment-429919738
* https://answers.ros.org/question/304645/ros2-robot-state-publisher-publishes-empty-robot_description/

This may not be visible to everyone, but I ticketed a request to add a note about this to the ROS1 - ROS2 migration guide (https://github.com/osrf/doc_requests/issues/4), which starts with 

> On ROS 1, packages like RViz, MoveIt and `gazebo_ros_pkgs` expected a robot's description (i.e. URDF file) to be stored in a global parameter. Packages like `robot_state_publisher` would be the ones setting that parameter so all other tools could get it.
> On ROS 2, there's no longer a parameter server, so the implementation has changed. Now the robot description is published by `robot_state_publisher` on a latched topic (i.e. transient local), and RViz et.al. subscribe to that topic to get the URDF.
> This change affects many packages, so the documentation should probably go on the Migration guide from ROS 1.

---

If we move forward with this, do you mind documenting it on the wiki with an example? https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Spawn-and-delete#spawn-entity-node

eborghi10:
Are there plans of moving forward with this, @clalancette?

clalancette:
> Are there plans of moving forward with this, @clalancette?

Probably not at the moment.  The project I was working with for this has ended, so I probably won't have time in the near future to take a look.  If you'd like to continue with it, please feel free to take this over.

PR #1038, opened 2020-02-04
Created a MarkerArray to publish normals in ROS

ahcorde:
We are exposing [normals in Gazebo9](https://bitbucket.org/osrf/gazebo/pull-requests/3193/added-normals-in-depth-camera-sensor/diff). This PR exposes the normals as a MarkerArray and then we can visualize the normals in RVIZ

Selección_014

PR #1047, opened 2020-02-11
[Gazebo 9] Created a MarkerArray to publish normals in ROS

ahcorde:
Moved this PR #1038 to `noetic-devel` to avoid ABI changes in `melodic`

We are exposing [normals in Gazebo9](https://bitbucket.org/osrf/gazebo/pull-requests/3193/added-normals-in-depth-camera-sensor/diff). This PR exposes the normals as a MarkerArray and then we can visualize the normals in RVIZ

Selección_014

Steps to reproduce it:

 - Launch the test world: 
`rosrun gazebo_ros gazebo test_worlds/gazebo_ros_depth_camera.world`
 - Create the tf between the world and the camera: 
`rosrun tf2_ros static_transform_publisher 0 0 5 0 0 -1.57 world base_link`
 - Open RVIZ

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
@@ -90,7 +90,7 @@ namespace gazebo
                    unsigned int _width, unsigned int _height,
                    unsigned int _depth, const std::string &_format);
 
-    /// \brief Update the controller
+    // Documentation inherited
     protected: virtual void OnNewNormalsFrame(const float * _normals,
                    unsigned int _width, unsigned int _height,
                    unsigned int _depth, const std::string &_format);
@@ -108,7 +108,9 @@ namespace gazebo
 
     /// \brief Keep track of number of connctions for point clouds
     private: int normals_connect_count_;
+    /// \brief Increase the counter which count the subscribers are connected
     private: void NormalsConnect();
+    /// \brief Decrease the counter which count the subscribers are connected
     private: void NormalsDisconnect();
 
     /// \brief Keep track of number of connctions for point clouds
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
@@ -35,6 +35,8 @@
 #include <sensor_msgs/fill_image.h>
 #include <std_msgs/Float64.h>
 #include <image_transport/image_transport.h>
+#include <visualization_msgs/MarkerArray.h>
+#include <visualization_msgs/Marker.h>
 
 // gazebo stuff
 #include <sdf/Param.hh>
@@ -88,6 +90,11 @@ namespace gazebo
                    unsigned int _width, unsigned int _height,
                    unsigned int _depth, const std::string &_format);
 
+    /// \brief Update the controller
+    protected: virtual void OnNewNormalsFrame(const float * _normals,
+                   unsigned int _width, unsigned int _height,
+                   unsigned int _depth, const std::string &_format);
+
     /// \brief Put camera data to the ROS topic
     private: void FillPointdCloud(const float *_src);
 
@@ -98,6 +105,11 @@ namespace gazebo
     private: int point_cloud_connect_count_;
     private: void PointCloudConnect();
     private: void PointCloudDisconnect();
+
+    /// \brief Keep track of number of connctions for point clouds
+    private: int normals_connect_count_;
+    private: void NormalsConnect();
+    private: void NormalsDisconnect();
 
     /// \brief Keep track of number of connctions for point clouds
     private: int depth_image_connect_count_;
@@ -116,15 +128,21 @@ namespace gazebo
     /// \brief A pointer to the ROS node.  A node will be instantiated if it does not exist.
     private: ros::Publisher point_cloud_pub_;
     private: ros::Publisher depth_image_pub_;
+    private: ros::Publisher normal_pub_;
 
     /// \brief PointCloud2 point cloud message
     private: sensor_msgs::PointCloud2 point_cloud_msg_;
     private: sensor_msgs::Image depth_image_msg_;
 
+    private: float * pcd_ = nullptr;
+
     private: double point_cloud_cutoff_;
 
     /// \brief ROS image topic name
     private: std::string point_cloud_topic_name_;
+
+    /// \brief ROS normals topic name
+    private: std::string normals_topic_name_;
 
     private: void InfoConnect();
     private: void InfoDisconnect();
@@ -148,4 +166,3 @@ namespace gazebo
 
 }
 #endif
-
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
@@ -140,6 +140,9 @@ namespace gazebo
 
     private: double point_cloud_cutoff_;
 
+    /// \brief adding one value each reduce_normals_ to the array marker
+    private: int reduce_normals_;
+
     /// \brief ROS image topic name
     private: std::string point_cloud_topic_name_;
 
--- a/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_depth_camera.cpp
@@ -114,6 +114,11 @@ void GazeboRosDepthCamera::Load(sensors::SensorPtr _parent, sdf::ElementPtr _sdf
   else
     this->point_cloud_cutoff_ = _sdf->GetElement("pointCloudCutoff")->Get<double>();
 
+  if (!_sdf->HasElement("reduceNormals"))
+    this->reduce_normals_ = 50;
+  else
+    this->reduce_normals_ = _sdf->GetElement("reduceNormals")->Get<int>();
+
   load_connection_ = GazeboRosCameraUtils::OnLoad(boost::bind(&GazeboRosDepthCamera::Advertise, this));
   GazeboRosCameraUtils::Load(_parent, _sdf);
 }
@@ -430,7 +435,7 @@ void GazeboRosDepthCamera::OnNewNormalsFrame(const float * _normals,
             m.pose.orientation.w = q.w();
 
             // plotting some of the normals, otherwise rviz will block it
-            if (index % 50 == 0)
+            if (index % this->reduce_normals_ == 0)
               m_array.markers.push_back(m);
           }
         }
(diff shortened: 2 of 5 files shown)


gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
```diff
@@ -88,6 +90,11 @@ namespace gazebo
                    unsigned int _width, unsigned int _height,
                    unsigned int _depth, const std::string &_format);
 
+    /// \brief Update the controller
+    protected: virtual void OnNewNormalsFrame(const float * _normals,
+                   unsigned int _width, unsigned int _height,
+                   unsigned int _depth, const std::string &_format);
```
chapulina, line 96: Could you add `override`? I know the other functions don't have it, but it's probably because they predate C++11.

The doxygen also doesn't make sense to me. You could just write `// Documentation inherited` since the upstream function is already documented.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h
```diff
@@ -99,6 +106,11 @@ namespace gazebo
     private: void PointCloudConnect();
     private: void PointCloudDisconnect();
 
+    /// \brief Keep track of number of connctions for point clouds
+    private: int normals_connect_count_;
+    private: void NormalsConnect();
+    private: void NormalsDisconnect();
```
chapulina, line 114: Mind documenting all the new functions?

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -45,9 +45,11 @@ GZ_REGISTER_SENSOR_PLUGIN(GazeboRosDepthCamera)
 GazeboRosDepthCamera::GazeboRosDepthCamera()
 {
   this->point_cloud_connect_count_ = 0;
+  this->normals_connect_count_ = 0;
   this->depth_image_connect_count_ = 0;
   this->depth_info_connect_count_ = 0;
   this->last_depth_image_camera_info_update_time_ = common::Time(0);
+
```
chapulina: Please remove the extra line at the end of the block, some linters complain about this

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -250,6 +283,12 @@ void GazeboRosDepthCamera::OnNewRGBPointCloud(const float *_pcd,
     if (this->point_cloud_connect_count_ > 0)
     {
       this->lock_.lock();
+
+      if(pcd_ == nullptr)
```
chapulina:
```suggestion
      if (pcd_ == nullptr)
```

Style nit :grin:

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -327,6 +366,72 @@ void GazeboRosDepthCamera::OnNewImageFrame(const unsigned char *_image,
...
+               unsigned int _width, unsigned int _height,
+               unsigned int _depth, const std::string &_format)
+{
+  if (!this->initialized_ || this->height_ <=0 || this->width_ <=0)
+    return;
+
+  visualization_msgs::MarkerArray m_array;
+
+  this->lock_.lock();
+  if(this->point_cloud_msg_.data.size()>0){
```
chapulina:
```suggestion
  if (this->point_cloud_msg_.data.size() > 0) {
```

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -327,6 +366,72 @@ void GazeboRosDepthCamera::OnNewImageFrame(const unsigned char *_image,
...
+        tf::Quaternion q(right_vector, -1.0*acos(axis_vector.dot(vector)));
+        q.normalize();
+
+        m.pose.orientation.x = q.x();
+        m.pose.orientation.y = q.y();
+        m.pose.orientation.z = q.z();
+        m.pose.orientation.w = q.w();
+
+        // plotting some of the normals, otherwise rviz will block it
+        if(index%50==0)
```
chapulina:
```suggestion
        if (index % 50 == 0)
```

gazebo_plugins/test/test_worlds/gazebo_ros_depth_camera.world
```diff
@@ -104,15 +104,23 @@
...
                         </image>
-                        <clip near="0.01" far="100.0" />
-                        <save enabled="false" path="/tmp" />
-                        <depth_camera output="points" />
+                        <clip>
+                          <near>0.1</near>
+                          <far>100</far>
+                        </clip>
+                        <save enabled='0'>
+                          <path>__default__</path>
```
chapulina: Did you mean to change the path? Since the default is disabled, I think the whole `save` block could be removed.

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -250,6 +283,12 @@ void GazeboRosDepthCamera::OnNewRGBPointCloud(const float *_pcd,
     if (this->point_cloud_connect_count_ > 0)
     {
       this->lock_.lock();
+
+      if(pcd_ == nullptr)
+        pcd_ = new float[_width * _height * 4];
```
iche033, line 290: delete this in the destructor?

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -156,6 +171,24 @@ void GazeboRosDepthCamera::PointCloudDisconnect()
...
+  this->parentSensor->SetActive(true);
+}
+////////////////////////////////////////////////////////////////////////////////
+// Decrement count
+void GazeboRosDepthCamera::NormalsDisconnect()
+{
+  this->normals_connect_count_--;
+  (*this->image_connect_count_)--;
+  if (this->normals_connect_count_ <= 0)
+    this->parentSensor->SetActive(false);
```
iche033: hmm if there are still subscribers to the point cloud topic (or other topics advertised by the depth camera), this line will disable the sensor causing no msgs to be published. I think we should disable the sensor only when there are no subscribers to all the topics

chapulina (commented):
I haven't looked into why, but when I `rosrun` the commands on the PR description, Gazebo just hangs with a black screen. Can you reproduce it? If you can't, I can dig into my setup later. The `worlds/depth_camera2.world` world doesn't hang though.

gazebo_plugins/test/test_worlds/gazebo_ros_depth_camera.world
```diff
@@ -104,15 +104,23 @@
...
                             <format>R8G8B8</format>
                         </image>
-                        <clip near="0.01" far="100.0" />
-                        <save enabled="false" path="/tmp" />
-                        <depth_camera output="points" />
+                        <clip>
+                          <near>0.1</near>
+                          <far>100</far>
+                        </clip>
+                        <save enabled='0'>
```
iche033: we can remove the <save> tag as it's not needed

ahcorde:
You should run the example with `world/gazebo_ros_depth_camera.world` inside this package.

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -327,6 +363,84 @@ void GazeboRosDepthCamera::OnNewImageFrame(const unsigned char *_image,
...
+            tf::Quaternion q(right_vector, -1.0*acos(axis_vector.dot(vector)));
+            q.normalize();
+
+            m.pose.orientation.x = q.x();
+            m.pose.orientation.y = q.y();
+            m.pose.orientation.z = q.z();
+            m.pose.orientation.w = q.w();
+
+            // plotting some of the normals, otherwise rviz will block it
+            if (index % 50 == 0)
```
chapulina: Wouldn't it be best if this was controlled by the camera's frame rate? Right now it's hidden from the end user and they can't change it.

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -327,6 +363,84 @@ void GazeboRosDepthCamera::OnNewImageFrame(const unsigned char *_image,
...
+    if (this->normals_connect_count_ > 0)
+      // do this first so there's chance for sensor to run 1 frame after activate
+      this->parentSensor->SetActive(true);
+  }
+  else
+  {
+    if (this->normals_connect_count_ > 0)
+    {
+      this->lock_.lock();
+      if (this->point_cloud_msg_.data.size() > 0){
```
chapulina, line 391: There's currently an implicit dependency between `points` and `normals`. If I run the test world, then echo only the normals, I get empty marker arrays. Once I also echo the points, then the marker array is populated.

I understand that  in order to create the markers we need the XYZ coordinates as well as the normals. So I think that on `OnNewRGBPointCloud`, we should also populate `point_cloud_msg` in case `this->normals_connect_count_ > 0`.

chapulina (changes requested):

chapulina (dismissed):
Works for me. Since @iche033 is also reviewing, let's wait for his approval before merging.

chapulina:
Actually, I just realized that this change will make this branch not compilable against earlier Gazebo 9 versions. @j-rivero , what's the minimum Gazebo 9 version that we need to support for Noetic? I imagine that's 9.12.0, which is going into Focal? In that case, we may need to wrap the new APIs with `ifdef`s.

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
```diff
@@ -327,6 +368,84 @@ void GazeboRosDepthCamera::OnNewImageFrame(const unsigned char *_image,
...
+            tf::Quaternion q(right_vector, -1.0*acos(axis_vector.dot(vector)));
+            q.normalize();
+
+            m.pose.orientation.x = q.x();
+            m.pose.orientation.y = q.y();
+            m.pose.orientation.z = q.z();
+            m.pose.orientation.w = q.w();
+
+            // plotting some of the normals, otherwise rviz will block it
+            if (index % this->reduce_normals_ == 0)
```
iche033, line 398: move this check earlier? We could skip creating the marker msg and doing computations when the results are not being published.

iche033:
looks good to me! just left one minor comment.

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
ahcorde: fixed ;)

chapulina (approved):
> compiling against 9.12.0

I talked to @j-rivero about this and since the official Gazebo version for Noetic is not decided yet, let's not worry about it now. We can go back and fix it if needed.

iche033 (approved):

gazebo_plugins/src/gazebo_ros_depth_camera.cpp
iche033: sorry didn't catch this earlier: `index` is declared later so it is not resolved here and this `if` block is missing curly brackects

ahcorde: I had this resolved in this other PR, but I can create a Pr just solving this issue

Issue #1043, opened 2020-02-10
Help to contribute a gazebo_ros_magnetometer plugin


Darkproduct:
To get a full simulation of our robot, we need a magnetometer sensor that publishes a `sensor_msgs/MagneticField` message. Unfortunately, there is only one magnetometer plugin out there that I know of. `gazebo_ros_magnetic` from [hector_gazebo_plugins](http://wiki.ros.org/hector_gazebo_plugins). 
The three main reasons why I don't want to use it are:

- It outputs a `geometry_msgs::Vector3Stamped` instead of `sensor_msgs/MagneticField`.
- It doesn't use gazebo world `<magnetic_field>` tag in sdf file.
- It is implemented as `ModelPlugin` and not as `SensorPlugin`.

Gazebo has a `MagnetometerSensor` since 2015 and no ROS plugin. See [this commit](https://bitbucket.org/osrf/gazebo/pull-requests/1788/add-magnetometer-sensor-redo-of-andrews/diff) and [this discussion](https://bitbucket.org/osrf/gazebo/issues/1294/inclusion-of-magnetic-field-strength).

I used `gazebo_ros_imu_sensor` as template and created [gazebo_ros_magnetometer](https://github.com/Darkproduct/gazebo_ros_magnetometer).

urdf entry:
```
<gazebo reference="magnetometer_link">
  <sensor name="magnetometer_sensor" type="magnetometer">
    <always_on>true</always_on>
    <update_rate>100</update_rate>
    <visualize>true</visualize>
    <topic>magnetometer</topic>
    <plugin filename="libgazebo_ros_magnetometer.so" name="magnetometer_plugin">
      <topicName>/robot/imu/mag</topicName>
      <frameName>magnetometer_link</frameName>
      <updateRateHZ>100.0</updateRateHZ>
    </plugin>
    <pose>0 0 0 0 0 0</pose>
  </sensor>
</gazebo>
```

But for some reason, my plugin never gets executed. The ros node is never initialized and therefore no ros topic gets advertised. My `GazeboRosMagneticSensor` class doesn't even get to its constructor. 

The only success I had was, with the magnetometer_sensor gazebo plugin. In Gazebo a magnetometer topic is advertised and `gz topic -e` outputs:
```
time {
  sec: 23
  nsec: 932000000
}
field_tesla {
  x: 4.8184694239418247e-06
  y: 2.3297836655578027e-05
  z: -4.1988012093698152e-05
}
```

[...]

PR #1048, opened 2020-02-11
[gazebo_plugins] Add Gazebo ROS magnetometer sensor

Darkproduct:
This is my followup pull request from #1043 .

To get a full simulation of our robot, we need a magnetometer sensor that publishes a `sensor_msgs/MagneticField` message. Unfortunately, there is only one magnetometer plugin out there that I know of. `gazebo_ros_magnetic` from [hector_gazebo_plugins](http://wiki.ros.org/hector_gazebo_plugins). 
The three main reasons why I don't want to use it are:

- It outputs a `geometry_msgs::Vector3Stamped` instead of `sensor_msgs/MagneticField`.
- It doesn't use gazebo world `<magnetic_field>` tag in sdf file.
- It is implemented as `ModelPlugin` and not as `SensorPlugin`.

Gazebo has a `MagnetometerSensor` since 2015 and no ROS plugin. See [this commit](https://bitbucket.org/osrf/gazebo/pull-requests/1788/add-magnetometer-sensor-redo-of-andrews/diff) and [this discussion](https://bitbucket.org/osrf/gazebo/issues/1294/inclusion-of-magnetic-field-strength).

I used `gazebo_ros_imu_sensor` as template and created [gazebo_ros_magnetometer_sensor](https://github.com/Darkproduct/gazebo_ros_magnetometer). 

**For everyone, who want to use this right now:**
Clone [gazebo_ros_magnetometer_sensor](https://github.com/Darkproduct/gazebo_ros_magnetometer_sensor) into `catkin_ws/src/` and it should work.

### DIFF
--- a/gazebo_plugins/CMakeLists.txt
+++ b/gazebo_plugins/CMakeLists.txt
@@ -115,6 +115,7 @@ catkin_package(
   gazebo_ros_planar_move
   gazebo_ros_range
   gazebo_ros_vacuum_gripper
+  gazebo_ros_magnetometer_sensor
 
   CATKIN_DEPENDS
   message_runtime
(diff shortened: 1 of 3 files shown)


Darkproduct:
Hmm, I don't see why there is this problem from `ros_gazebo_pkgs-ci-pr_any_kinetic-xenial-amd64`. Have I done something wrong?

Darkproduct:
Hi, it's more then a year now. Are there any issues with this PR, or why is it still open?

PR #1062, opened 2020-03-19
gazebo_ros_control: catch all pluginlib exceptions

xqms:
Before we only displayed an error message on LibraryLoadExceptions, but there is also ClassLoaderException, CreateClassException, InvalidXMLException, and some more. So catch pluginlib::PluginlibException, which is base class of all of these.

Uncaught exceptions don't show up at all (probably a bug in gazebo itself), which leads to situations which are quite hard to debug.

This should apply to noetic as well.

### DIFF
--- a/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp
+++ b/gazebo_ros_control/src/gazebo_ros_control_plugin.cpp
@@ -199,7 +199,7 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
       (boost::bind(&GazeboRosControlPlugin::Update, this));
 
   }
-  catch(pluginlib::LibraryLoadException &ex)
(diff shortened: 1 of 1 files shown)


chapulina (approved):
LGTM. Even though a clarification of what you mean by this would be appreciated:

> Uncaught exceptions don't show up at all (probably a bug in gazebo itself)

xqms:
> LGTM. Even though a clarification of what you mean by this would be appreciated:
> 
> > Uncaught exceptions don't show up at all (probably a bug in gazebo itself)

In my case a CreateClassException was thrown, the catch block didn't catch it, and instead of gazebo crashing (which I would expect with an uncaught exception), there was no output at all. I assume that there is some exception handling in gazebo that kicks in and aborts loading of the GazeboRosControl plugin, but does not print `e.what()` (or anything else).

PR #1069, opened 2020-03-26
[Noetic] changes to make it work with Python3

ahcorde:
Changes to make it work with Python3

### DIFF
--- a/gazebo_plugins/scripts/gazebo_model
+++ b/gazebo_plugins/scripts/gazebo_model
@@ -56,7 +56,7 @@ if __name__ == "__main__":
 
         model_msg = GazeboModel(model_name,param_name,GazeboModel.URDF_PARAM_NAME,namespace,Pose(Point(lx,ly,lz),Quaternion()))
         success = gazebo_plugins_interface.load_model(model_msg)
-        print("spawning success", success)
+        print("spawning success ", success)
 
     elif sys.argv[1] == 'delete':
         print("not implemented yet")
--- a/gazebo_plugins/scripts/gazebo_model
+++ b/gazebo_plugins/scripts/gazebo_model
@@ -15,9 +15,10 @@ from gazebo_plugins.msg import GazeboModel
 from geometry_msgs.msg import Pose, Point, Quaternion
 
 def usage():
-    print('Commands:')
-    print('load [param|file] [urdf|gazebo] <param/file name> <model name> <initial pose: x y z r p y> <namespace> - Load model')
-    print('delete <model name>                                                        - Deletes the model named <model name>')
+    print('''Commands:
+    load [param|file] [urdf|gazebo] <param/file name> <model name> <initial pose: x y z r p y> <namespace> - Load model
+    delete <model name>                                                        - Deletes the model named <model name>
+    ''')
     sys.exit(1)
 
 if __name__ == "__main__":
--- a/gazebo_plugins/scripts/gazebo_model
+++ b/gazebo_plugins/scripts/gazebo_model
@@ -15,32 +15,31 @@ from gazebo_plugins.msg import GazeboModel
 from geometry_msgs.msg import Pose, Point, Quaternion
 
 def usage():
-    print '''Commands:
-    load [param|file] [urdf|gazebo] <param/file name> <model name> <initial pose: x y z r p y> <namespace> - Load model
-    delete <model name>                                                        - Deletes the model named <model name>
-    '''
+    print('Commands:')
+    print('load [param|file] [urdf|gazebo] <param/file name> <model name> <initial pose: x y z r p y> <namespace> - Load model')
+    print('delete <model name>                                                        - Deletes the model named <model name>')
     sys.exit(1)
 
 if __name__ == "__main__":
     if len(sys.argv) < 13:
-        print usage()
+        usage()
 
     if sys.argv[1] == 'load':
         if sys.argv[2] == 'param':
-            print "loading from parameter"
+            print("loading from parameter")
         elif sys.argv[2] == 'file':
-            print "loading from file"
+            print("loading from file")
         else:
-            print "invalid syntax, first argument has to be either param or file"
-            print usage()
+            print("invalid syntax, first argument has to be either param or file")
+            usage()
 
         if sys.argv[3] == 'urdf':
-            print "loading urdf"
+            print("loading urdf")
         elif sys.argv[3] == 'gazebo':
-            print "loading gazebo model"
+            print("loading gazebo model")
         else:
-            print "invalid syntax, xml type has to be either urdf or gazebo"
-            print usage()
+            print("invalid syntax, xml type has to be either urdf or gazebo")
+            usage()
 
         param_name = sys.argv[4]
         model_name = sys.argv[5]
@@ -57,12 +56,11 @@ if __name__ == "__main__":
 
         model_msg = GazeboModel(model_name,param_name,GazeboModel.URDF_PARAM_NAME,namespace,Pose(Point(lx,ly,lz),Quaternion()))
         success = gazebo_plugins_interface.load_model(model_msg)
-        print "spawning success",success
+        print("spawning success", success)
 
     elif sys.argv[1] == 'delete':
-        print "not implemented yet"
+        print("not implemented yet")
 
     else:
-        print "unknown command: ",sys.argv[1]
-        print usage()
-
+        print("unknown command: ", sys.argv[1])
+        usage()
--- a/gazebo_plugins/scripts/gazebo_model
+++ b/gazebo_plugins/scripts/gazebo_model
@@ -57,7 +57,7 @@ if __name__ == "__main__":
 
         model_msg = GazeboModel(model_name,param_name,GazeboModel.URDF_PARAM_NAME,namespace,Pose(Point(lx,ly,lz),Quaternion()))
         success = gazebo_plugins_interface.load_model(model_msg)
-        print("spawning success ", success)
+        print("spawning success", success)
 
     elif sys.argv[1] == 'delete':
         print("not implemented yet")
--- a/gazebo_plugins/scripts/set_pose.py
+++ b/gazebo_plugins/scripts/set_pose.py
@@ -95,14 +95,14 @@ class SimIfaceControl():
     rospy.init_node(NAME, anonymous=True)
 
   def setPoseService(self,pose_msg):
-    print "waiting for service to set pose"
+    print('waiting for service to set pose')
     rospy.wait_for_service(self.service_name);
     try:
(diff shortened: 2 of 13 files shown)


ahcorde:
For some reason when the second test starts in `gazebo_ros` called `ros_network/ros_network_disabled.test` Gazebo dies immediately. The test fails with a timeout failure.

I tested each one of them individually, and it works. any thoughts?

ahcorde:
Chnged enviroment variable `GAZEBO_MASTER_URI` to avoid the issue the collision of gazebo instances because of the binded ports

gazebo_plugins/scripts/gazebo_model
```diff
@@ -57,12 +56,11 @@ if __name__ == "__main__":
 
         model_msg = GazeboModel(model_name,param_name,GazeboModel.URDF_PARAM_NAME,namespace,Pose(Point(lx,ly,lz),Quaternion()))
         success = gazebo_plugins_interface.load_model(model_msg)
-        print "spawning success",success
+        print("spawning success", success)
```
mabelzhang, line 60: Probably a space to separate the string from the value would look nicer
```suggestion
        print("spawning success ", success)
```

gazebo_plugins/test/bumper_test/test_bumper.py
```diff
@@ -87,12 +87,12 @@ def checkContact(self):
...
         if self.sample_count > 20:
           if abs(self.fz_avg - 98.1) < 0.01:
-            print "z force ",self.fz_avg
+            print("z force ", self.fz_avg)
             self.success = True
 
 
     def test_bumper(self):
-        print "LNK\n"
+        print("LNK")
```
mabelzhang: Is the `\n` removed on purpose?

gazebo_plugins/test/p3d_test/test_link_pose.py
```diff
@@ -130,19 +130,19 @@ def validP3dInput(self, p3d):
...
+              + " count: " + str(self.link_sample_count))
         if self.link_sample_count >= self.min_link_samples:
           if self.link_error_rms < self.tolerance:
             if self.link_error_max < self.max_error:
               self.success = True
 
 
     def test_link_pose(self):
-        print "LNK\n"
+        print("LNK")
```
mabelzhang: Same as above

mabelzhang (commented):
A few minor comments. Should the Python files have `python3` in the shebang or are they intended to still support python 2?

`gazebo_ros` tests worked for me. I didn't run the `gazebo_plugins` tests (that's problem on my end. I can't get its dependencies to compile yet...)

ahcorde:
shebang lines look like : `#! /usr/bin/env python`. Ubuntu Focal only allows Python3 which I think it's fine with the python version installed in the system

gazebo_plugins/test/bumper_test/test_bumper.py
ahcorde: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1069/commits/d4198f56f4cd8510139aa36ff5693bc97751d21e

gazebo_plugins/test/p3d_test/test_link_pose.py
ahcorde: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1069/commits/d4198f56f4cd8510139aa36ff5693bc97751d21e

mabelzhang:
I think for Python3, usually the shebang lines would look like `#!/usr/bin/env python3`.

sloretz:
> A few minor comments. Should the Python files have python3 in the shebang or are they intended to still support python 2?

@mabelzhang @ahcorde IIUC PEP 394 recommends shebangs match the python version  (`python2` or `python3`) when they're installed. There's a CMake macro `catkin_install_python()` that writes the shebang to the specific python version used. The current recommendation is to use `#!/usr/bin/env python` and `catkin_install_python()` http://wiki.ros.org/UsingPython3/SourceCodeChanges#Changing_shebangs . Is that macro being used here?

mabelzhang:
Thank you for the clarification and the link!
Looks like the current `CMakeLists.txt` for `gazebo_plugins` and `gazebo_ros` do not have `catkin_install_python()`
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/noetic-devel/gazebo_plugins/CMakeLists.txt
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/noetic-devel/gazebo_ros/CMakeLists.txt

ahcorde:
the scripts in [gazebo_ros](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/noetic-devel/gazebo_ros/scripts) are written in bash not in Python. Should I install them with `catkin_install_python()`?

sloretz:
> the scripts in gazebo_ros are written in bash not in Python. Should I install them with catkin_install_python()?

Definitely not. If they're bash scripts, what has the `#!/usr/bin/env python` shebang?

ahcorde:
`gazebo_ros` has some, but right now it's fixed https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1069/commits/6cabcc7fd0f370d8c0ca2f0754d836e5d1888c4e

gazebo_plugins/scripts/gazebo_model
```diff
@@ -15,32 +15,31 @@ from gazebo_plugins.msg import GazeboModel
 from geometry_msgs.msg import Pose, Point, Quaternion
 
 def usage():
-    print '''Commands:
-    load [param|file] [urdf|gazebo] <param/file name> <model name> <initial pose: x y z r p y> <namespace> - Load model
-    delete <model name>                                                        - Deletes the model named <model name>
-    '''
+    print('Commands:')
+    print('load [param|file] [urdf|gazebo] <param/file name> <model name> <initial pose: x y z r p y> <namespace> - Load model')
```
sloretz: nit, this changes the whitespace in the output message from `____load` to `load`. I think putting `()` around the `'''...'''` should be fine.

gazebo_plugins/scripts/set_pose.py
```diff
@@ -191,20 +191,19 @@ def setPose(self):
...
-    -R <roll in radians>
-    -P <pitch in radians>
-    -Y <yaw in radians>
-    -f target frame_id
-    -s set pose service name
-    -t set pose topic name
-    -p wait for this ros topic to be published first
-'''
+    print('Commands:')
+    print('-update_rate <Hz> - update rate, default to 10 Hz')
```
sloretz: same nit that this changed leading whitespace

gazebo_plugins/scripts/set_wrench.py
```diff
@@ -107,17 +107,17 @@ def setWrench(self):
...
-    -x <x in meters>
-    -y <y in meters>
-    -z <z in meters>
-    -R <roll in radians>
-    -P <pitch in radians>
-    -Y <yaw in radians>
-    -t set wrench topic name
-'''
+    print('Commands:')
+    print('-update_rate <Hz> - update rate, default to 10 Hz')
```
sloretz: ditto about leading whitespace

gazebo_ros/test/ros_network/ros_api_checker
```diff
@@ -21,8 +21,9 @@ class Tester(unittest.TestCase):
         cmd = ['rostopic', 'list']
         po = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         out, err = po.communicate()
-        self.assertEqual(po.returncode, 0, 'rostopic failed (%s). stdout: %s stderr: %s'%(cmd, out, err))
-        topics_actual = set(out.split('\n')) - set([''])
+        self.assertEqual(po.returncode, 0, 'rostopic failed (%s). stdout: %s stderr: %s'%(cmd, out.decode(), err.decode()))
```
sloretz: nit, I recommend `subproces.Popen(..., text=True)` instead of adding calls to `decode()` so there are fewer lines changed by this PR

gazebo_ros/test/ros_network/ros_api_checker
```diff
@@ -188,7 +190,7 @@ if __name__ == '__main__':
     # amount of time when we're downloading models.
     while rospy.Time.now().to_sec() == 0.0:
         print('Waiting for Gazebo to start...')
-        time.sleep(1.0)
+        time.sleep(0.1)
```
sloretz: nit, I would avoid changing this timeout unless shortening is necessary for Noetic/Python 3

gazebo_plugins/scripts/gazebo_model
sloretz: switched to `('''...''')` in 8c1cad7

gazebo_plugins/scripts/set_pose.py
sloretz: switched to `('''...''')` in 8c1cad7

gazebo_plugins/scripts/set_wrench.py
sloretz: switched to `('''...''')` in 8c1cad7

gazebo_ros/test/ros_network/ros_api_checker
sloretz: Uses `text=True` in 9631b23

gazebo_plugins/scripts/gazebo_model
sloretz: `print()` joins arguments with a space already

```python3
>>> print("foo", "bar")
foo bar
```

sloretz: 4b50082

gazebo_ros/test/ros_network/ros_api_checker
sloretz: Restored sleep in 4b50082

gazebo_ros/test/ros_network/ros_api_checker
sloretz: oops, cleaned up stuff I missed in be09f22

sloretz (approved):
LGTM! Tests pass on my machine using Gazebo 11 (currently using gazebo11 from packages.osrfoundation.org, but hope to use packages.ros.org when that becomes available).

sloretz:
@chapulina @j-rivero I think this should go in before the other Noetic PRs since it has changes required for Noetic. Mind taking a look, and if it looks ok then merging it?

chapulina (approved):
Works for me. Let's just merge it like this, but beware that the `*11` rosdep keys are not working yet, see https://github.com/ros-infrastructure/reprepro-updater/pull/75#issuecomment-628879195.

PR #1071, opened 2020-03-31
Include preprocessor conditions to avoid compile OnNewNormalsFrame and OnNewReflectanceFrame

ahcorde:
`OnNewReflectanceFrame` and `OnNewNormalsFrame` will be introduce exclusively in Gazebo 9 for minor version > 12.

These preprocessor conditions allow to compile for other Gazebo versions


chapulina (approved):
LGTM. Once this is ported forward to Gazebo 10 and 11 we'll need to add their versions here too. Especially 11, since that's going to be the official version on Noetic.

PR #1073, opened 2020-03-31
Use configurable timeout in other wait for service calls

jacobperron:
Follow-up to #1072


chapulina (changes requested):
LGTM, I'd just change the argument description not to say `spawn entity service`

jacobperron:
Good point. Updated in e7c712d.

chapulina (approved):

PR #1076, opened 2020-04-08
Use <depend> for gazebo_dev in gazebo_ros

mabelzhang:
Attempt to fix build farm CMake errors about not finding `gazebo_dev`: https://build.osrfoundation.org/job/ros_gazebo_pkgs-ci-pr_regression_any_kinetic-xenial-amd64/121/consoleFull#-274357824aa60f765-a60a-427d-900c-dc128ab22287

The hypothesis is that `<build_depend>` is not enough. Now using `<depend>`. Removed `<exec_depend>` because otherwise compiler complains about redundancy.

Will see result in the automatic checks below.

### DIFF
--- a/gazebo_plugins/package.xml
+++ b/gazebo_plugins/package.xml
@@ -18,12 +18,11 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
-  <build_depend>gazebo_dev</build_depend>
   <!--
-    Need to use gazebo_dev since run script needs pkg-config
+    Need to exec_depend on gazebo_dev since run script needs pkg-config
     See: https://github.com/ros-simulation/gazebo_ros_pkgs/issues/323 for more info
   -->
-  <exec_depend>gazebo_dev</exec_depend>
+  <depend>gazebo_dev</depend>
 
   <depend>gazebo_msgs</depend>
   <depend>geometry_msgs</depend>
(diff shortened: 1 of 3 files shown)


mabelzhang:
Check did not pass with the `<depend>` fix :-/

scpeters:
I think `gazebo_plugins` needs it as well

scpeters:
oh that didn't help

mabelzhang:
It looks like `gazebo_dev` itself didn't build
https://build.osrfoundation.org/job/ros_gazebo_pkgs-ci-pr_any_kinetic-xenial-amd64/136/consoleFull#-274357824aa60f765-a60a-427d-900c-dc128ab22287
```
Aborted  <<< gazebo_dev
```
There's an error further up about setuptools, not sure which package it comes from:
```
[0.373s] ERROR:colcon.colcon_core.entry_point:Exception loading extension 'colcon_core.package_identification.python_setup_py': No module named 'setuptools.config', 'setuptools' needs to be at least version 30.3.0, if a newer version is not available from the package manager use 'pip3 install -U setuptools' to update to the latest version
```

mabelzhang:
Summary of findings tested locally:
Looks like installing `setuptools` this way as suggested in the error message
```
pip3 install -U setuptools
```
makes `colcon graph` happy, which otherwise does not print all the dependencies.
This probably needs to be done in CI.

mabelzhang:
Closing. Fixed by https://github.com/colcon/colcon-core/pull/336 as shown in [new build results](https://build.osrfoundation.org/job/ros_gazebo_pkgs-ci-default_kinetic-xenial-amd64/111/)

PR #1077, opened 2020-04-09
[eloquent] Fix Windows build.

seanyen:
This is a minimum fix to unblock Windows build breaks.

* Adding visibility control headers.
* Minimizing Windows header namespace exposure.
* Explicitly defining `_USE_MATH_DEFINES` to use math constants, like `M_PI`.
* Adding missing linkage dependency for `gazebo_ros_bumper`.
* Adding missing header inclusion for `gazebo_ros_ray_sensor.cpp`.

### DIFF
--- a/gazebo_ros/CMakeLists.txt
+++ b/gazebo_ros/CMakeLists.txt
@@ -11,6 +11,16 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic")
 endif()
 
+if(WIN32)
+  add_compile_definitions(
+    # For math constants
+    _USE_MATH_DEFINES
+    # Minimize Windows namespace collision
+    NOMINMAX
+    WIN32_LEAN_AND_MEAN
+  )
+endif()
+
 find_package(ament_cmake REQUIRED)
 find_package(builtin_interfaces REQUIRED)
 find_package(gazebo_dev REQUIRED)
@@ -38,6 +48,7 @@ target_include_directories(gazebo_ros_node
     $<INSTALL_INTERFACE:include>
     ${gazebo_dev_INCLUDE_DIRS}
 )
+target_compile_definitions(gazebo_ros_node PRIVATE "GAZEBO_ROS_NODE_BUILDING_DLL")
 
 ament_target_dependencies(gazebo_ros_node
   "gazebo_dev"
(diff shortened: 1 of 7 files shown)


seanyen:
@chapulina @j-rivero This is ready for review and merge. Thanks!

gazebo_ros/CMakeLists.txt
```diff
@@ -38,6 +48,7 @@ target_include_directories(gazebo_ros_node
     $<INSTALL_INTERFACE:include>
     ${gazebo_dev_INCLUDE_DIRS}
 )
+target_compile_definitions(gazebo_ros_node PRIVATE "GAZEBO_ROS_NODE_BUILDING_DLL")
```
chapulina: Shouldn't these only be done `if(WIN32)`?

gazebo_ros/include/gazebo_ros/utils_visibility_control.h
```diff
@@ -0,0 +1,56 @@
...
+
+#ifndef GAZEBO_ROS__UTILS_VISIBILITY_CONTROL_H_
+#define GAZEBO_ROS__UTILS_VISIBILITY_CONTROL_H_
+
+// This logic was borrowed (then namespaced) from the examples on the gcc wiki:
+//     https://gcc.gnu.org/wiki/Visibility
+
+#if defined _WIN32 || defined __CYGWIN__
+  #ifdef __GNUC__
+    #define GAZEBO_ROS_UTILS_EXPORT __attribute__ ((dllexport))
```
chapulina, line 30: On Gazebo and Ignition, we've been [using](https://bitbucket.org/osrf/gazebo/src/5f48d294547f2eccf368de73080470290dd6a3b1/gazebo/util/system.hh#lines-42) the `VISIBLE` suffix. Unless there's a good reason to use `EXPORT`, I'd suggest sticking to the existing pattern.

chapulina (changes requested):

gazebo_ros/include/gazebo_ros/utils_visibility_control.h
seanyen: I don't have strong opinion on the naming. The `visibility_control.h` is largely taken from the examples I can see from [ROS 2 GitHub org](https://github.com/search?q=org%3Aros2+filename%3Avisibility_control.h&type=Code). If you prefer to use `VISIBLE` suffix convention like Gazebo, then I can rename it.

gazebo_ros/CMakeLists.txt
seanyen: Added!

gazebo_ros/include/gazebo_ros/utils_visibility_control.h
chapulina: Thanks for the links, following ROS 2 convention sounds good to me.

chapulina:

seanyen:
@chapulina Following up.

chapulina (approved):
Seems reasonable to me. But I'd like @j-rivero 's opinion before merging.

gazebo_ros/include/gazebo_ros/utils_visibility_control.h
j-rivero: I agree here.

j-rivero:
> Seems reasonable to me. But I'd like @j-rivero 's opinion before merging.

All the changes seems completely reasonable to me. Failures and warnings in the CI are not related to this PR as far as I can say. I don't see anything dangerous. Thanks @seanyen @chapulina for the code and review.

j-rivero (approved):

PR #1074, opened 2020-04-08
Add node required parameter to empty_world.launch file

mabelzhang:
This exposes the `required` fields of the `<node>` tags for gzserver and gzclient in `empty_world.launch`, such that if either the server or client dies, the entire `roslaunch` will terminate too.

Actual names of parameters, currently `gui_required` and `server_required`, are up for suggestions.
Tested on Xenial.

Test with:
```
$ roslaunch gazebo_ros empty_world.launch gui_required:=true
```
Close the GUI window, and launch script should quit by itself:
```
================================================================================REQUIRED process [gazebo_gui-3] has died!
process has finished cleanly
log file: /home/developer/.ros/log/03fcfb24-793e-11ea-bd1e-0242ac110002/gazebo_gui-3*.log
Initiating shutdown!
================================================================================
[gazebo_gui-3] killing on exit
[gazebo-2] killing on exit
```

Test requirement on server with:
```
$ roslaunch gazebo_ros empty_world.launch server_required:=true
$ kill <gzserver_PID>
```
Launch script should quit, similar to above.

Signed-off-by: Mabel Zhang <mabel@openrobotics.org>

PR #1078, opened 2020-04-14
[forward port to melodic] Add node required parameter to empty world (#1074)

mabelzhang:
Forward port from #1074 in kinetic-devel to melodic-devel.

Only difference is in testing, it seems that in Melodic, closing the GUI window does not terminate gzclient - it is still shown in `top`. I had to `kill <gzclient_PID>` to kill the client. Then the launch script exits.


chapulina (approved):
This LGTM. And I got the same behaviour as you, the client is not dying on close.

PR #1082, opened 2020-04-18
[forward port to noetic] Add node required parameter to empty world (#1074)

mabelzhang:
Forward port from #1074 in kinetic-devel to noetic-devel.

Tested in Noetic + gazebo11 + Ubuntu 20.04.
Similar to Melodic (#1078), gzclient requires `kill` - closing GUI isn't enough.


chapulina (approved):
I didn't test with Noetic but the forward port looks reasonable.

PR #1086, opened 2020-04-23
[forward port to Foxy] Add node required parameter to launch (#1074) 

mabelzhang:
Adds `on_exit=Shutdown()` to `ExecuteProcess()`, conditional on launch arguments `server_required:=true` and  `gui_required:=true` specified on command line.

To test:
```
$ ros2 launch gazebo_ros gazebo.launch.py server_required:=true gui_required:=true
```

### Better alternative

The code is kind of ugly right now.  It duplicates `ExecuteProcess()`. One call has `IfCondition(LaunchCOnfiguration('server_required'))` and `on_exit=Shutdown()` in the case the node is required, e.g. `server_required:=true`. The other call has `UnlessCondition` in the case the node is not required. That is based on this [ROS Answers example](https://answers.ros.org/question/314903/ros2-optionally-launch-node-with-no-arguments/). It's similar to the `<group if>` in ROS 1 XML launch.

I've asked William whether there's a better way now, and he doesn't think we've implemented a better way. Probably this has to be implemented https://github.com/ros2/launch/issues/290 for a cleaner way to exist. #962  also works around it.

The [ROS 2 XML launch](https://index.ros.org/doc/ros2/Tutorials/Launch-files-migration-guide/#group) does have `<group if>`. But since the launch script is already in Python and XML launch doesn't exist before Eloquent, I just stuck to the Python script.

### Remaining issue

One issue out of the scope of this PR is that when the launch script automatically shuts down, it does not kill the other process successfully. If `gzserver` is killed, `gzclient` remains running and the GUI is still open, and vice versa, even though launch exits in the terminal.

The printout shows that SIGINT and SIGTERM were sent, but they did not terminate the process.
```
[gzserver-1] Terminated
[ERROR] [gzserver-1]: process has died [pid 6902, exit code 143, cmd 'gzserver                                                                    -s libgazebo_ros_init.so   -s libgazebo_ros_factory.so       '].
[INFO] [launch]: process[gzserver-1] was required: shutting down launched system
[INFO] [gzclient   -2]: sending signal 'SIGINT' to process[gzclient   -2]
[ERROR] [gzclient   -2]: process[gzclient   -2] failed to terminate '5' seconds after receiving 'SIGINT', escalating to 'SIGTERM'
[INFO] [gzclient   -2]: sending signal 'SIGTERM' to process[gzclient   -2]
[ERROR] [gzclient   -2]: process has died [pid 6904, exit code -15, cmd 'gzclient   '].
[INFO] [launch]: process[gzclient   -2] was required: shutting down launched system
```
The `SIGINT` and `SIGTERM` messages look similar for `gzserver` if the GUI is killed.


mabelzhang:
On [`empty_world.launch.py`](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_ros/launch/empty_world.launch.py), which I was going to update -

I now realize that "fixing" it would basically amount to it being identical to `gazebo.launch.py`. Do we want to just remove `empty_world.launch.py`? Or is it referenced somewhere in examples? Maybe we should just update the examples to point to `gazebo.launch.py` instead.

`empty_world.launch.py` is last modified 2 years ago and only starts `gzserver` and nothing else. So it isn't very useful. It doesn't even open the empty world promised ^^||

chapulina:
> Do we want to just remove empty_world.launch.py?

I'm ok with it being removed from Foxy.

mabelzhang:
Removed.

scpeters:

chapulina:
> @ros-pull-request-builder retest this please

@scpeters , just note that this re-triggers the ROS farm. For the OSRF farm the incantation is:

I assume that's what you were going for. I documented this [here](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-CI) because I always forget :sweat_smile:

j-rivero:
> I assume that's what you were going for. I documented this [here](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-CI) because I always forget sweat_smile

We can make osrf-jenkins to react to the same sentence we use for ros-pull-request-builder if that is useful.

chapulina:
> We can make osrf-jenkins to react to the same sentence we use for ros-pull-request-builder if that is useful.

I don't have a strong preference. I've been in situations where I only wanted to trigger one but not the other, so sometimes being different can be useful.

gazebo_ros/launch/gzclient.launch.py
```diff
@@ -54,23 +69,28 @@ def generate_launch_description():
                               description='Set "true" to run gzserver with gdb'),
         DeclareLaunchArgument('valgrind', default_value='false',
                               description='Set "true" to run gzserver with valgrind'),
+        DeclareLaunchArgument('gui_required', default_value='false',
```
j-rivero, line 72: it might be a bit late but `gui_required` inspire my something like "we want to spawn the GUI for sure when running the roslaunch". The description is far away from this thought and explains the shutdown behavior.

gazebo_ros/launch/gzclient.launch.py
```diff
@@ -54,23 +69,28 @@ def generate_launch_description():
...
+                              description='Set "true" to shut down launch script when GUI is terminated'),
 
-        # Execute
+        # Execute node with Shutdown if required
```
j-rivero: It really helped me here for undestanding the on_exit + condition combo your explanation in the pull request description:
```

Better alternative

The code is kind of ugly right now. It duplicates ExecuteProcess(). One call has IfCondition(LaunchCOnfiguration('server_required')) and on_exit=Shutdown() in the case the node is required, e.g. server_required:=true. The other call has UnlessCondition in the case the node is not required. That is based on this ROS Answers example. It's similar to the <group if> in ROS 1 XML launch.

I've asked William whether there's a better way now, and he doesn't think we've implemented a better way. Probably this has to be implemented ros2/launch#290 for a cleaner way to exist. #962 also works around it.

The ROS 2 XML launch does have <group if>. But since the launch script is already in Python and XML launch doesn't exist before Eloquent, I just stuck to the Python script.
```
Can we add a pointer here in the code or a short version of it?

gazebo_ros/launch/gzserver.launch.py
```diff
@@ -91,41 +124,29 @@ def generate_launch_description():
         # DeclareLaunchArgument('force_system', default_value='true',
         #                       description='Set "false" not to load \
         #                                   "libgazebo_ros_force_system.so"'),
+        DeclareLaunchArgument('server_required', default_value='false',
```
j-rivero, line 127: same than with `gui_required`

j-rivero (commented):
Waiting the docker image to compile to give a try at runtime. I did not find anything that needs to be fixed, a couple of comments.

j-rivero:
> Waiting the docker image to compile to give a try at runtime.

The PR changes look good to me and it works as expected with the limitations exposed in the description about not being able to kill both processes in any of the situations (close GUI or kill server).

gazebo_ros/launch/gzclient.launch.py
mabelzhang: I added a bit of description and link in b393d5f

gazebo_ros/launch/gzclient.launch.py
mabelzhang: Yeah... that makes sense. Do you have a suggestion in mind? It's kind of an odd thing to name. I used *_required following the ROS 1 syntax of `<node required>`, assuming somebody who wants this option would know what that is. But to somebody who doesn't, it is more like the meaning you said.

If we want the flags to be the same as the ROS 1 ones, we should change the names in Kinetic, Melodic, Noetic too, since they have just been added and not released (I think?) yet.

chapulina: So... we released melodic today, so I think it's too late to change the parameter :confused:

j-rivero: let's live with it, no big problem. The help is clear and we can document the parameters in the tutorials to mitigate problems with naming.

j-rivero (approved):

PR #1090, opened 2020-04-28
wait for service with variable timeout

Karsten1987:
We've noticed that 5 seconds might be too little to wait for Gazebo to start when the world is very large. We introduce here a new parameter which lets the spawner wait longer for the Gazebo service server to be available.

### DIFF
--- a/gazebo_ros/scripts/spawn_entity.py
+++ b/gazebo_ros/scripts/spawn_entity.py
@@ -77,7 +77,7 @@ class SpawnEntityNode(Node):
         parser.add_argument('-wait', type=str, metavar='ENTITY_NAME',
                             help='Wait for entity to exist')
         parser.add_argument('-spawn_service_timeout', type=float, metavar='TIMEOUT',
-                            default=5.0, help='Spawn service wait timeout')
+                            default=5.0, help='Spawn service wait timeout in seconds')
         parser.add_argument('-x', type=float, default=0,
                             help='x component of initial position, meters')
         parser.add_argument('-y', type=float, default=0,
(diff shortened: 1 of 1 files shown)


gazebo_ros/scripts/spawn_entity.py
```diff
@@ -76,6 +76,8 @@ def __init__(self, args):
                             help='unpause physics after spawning entity')
         parser.add_argument('-wait', type=str, metavar='ENTITY_NAME',
                             help='Wait for entity to exist')
+        parser.add_argument('-spawn_service_timeout', type=float, metavar='TIMEOUT',
+                            default=5.0, help='Spawn service wait timeout')
```
chapulina:
```suggestion
                            default=5.0, help='Spawn service wait timeout in seconds')
```

chapulina (approved):
LGTM, just one nitpick.

Karsten1987:
done

chapulina (approved):

jacobperron:
I don't know what happened, but I thought I added a timeout for this very thing in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1072 and https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1073 :upside_down_face: 

I think we should only have one timeout option, having two is confusing...

jacobperron:
> I don't know what happened

I guess, I made the change on the `ros2` branch and this change landed on `eloquent`, then later forward-ported :thinking:

jacobperron:
I'm proposingl to deprecate the `-spawn_service_timeout` in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1238.

Issue #1079, opened 2020-04-15
Configurable QoS settings for ROS 2 publishers and subscriptions


jacobperron:
## Description

Currently, all of the QoS settings for ROS 2 publishers and subscriptions created in Gazebo plugins are hardcoded (ie. cannot be changed by the user). In particular, many sensor plugins use the "SensorDataQoS" profile introduced in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/926. This QoS profile is generally good for sensor data, using *best effort* for the reliability setting. But, there are cases where applications would like to ensure the arrival of sensor data and use *reliable* reliability.

IMO, gazebo_ros_pkgs should allow for the QoS settings to be entirely configurable so that users can mimic their real ROS 2 drivers accurately (e.g. maybe they have a camera driver that uses the *reliable* setting).

## Implementation considerations

ROS 2 does not officially support a way to configure QoS settings at runtime. Therefore, it is up to the node developer to enable this<sup>1</sup>. In any case, gazebo_ros_pkgs plugins and their parameters are already defined in SDF, so I think it makes sense to add SDF tags for setting various QoS settings for ROS 2 entities in a plugin. I propose introducing a new helper class for parsing QoS SDF tags and returning a `rclcpp::QoS` that can be used by the plugins.

I'm willing to do the work, but I'm looking for some feedback before I go ahead.

---
<sup>1</sup> I think there should be a standard mechanism builtin to ROS 2 to opt-in to configurable QoS settings as a node developer, but this is a separate discussion.

chapulina:
> I think there should be a standard mechanism builtin to ROS 2

+1, do you know if there's an issue ticketed for this, it would be nice to link it here.

> I think it makes sense to add SDF tags for setting various QoS settings for ROS 2 entities in a plugin

Sounds reasonable to me :+1: 

It may be worth it writing down the desired tags before implementing, i.e. `<ros><qos>`... If there's a plan to expose those standardly in ROS 2, as command line args for example, it would be nice to match the terminology so we can migrate more cleanly later.

wjwwood:
Seems like the right thing to do. If you're emulating some sensor and can configure things like rate or noise settings via SDF parameters, it makes sense to also be able to control how you emulate the communication settings.

jacobperron:
> If there's a plan to expose those standardly in ROS 2, as command line args for example, it would be nice to match the terminology so we can migrate more cleanly later

[...]

PR #1091, opened 2020-04-29
Add gazebo_ros::QoS class

jacobperron:
Connects to https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1079

This PR adds a new class that parses SDF for ROS 2 QoS settings that plugins can use when creating publishers and subscriptions.

Here's an example of valid SDF (from one of the tests):

```xml
<?xml version='1.0' ?>
<sdf version='1.6'>
  <world name='default'>
    <plugin name='node_1' filename='libnode_name.so'>
      <ros>
        <qos>
          <topic name="foo">
            <publisher>
              <reliability>best_effort</reliability>
              <durability>transient_local</durability>
              <history>keep_all</history>
              <deadline>123</deadline> <!-- milliseconds -->
              <lifespan>42</lifespan> <!-- milliseconds -->
              <liveliness>manual_by_topic</liveliness>
              <liveliness_lease_duration>1000</liveliness_lease_duration> <!-- milliseconds -->
            </publisher>
            <subscription>
              <reliability>reliable</reliability>
              <durability>volatile</durability>
              <history depth="3">keep_last</history>
            </subscription>
          </topic>
        </qos>
      </ros>
    </plugin>
  </world>
</sdf>
```

The new `<qos>` tag can contain zero or more `<topic>` tags. Each topic tag must have a `name` attribute.

The implementation will look for optional `<publisher>` and `<subscription>` tags, which can contain any number of tags corresponding to all of the QoS settings supported by ROS 2.

A description of QoS settings can be found here:

- http://design.ros2.org/articles/qos.html
- http://design.ros2.org/articles/qos_deadline_liveliness_lifespan.html

One limitation of the current approach is that **all** subscriptions (or publishers) for a given topic will use the same QoS overrides. I don't think there's a way to differentiate them in ROS 2, but I also don't think this will be an issue for any of the plugins in this repository.

The choice to use milliseconds for all duration values is arbitrary. If there is a more preferable unit, it's easy to change.


jacobperron:
Here is a follow-up PR that makes use of the new `gazebo_ros::QoS` object in gazebo_plugins: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1092

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -108,6 +114,11 @@ class Node : public rclcpp::Node
    */
   static rclcpp::Parameter sdf_to_ros_parameter(sdf::ElementPtr const & _sdf);
 
+  inline gazebo_ros::QoS get_qos() const
```
ivanpauno: can be `const &` ?

gazebo_ros/test/test_qos.cpp
```diff
@@ -0,0 +1,264 @@
...
+      "<world name='default'>"
+      "</world>"
+      "</sdf>";
+    sdf::SDF sdf;
+    sdf.SetFromString(sdf_str);
+    gazebo_ros::QoS qos(sdf.Root());
+
+    // We should get the default QoS for an arbitrary topic
+    rclcpp::QoS some_qos = qos.get_publisher_qos("some_random_test_topic");
+    EXPECT_EQ(rclcpp::QoS(10), some_qos);
```
ivanpauno, line 38: I would use https://github.com/ros2/rclcpp/blob/e2dbc5d5d5fbfc9a4b929e0eb73c4d150082e08e/rclcpp/include/rclcpp/qos.hpp#L43 with `rmw_qos_profile_default`, just in case the default profile gets updated.

gazebo_ros/include/gazebo_ros/qos.hpp
```diff
@@ -0,0 +1,166 @@
...
+   * \param[in] topic: Get QoS for subscriptions on this topic name.
+   * \param[in] default_qos: The default quality of service used for settings that have not been
+   *   overridden.
+   */
+  rclcpp::QoS get_subscription_qos(
+    const std::string topic, rclcpp::QoS default_qos = rclcpp::QoS(10)) const;
+
+private:
+  /// Storage container for QoS overrides for a single profile.
+  struct QoSOverrides
```
ivanpauno: why is this needed?
why not just creating a `rclccp::QoS` object from the SDF?

gazebo_ros/src/qos.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+/// Helper function for creating an rclcpp::QoS from SDF and a default QoS as a base.
+QoS::QoSOverrides QoS::get_qos_overrides_from_sdf(sdf::ElementPtr _sdf)
+{
+  QoS::QoSOverrides qos_overrides;
+
+  // Parse 'reliability' QoS
+  if (_sdf->HasElement("reliability")) {
+    auto reliability = _sdf->GetElement("reliability")->Get<std::string>();
+    if ("system" == reliability) {
```
ivanpauno: nit: instead of multiple `if-else` blocks, we could use a map. IMHO, it's easier to read and extend.

gazebo_ros/src/qos.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  // Parse 'liveliness' QoS
+  if (_sdf->HasElement("liveliness")) {
+    auto liveliness = _sdf->GetElement("liveliness")->Get<std::string>();
+    if ("system" == liveliness) {
+      qos_overrides.liveliness = RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT;
+    } else if ("automatic" == liveliness) {
+      qos_overrides.liveliness = RMW_QOS_POLICY_LIVELINESS_AUTOMATIC;
+    } else if ("manual_by_node" == liveliness) {
+      qos_overrides.liveliness = RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE;
```
ivanpauno: This makes me remember that I should have removed that option :joy:

gazebo_ros/src/qos.cpp
```diff
@@ -0,0 +1,219 @@
...
+  if (overrides.liveliness != RMW_QOS_POLICY_LIVELINESS_UNKNOWN) {
+    qos.liveliness(overrides.liveliness);
+  }
+
+  if (overrides.deadline != std::chrono::milliseconds::zero()) {
+    qos.deadline(overrides.deadline);
+  }
+
+  if (overrides.lifespan != std::chrono::milliseconds::zero()) {
+    qos.lifespan(overrides.lifespan);
```
ivanpauno, line 299: `lifespan` is only valid for a `Publisher`, and not for a `Subscription`.
This is actually a problem in `rmw` and `rclcpp`, so I don't mind.

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -84,11 +89,12 @@ class Node : public rclcpp::Node
   /**
    * \details This will call rclcpp::init if it hasn't been called yet.
    * \details Forwards arguments to the constructor for rclcpp::Node
+   * \param[in] qos Quality of service profile for the node entities.
    * \param[in] args List of arguments to pass to <a href="http://docs.ros2.org/latest/api/rclcpp/classrclcpp_1_1_node.html">rclcpp::Node</a>
    * \return A shared pointer to a new #gazebo_ros::Node
    */
   template<typename ... Args>
-  static SharedPtr CreateWithArgs(Args && ... args);
+  static SharedPtr CreateWithArgs(gazebo_ros::QoS & qos, Args && ... args);
```
ivanpauno: Is this parameter needed here?
I think that `static SharedPtr Get(sdf::ElementPtr _sdf);` can directly do all the work, without forwarding a parameter to this function.

gazebo_ros/src/qos.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  if (overrides.liveliness_lease != std::chrono::milliseconds::zero()) {
+    qos.liveliness_lease_duration(overrides.liveliness_lease);
+  }
+
+  return qos;
+}
+
+
+rclcpp::QoS QoS::get_publisher_qos(const std::string topic, rclcpp::QoS default_qos) const
```
ivanpauno, line 310: We will need to make use of this function and `get_subscription_qos` in all plugins, right?

gazebo_ros/test/test_qos.cpp
```diff
@@ -0,0 +1,264 @@
...
+  {
+    auto sdf_str =
+      "<?xml version='1.0' ?>"
+      "<sdf version='1.6'>"
+      "<world name='default'>"
+      "<plugin name='node_1' filename='libnode_name.so'>"
+      "<qos />"
+      "</plugin>"
+      "</world>"
+      "</sdf>";
```
ivanpauno, line 50: nit: I would indent (using spaces in the strings) in these in line `xml` examples.
IMHO, it's more readable with indentation.

ivanpauno (commented):
Great work @jacobperron !!

I've left some minor comments

gazebo_ros/src/qos.cpp
jacobperron: That's right. See the connected PR for usage: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1092

gazebo_ros/test/test_qos.cpp
jacobperron: Yeah, I'll try doing that. I was just following the style of the other tests in `gazebo_ros`.

gazebo_ros/include/gazebo_ros/qos.hpp
jacobperron: The problem is we are overlaying the QoS settings from the SDF onto the default provided by the plugin. In the current design, the default QoS are provided when the plugin calls `get_*_qos()`. I.e. we don't know the default QoS when parsing the SDF for overrides. The other alternative I can think of would be to parse the SDF each time `get_*_qos()` is called, but that seems redundant.

gazebo_ros/src/qos.cpp
jacobperron: Yeah, unless the behaviour changes in `rclcpp`, I don't see what else we can do here. We could log a warning, but maybe that's something best left to `rclcpp`.

gazebo_ros/src/qos.cpp
jacobperron: I don't know why, but should we not include it here then?

gazebo_ros/include/gazebo_ros/node.hpp
jacobperron: Yep, 34b862d

gazebo_ros/test/test_qos.cpp
jacobperron: Done: 4ca784e

gazebo_ros/test/test_qos.cpp
jacobperron: I just explicitly set the default QoS being tested with instead: d85fbd6

gazebo_ros/include/gazebo_ros/node.hpp
jacobperron: You're right. Removed the parameter in da5b314

gazebo_ros/src/qos.cpp
jacobperron: Good idea, see 728bb0e

gazebo_ros/src/qos.cpp
ivanpauno: That feature depended on participants being mapped to nodes, and there was agreement to remove it.
I will create a couple of PRs to actually remove it, but it can already be removed from here.

gazebo_ros/include/gazebo_ros/qos.hpp
ivanpauno: Oh, I see. The current solution looks fine :+1:

gazebo_ros/src/qos.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include <string>
+
+namespace gazebo_ros
+{
+
+/// Helper function for creating an rclcpp::QoS from SDF and a default QoS as a base.
+QoS::QoSOverrides QoS::get_qos_overrides_from_sdf(sdf::ElementPtr _sdf)
+{
+  // Map strings to QoS policies
+  static std::map<std::string, rmw_qos_reliability_policy_t> reliability_map = {
```
ivanpauno: nit^2: prefer `std::unordered_map`

gazebo_ros/src/qos.cpp
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  // Parse 'durability' QoS
+  if (_sdf->HasElement("durability")) {
+    auto durability = _sdf->GetElement("durability")->Get<std::string>();
+    auto find_result = durability_map.find(durability);
+    if (find_result != durability_map.end()) {
+      qos_overrides.durability = find_result->second;
+    } else {
+      std::stringstream oss;
```
ivanpauno: nit^N: prefer `std::ostringstream`.

ivanpauno (approved):

jacobperron:
I don't think the current implementation will work with remapped topic names. I'm going to look into it and add some tests.

ivanpauno:
> I don't think the current implementation will work with remapped topic names. I'm going to look into it and add some tests.

Great point.
I was going to ask that question in my first review, and then I completely forgot :joy:.

IMO, remapping should be done before querying the qos profile.
i.e. I expect a `SDF` with both remapping rules and QoS settings to look like:

```
<ros>
     <remapping>topic:=my_topic</remapping>
</ros>
 <qos>
    <topic name='my_topic'>
        ...
    </topic>
</qos>
```

For that, I think we need to apply remapping rules/topic name expansion in `get_*_qos()`.

P.S.: Shouldn't `qos` tag be a child of `ros` tag?

jacobperron:
I agree with your assessment. I'll take a look at implementing it now.

> P.S.: Shouldn't qos tag be a child of ros tag?

Yes, when the SDF is parsed here in `gazebo_ros::Node`:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/728bb0e05023d78b8e9e9d663ace58d08134f1aa/gazebo_ros/src/node.cpp#L105

The `sdf` is a `<ros>` element.

I'll update the example in this PR description.

For unit testing `gazebo_ros::QoS`, it's not strictly required.

gazebo_ros/src/qos.cpp
jacobperron: updated dc9546f

gazebo_ros/src/qos.cpp
jacobperron: updated a7eaa63

gazebo_ros/src/qos.cpp
jacobperron: Removed in b52f8c2

jacobperron:
@ivanpauno Here's my attempt at supporting remapping: 9e753f5

I don't believe there's a way to get the topic remappings from `rclcpp`, so I'm making the necessary `rcl` function calls.

ivanpauno:
@jacobperron By mistake I did my comments in commit https://github.com/ros-simulation/gazebo_ros_pkgs/commit/9e753f5e6f306e759e2864efc2d4814b1059564e, if you open it you can see the comments in their context.

It would be nice to add a test were a namespace is specified in the `sdf`.
So, it's clear how both expansion and remapping interact with this new feature.

ivanpauno (approved):
LGTM!!

gazebo_ros/include/gazebo_ros/qos.hpp
```diff
@@ -0,0 +1,202 @@
...
+  // The following members are needed for getting topic remappings
+
+  /// The name of the node associated with the QoS settings
+  std::string node_name_;
+
+  /// The namespace of the node associated with the QoS settings
+  std::string node_namespace_;
+
+  /// The options of the node associated with the QoS settings
+  rclcpp::NodeOptions node_options_;
```
scpeters: @chapulina I'm not sure if we're using the PIMPL pattern in `gazebo_ros_pkgs` yet, but I'm wondering if this new class is a good candidate for it, since it has lots of private member variables

jacobperron:
> It would be nice to add a test were a namespace is specified in the sdf.

I've updated the test to include a node namespace: 19469da

ivanpauno (approved):

ivanpauno:
@jacobperron @scpeters is this ready to go in?

jacobperron:
@ivanpauno I think the point about considering to use a pimpl pattern is something we should consider (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1091#discussion_r424761989).

gazebo_ros/include/gazebo_ros/node.hpp
ivanpauno: I was referring to the returned object :joy:.
If the returned value is a `const &`, it's also good to keep the ref qualifier for the method.

jacobperron: Ooops, that's what I meant to do, but wasn't thinking :upside_down_face:

jacobperron: Fixed in 5bcfe93

jacobperron:
I rebased on `ros2` to get changed needed for compilation on Ubuntu Focal (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1093).

gazebo_ros/include/gazebo_ros/qos.hpp
chapulina: Thank you for catching this, @scpeters . Indeed we're using PIMPL for the ROS 2 version of `gazebo_ros_pkgs`, as described on the [Contribution guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-contribution-guide#in-the-code).

ivanpauno:
@scpeters @jacobperron I finally figured out the remaining build problems before closing.
I've introduced some linter failures that I will fix tomorrow, but you can already take a look to the last two commits (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1091/commits/1eb881aca037fd80868f54df31bd89ba4e838332, https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1091/commits/3b8d0f535047883bad91d1d584e787add600fc26).

jacobperron:
The latest changes LGTM :+1:

gazebo_ros/src/node.cpp
```diff
@@ -38,7 +38,7 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
 {
   // Initialize arguments
   std::string name = "";
-  std::string ns = "";
+  std::string ns = "/";
```
chapulina, line 41: Is this change needed?

gazebo_ros/test/test_qos.cpp
```diff
@@ -0,0 +1,292 @@
...
+    expected_sub_qos.lifespan(std::chrono::milliseconds(24));
+    expected_sub_qos.liveliness_lease_duration(std::chrono::milliseconds(1));
+    expected_sub_qos.liveliness(RMW_QOS_POLICY_LIVELINESS_AUTOMATIC);
+
+    auto sdf_str =
+      "<?xml version='1.0' ?>"
+      "<sdf version='1.6'>"
+      "  <world name='default'>"
+      "    <plugin name='node_1' filename='libnode_name.so'>"
+      "      <qos>"
```
chapulina: I know both `<plugin><qos>` and `<plugin><ros><qos>` are supported, but I guess the latter is the recommended one (based on documentation). How about using that on most tests? Just so people looking have a good example to follow.

chapulina (commented):
I haven't tried running it, but I skimmed through the code and it looks good to me.

Besides those 2 comments, I'd also ask if it would be possible to use the new `<qos>` elements in one of the example worlds [here](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/ros2/gazebo_ros/worlds) or [here](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/ros2/gazebo_plugins/worlds), or create a new example world to demonstrate the feature.

gazebo_ros/src/node.cpp
ivanpauno: Valid namespaces should be always absolute (contrary to names, which can be relative/private/absolute).

I guess @jacobperron did that change because the namespace is being passed to the new QoS class, and some of the API it uses may be doing namespace validation.

@jacobperron can you confirm if that was the case?

chapulina: Got it. Since we started going down this road on this PR, do you think we should also do some checking on line 58 in case the user passes an invalid namespace? Or maybe we prepend the `/` for them in case it's missing?

gazebo_ros/test/test_qos.cpp
ivanpauno: Done in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1091/commits/7a5188abb9177bb28ef644d1a2def14fa7193cfa

ivanpauno:
> Besides those 2 comments

I've addressed/replied to the commets.

> I'd also ask if it would be possible to use the new <qos> elements in one of the example worlds here or here, or create a new example world to demonstrate the feature.

I would prefer to do that in a follow up if you don't mind.
Feel free to create a ticket and assign it to me.

gazebo_ros/src/node.cpp
jacobperron: @ivanpauno is exactly right. Without this change, I was seeing runtime errors from some `rcl` / `rclcpp` API calls (though I can't recall exactly which ones).

I guess if a user passes an invalid namespace, they will see similar runtime errors. I think the validation and error message from the underlying layers is enough (it also is nice to keep it encapsulated there in case the naming requirement changes for some reason).

chapulina: Got it, makes sense. I'm just wondering why we haven't seen problems so far with this default value, and whether this change is backwards compatible. But you peeps know best about these details :wink:

jacobperron: It's a good question. It looks like the [node initialization logic handles an empty string specifically](https://github.com/ros2/rcl/blob/85ef95514d6ee47bb8285009add84e956595c018/rcl/src/rcl/node.c#L173-L176), but the function for [expanding topic names](https://github.com/ros2/rcl/blob/85ef95514d6ee47bb8285009add84e956595c018/rcl/src/rcl/expand_topic_name.c#L44) does not. It's possible the latter function could be updated in the future to handle this case, but at the end of the day and empty string will resolve to `"/"`.

ivanpauno:
@chapulina is this ready to go in?

chapulina:
> @chapulina is this ready to go in?

I haven't tried it out but looking at the code I don't see anything weird. Could you peeps just run CI before merging? Instructions here: https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-CI#next-turtle

ivanpauno:
> I haven't tried it out but looking at the code I don't see anything weird. Could you peeps just run CI before merging? Instructions here: https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-CI#next-turtle

Thanks for the instructions! I didn't know how to run CI here.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10868/)

chapulina (approved):
Thanks for triggering CI. I see all QoS tests passed. I don't think the flake8 failures are this PR's fault.

All other tests failing with:

`[Err] [Master.cc:95] EXCEPTION: Unable to start server[bind: Address already in use]. There is probably another Gazebo process running.`

are being caused by some earlier test not being properly terminated and leaving Gazebo running. I think it's because of `test_entity_spawner.test.py`, see the testing adventure here: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1033#issuecomment-577363549. Again, not this PR's fault.

ivanpauno:
> Thanks for triggering CI. I see all QoS tests passed. I don't think the flake8 failures are this PR's fault.

I saw the flake8 failures locally, and they are also reproducible in `ros2` branch.

> All other tests failing with:
>
> [Err] [Master.cc:95] EXCEPTION: Unable to start server[bind: Address already in use]. There is probably another Gazebo process running.
> 
> are being caused by some earlier test not being properly terminated and leaving Gazebo running. > I think it's because of test_entity_spawner.test.py, see the testing adventure here: #1033 (comment). Again, not this PR's fault.

I experienced that locally too :joy:.
Killing the gazebo server running in the background before re-running the tests solved the issue (we should probably have a script in CI machines to kill it).

----

@chapulina I don't have write access here, please merge if it's ready.
Thanks!!

jacobperron:
I can rebase the connected PR.

chapulina:
> we should probably have a script in CI machines to kill it

Yeah I think this could be built into the test cases themselves to kill any running `gzserver`s in between tests. The downside is if someone is running another simulation on the same machine and that gets killed by mistake :sweat_smile:

gazebo_ros/src/node.cpp
chapulina: I think this may have messed something up :thinking: https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1116

ivanpauno: Doesn't seem related to me.
That default value is being immediately overridden by the value in the SDF if it exists https://github.com/ros-simulation/gazebo_ros_pkgs/blob/57732fe150efcb324e905f0f68537888c1eceb30/gazebo_ros/src/node.cpp#L57-L59.

ivanpauno: Maybe, it's related to this https://github.com/ros-simulation/gazebo_ros_pkgs/blob/57732fe150efcb324e905f0f68537888c1eceb30/gazebo_ros/src/gazebo_ros_factory.cpp#L231-L235
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/57732fe150efcb324e905f0f68537888c1eceb30/gazebo_ros/src/gazebo_ros_factory.cpp#L354-L355
?

According to grep that's the only place where we set the `namespace` SDF tag.

chapulina: Nice catch! @ahcorde , are you spawning your robot through the factory? In my case I was.

ahcorde: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1117

srushtibobade:
@jacobperron 
Thanks for all the good work!!
 where should this XML file be located/ or added to the project?

ivanpauno:
this can just be added in a normal gazebo sdf world file, the new qos parameters can be applied to gazebo_ros_pkgs plugins as shown in the first post https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1091#issue-608735085

PR #1092, opened 2020-04-29
Make QoS for publishers and subscriptions configurable 

jacobperron:
Closes https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1079

This PR builds on top of https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1091

Whenever a plugin creates a ROS publisher or subscription, use the QoS profile provided by the node for the given topic.
This enables users to override the QoS settings in SDF.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
+++ b/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
@@ -210,7 +210,7 @@ void GazeboRosAckermannDrive::Load(gazebo::physics::ModelPtr _model, sdf::Elemen
   impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
 
   // Get QoS profiles
-  const gazebo_ros::QoS qos = impl_->ros_node_->get_qos();
+  const gazebo_ros::QoS & qos = impl_->ros_node_->get_qos();
 
   impl_->joints_.resize(7);
 
--- a/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
+++ b/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
@@ -209,6 +209,9 @@ void GazeboRosAckermannDrive::Load(gazebo::physics::ModelPtr _model, sdf::Elemen
   // Initialize ROS node
   impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
 
+  // Get QoS profiles
+  const gazebo_ros::QoS qos = impl_->ros_node_->get_qos();
+
   impl_->joints_.resize(7);
 
   auto steering_wheel_joint =
@@ -355,7 +358,7 @@ void GazeboRosAckermannDrive::Load(gazebo::physics::ModelPtr _model, sdf::Elemen
   impl_->last_update_time_ = _model->GetWorld()->SimTime();
 
   impl_->cmd_vel_sub_ = impl_->ros_node_->create_subscription<geometry_msgs::msg::Twist>(
-    "cmd_vel", rclcpp::QoS(rclcpp::KeepLast(1)),
+    "cmd_vel", qos.get_subscription_qos("cmd_vel", rclcpp::QoS(1)),
     std::bind(&GazeboRosAckermannDrivePrivate::OnCmdVel, impl_.get(), std::placeholders::_1));
 
   RCLCPP_INFO(
@@ -369,7 +372,7 @@ void GazeboRosAckermannDrive::Load(gazebo::physics::ModelPtr _model, sdf::Elemen
   impl_->publish_odom_ = _sdf->Get<bool>("publish_odom", false).first;
   if (impl_->publish_odom_) {
     impl_->odometry_pub_ = impl_->ros_node_->create_publisher<nav_msgs::msg::Odometry>(
-      "odom", rclcpp::QoS(rclcpp::KeepLast(1)));
+      "odom", qos.get_publisher_qos("odom", rclcpp::QoS(1)));
 
     RCLCPP_INFO(
       impl_->ros_node_->get_logger(), "Advertise odometry on [%s]",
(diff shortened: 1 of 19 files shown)


ivanpauno:
@jacobperron is this ready to be reviewed?

jacobperron:
> @jacobperron is this ready to be reviewed

I planned to rebase after #1091 is merged, but you can take a look at the last commit (8f7e1030dff186682ab3e383f6e022cbf1475da7), which is the actual change being proposed here (and I don't expect it to change).

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
```diff
@@ -209,6 +209,9 @@ void GazeboRosAckermannDrive::Load(gazebo::physics::ModelPtr _model, sdf::Elemen
   // Initialize ROS node
   impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
 
+  // Get QoS profiles
+  const gazebo_ros::QoS qos = impl_->ros_node_->get_qos();
```
ivanpauno: If the comment here is addressed https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1091#discussion_r427531640, it would be nice to use `const &` here too.

Same elsewhere.

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
```diff
@@ -355,7 +358,7 @@ void GazeboRosAckermannDrive::Load(gazebo::physics::ModelPtr _model, sdf::Elemen
   impl_->last_update_time_ = _model->GetWorld()->SimTime();
 
   impl_->cmd_vel_sub_ = impl_->ros_node_->create_subscription<geometry_msgs::msg::Twist>(
-    "cmd_vel", rclcpp::QoS(rclcpp::KeepLast(1)),
+    "cmd_vel", qos.get_subscription_qos("cmd_vel", rclcpp::QoS(1)),
```
ivanpauno, line 361: It would be nice to avoid the repetition of the topic name here.
I'm not sure what is the right way of doing that, is overriding `create_publisher` and `create_subscription` in `gazebo_ros::Node` acceptable?

If not, `get_subscription_qos` could return a `struct` composed by the `rclcpp::QoS` object and the topic name.
The final code would look like:

```cpp
  auto result = qos.get_subscription_qos("cmd_vel", rclcpp::QoS(1));
  impl_->cmd_vel_sub_ = impl_->ros_node_->create_subscription<geometry_msgs::msg::Twist>(
    result.topic_name, result.qos,
    std::bind(&GazeboRosAckermannDrivePrivate::OnCmdVel, impl_.get(), std::placeholders::_1));
```

I don't have a strong opinion about this, the repetition isn't too bad either.

jacobperron: I thought about overriding `create_publisher` and `create_subscription`, but I thought it seemed a little invasive and could take away from the readability of the code. I think it's clear what's happening when we get the QoS from the `gazebo_ros::QoS` object compared to hiding it in a call to `create_publisher`.

On the other hand, maybe it's fine to hide this detail from the plugin developer. Then all plugins will have configurable QoS without needing to opt-in.

ivanpauno: > I thought about overriding create_publisher and create_subscription, but I thought it seemed a little invasive and could take away from the readability of the code. I think it's clear what's happening when we get the QoS from the gazebo_ros::QoS object compared to hiding it in a call to create_publisher.

Yes, I have a similar feeling about overriding.
I won't block on this, you can decide what you think is best.

---

Besides my strange idea of modifying `get_subscription_qos` return value, another option is:

```cpp
  std::string topic_name = "cmd_vel";
  impl_->cmd_vel_sub_ = impl_->ros_node_->create_subscription<geometry_msgs::msg::Twist>(
    topic_name, qos.get_subscription_qos(topic_name, rclcpp::QoS(1)),
    std::bind(&GazeboRosAckermannDrivePrivate::OnCmdVel, impl_.get(), std::placeholders::_1));
```

There's no great benefit either.

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
jacobperron: See 32a0b52

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
jacobperron: I'd like to try overriding `create_publisher` and `create_subscription` to see what it looks like; whenever I get a moment. Feel free to propose the change yourself if you have time (either updating this branch or targeting a PR against it).

ivanpauno (approved):
LGTM!

gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
ivanpauno: We also have things like `image_transport::create_publisher`, and I'm not sure if overriding `Node` methods will be enough (maybe, we're passing the NodeBase interface ....).

I will prefer to merge this as-is, and we can do in improvement later.
What do you think?

jacobperron: SGTM

jacobperron:
Linux: [Build Status](https://ci.ros2.org/job/ci_linux/10886/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10892/)

ivanpauno:
The test failures are:
- unrelated flake8
- unrelated cppcheck (I can't reproduce them locally)
- failures because of a gazebo server running in the background ....

@chapulina do you think this is ready to go?
I will create a PR to fix the unrelated flake8 failures.

ivanpauno:
> I will create a PR to fix the unrelated flake8 failures.

Opened https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1110.

chapulina (approved):
Nice!

PR #1096, opened 2020-05-01
[noetic-devel] Set valid marker orientation and remove debug msg

iche033:
When values of normals are zero, i.e., no models are in the view of the camera, the marker orientation becomes `nan` and causes error messages in rviz. This PR sets it to identity for this case.

Also removed a debug msg


ahcorde (approved):
LGTM

chapulina (approved):

PR #1099, opened 2020-05-07
Port wheel slip plugin to ros2 (eloquent)

scpeters:
This ports the wheel slip plugin from melodic (#995) to ros2. It uses ros2 parameters instead of dynamic reconfigure.

An example world is provided that shows two `trisphere_cycle` models that each have two instances of the wheel slip plugin, allowing separate parameters for the front and rear wheels. The demo world has some example `ros2 param` commands to adjusting the parameters dynamically.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
+++ b/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
@@ -45,6 +45,46 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr 
 {
   // Initialize ROS node
   impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
+
+  impl_->ros_node_->declare_parameter("slip_compliance_unitless_lateral", 0.0);
+  impl_->ros_node_->declare_parameter("slip_compliance_unitless_longitudinal", 0.0);
+
+  auto existing_callback = impl_->ros_node_->set_on_parameters_set_callback(nullptr);
+  auto param_change_callback =
+    [this, existing_callback](std::vector<rclcpp::Parameter> parameters) {
+      auto result = rcl_interfaces::msg::SetParametersResult();
+      if (nullptr != existing_callback) {
+        result = existing_callback(parameters);
+        if (!result.successful) {
+          return result;
+        }
+      }
+
+      result.successful = true;
+      for (const auto & parameter : parameters) {
+        std::string param_name = parameter.get_name();
+        if (param_name == "slip_compliance_unitless_lateral") {
+          RCLCPP_INFO(impl_->ros_node_->get_logger(),
+            "New lateral slip compliance: %.3e", parameter.as_double());
+          this->SetSlipComplianceLateral(parameter.as_double());
+        } else if (param_name == "slip_compliance_unitless_longitudinal") {
+          RCLCPP_INFO(impl_->ros_node_->get_logger(),
+            "New longitudinal slip compliance: %.3e", parameter.as_double());
+          this->SetSlipComplianceLongitudinal(parameter.as_double());
+        } else {
+          RCLCPP_WARN(impl_->ros_node_->get_logger(),
+            "Unrecognized parameter name[%s]", param_name);
+          result.successful = false;
+        }
+      }
+      return result;
+    };
+
+  impl_->ros_node_->set_on_parameters_set_callback(param_change_callback);
+
+  // Initialize the WheelSlipPlugin after the ros node so it doesn't overwrite
+  // the sdf parameters.
+  WheelSlipPlugin::Load(_model, _sdf);
 }
 
 GZ_REGISTER_MODEL_PLUGIN(GazeboRosWheelSlip)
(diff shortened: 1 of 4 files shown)


scpeters:
I may need to rework this a bit to match the changes in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1111

scpeters:
@jacobperron considering the change in #1126 in noetic that attempts to use the parameter value specified in the launch file, what would the equivalent be for ros2? I'm guessing this PR might need a change to match that behavior

scpeters:
I can make a new branch and target to foxy as well

chapulina:
> I can make a new branch and target to foxy as well

Note that technically Foxy has already been released (bloomed but not synced yet, still in [ros2-testing](http://packages.ros.org/ros2-testing/ubuntu/pool/main/r/ros-foxy-gazebo-plugins/)). So if you really want to break behaviour on Foxy, I'd recommend doing it soon. Otherwise, targeting Galactic.

scpeters:
> Note that technically Foxy has already been released (bloomed but not synced yet, still in ros2-testing). So if you really want to break behaviour on Foxy, I'd recommend doing it soon. Otherwise, targeting Galactic.

This is adding a new plugin, so I don't think it breaks anything

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr _sdf)
+{
+  // Initialize ROS node
+  impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
+
+  impl_->ros_node_->declare_parameter("slip_compliance_unitless_lateral", 0.0);
+  impl_->ros_node_->declare_parameter("slip_compliance_unitless_longitudinal", 0.0);
+
+  auto existing_callback = impl_->ros_node_->set_on_parameters_set_callback(nullptr);
```
jacobperron: `set_on_parameters_set_callback` has been deprecated in Foxy. I recommend using [add_on_set_parameters_callback()](http://docs.ros2.org/eloquent/api/rclcpp/classrclcpp_1_1Node.html#a0c3ad2ca5c56eb54af0227f04e943679) as of Eloquent to avoid future deprecation warnings. This should also simplify the logic a little bit since we don't have to worry about handling any existing callbacks.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -0,0 +1,91 @@
...
+{
+}
+
+void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr _sdf)
+{
+  // Initialize ROS node
+  impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
+
+  impl_->ros_node_->declare_parameter("slip_compliance_unitless_lateral", 0.0);
+  impl_->ros_node_->declare_parameter("slip_compliance_unitless_longitudinal", 0.0);
```
jacobperron: If you declare parameters *after* registering the parameters callback, then the callback will triggered when calling `declare_parameter`. Then if the defaults change in `declare_parameters`, the plugins values will also be updated.

jacobperron:
> @jacobperron considering the change in #1126 in noetic that attempts to use the parameter value specified in the launch file, what would the equivalent be for ros2? I'm guessing this PR might need a change to match that behavior

I don't think we have the same problem in ROS 2, since native parameters are reconfigurable. A parameter passed to the node in a launch file should override the default value in `declare_parameter`.

scpeters:
I rebased and changed the behavior in 00257a609f0de78606dafe0671b852e69cf15f64 to match the change in #1111 / #1126.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
scpeters: I've switched to this API in 00257a6

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
scpeters: ok, I changed the order of these function calls in 563c621

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -0,0 +1,93 @@
...
+        } else {
+          RCLCPP_WARN(impl_->ros_node_->get_logger(),
+            "Unrecognized parameter name[%s]", param_name);
+          result.successful = false;
+        }
+      }
+      return result;
+    };
+
+  impl_->ros_node_->add_on_set_parameters_callback(param_change_callback);
```
jacobperron: You might want to double-check, but I think we should maintain a reference to the returned callback handle, otherwise the callback may be unregistered.

scpeters: yeah, it looks like this was done in https://github.com/ros2/rclcpp/pull/1136

scpeters: I copied the approach from disparity_node in stereo_image_proc: https://github.com/ros-perception/image_pipeline/blob/2.2.0/stereo_image_proc/src/stereo_image_proc/disparity_node.cpp#L92-L182 in 0bef26d

jacobperron (approved):

scpeters:

scpeters:

nuclearsandwich:

scpeters:
some unrelated test failures, but CI is finally working

Issue #323, opened 2015-05-01
gazebo_ros has undeclared run dependency on libgazebo5-dev


scpeters:
The jade version of `gazebo_ros` declares a [build dependency on libgazebo5-dev](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_ros/package.xml#L27) and a [run dependency on gazebo5](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_ros/package.xml#L40).

However, the [gazebo_ros/scripts](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/jade-devel/gazebo_ros/scripts) are [using pkg-config](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_ros/scripts/gzserver#L21) to find the prefix for the gazebo setup.sh file. This implies a run dependency on the gazebo.pc file in `libgazebo5-dev`.

If you don't have `libgazebo5-dev` installed, this will cause simple roslaunch's of gazebo to fail with something like the following error message:

```
$ roslaunch gazebo_ros empty_world.launch
...
/opt/ros/jade/lib/gazebo_ros/gzclient: 17: .: Can't open /share/gazebo//setup.sh
/opt/ros/jade/lib/gazebo_ros/gzserver: 22: .: Can't open /share/gazebo//setup.sh
[gazebo-2] process has died [pid 15709, exit code 2, cmd /opt/ros/jade/lib/gazebo_ros/gzserver -e ode worlds/empty.world __name:=gazebo __log:=~/.ros/log/bb3dc582-f051-11e4-96c1-902b3456ea33/gazebo-2.log].
log file: ~/.ros/log/bb3dc582-f051-11e4-96c1-902b3456ea33/gazebo-2*.log
[gazebo_gui-3] process has died [pid 15719, exit code 2, cmd /opt/ros/jade/lib/gazebo_ros/gzclient __name:=gazebo_gui __log:=~/.ros/log/bb3dc582-f051-11e4-96c1-902b3456ea33/gazebo_gui-3.log].
log file: ~/.ros/log/bb3dc582-f051-11e4-96c1-902b3456ea33/gazebo_gui-3*.log
```

Installing `libgazebo5-dev` is currently a valid workaround.

@wjwwood

cmaestre:
[...]

PR #1100, opened 2020-05-07
Use fallback instead of pkg-config for gazebo discovery

simonschmeisser:
Currently `gazebo_ros` depends on `gazebo_dev` which depends on `libgazebo11-dev` in order to use the pkg-config file included there to locate the `setup.sh` (see #323 )

`libgazebo11-dev` however also depends on `libogre-1.9-dev` which conflicts with `libogre-1.12-dev` which `rviz` would like to use for focal (see https://github.com/ros/rosdistro/pull/24448 for discussion)

This PR adds a fallback to `/usr/share/gazebo` as this is the location used in the packages and therefore allows to drop the runtime dependency on `gazebo_dev`

### DIFF
--- a/gazebo_ros/package.xml
+++ b/gazebo_ros/package.xml
@@ -23,11 +23,8 @@
 
   <build_depend>cmake_modules</build_depend>
   <build_depend>gazebo_dev</build_depend>
-  <!--
-    Need to use gazebo_dev since run script needs pkg-config
-    See: https://github.com/ros-simulation/gazebo_ros_pkgs/issues/323 for more info
-  -->
-  <exec_depend>gazebo_dev</exec_depend>
+
+  <exec_depend>gazebo</exec_depend>
   <depend>gazebo_msgs</depend>
   <depend>roslib</depend>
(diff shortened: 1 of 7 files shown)


chapulina:
> libgazebo11-dev however also depends on libogre-1.9-dev which conflicts with libogre-1.12-dev which rviz would like to use for focal 

That sounds like a no-go to me. We want users to be able to develop Gazebo and RViz plugins side-by-side on Noetic, even if each individual plugin is not compiled against both.

simonschmeisser:
Nobody stops you from doing that, it just requires compiling either gazebo or rviz from source with either libogre-1.12-dev or libogre-1.9-dev installed. rviz will stay compatible with ogre-1.9 as only that is included in Debian Buster.

chapulina:
> it just requires compiling either gazebo or rviz from source

That's already possible if someone really wants to use RViz with 1.12, right? I believe we're talking about the official versions that will be released into `rosdistro`. And I believe they shouldn't conflict so users have a better out-of-box experience.

PR #1110, opened 2020-05-27
Fix flake8 failures

ivanpauno:
:point_up:

According to [ros2 style guide](https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#id4), "hanging indents" are preferred over "continuation lines". So, I also enforced that style (which flake8 doesn't enforce).

### DIFF
--- a/gazebo_ros/launch/gzclient.launch.py
+++ b/gazebo_ros/launch/gzclient.launch.py
@@ -48,40 +48,48 @@ def generate_launch_description():
     env = {
         'GAZEBO_MODEL_PATH': model,
         'GAZEBO_PLUGIN_PATH': plugin,
-        'GAZEBO_RESOURCE_PATH': media}
+        'GAZEBO_RESOURCE_PATH': media,
+    }
 
     prefix = PythonExpression([
         '"gdb -ex run --args" if "true" == "',
         LaunchConfiguration('gdb'),
         '"else "valgrind" if "true" == "',
         LaunchConfiguration('valgrind'),
-        '"else ""'
+        '"else ""',
     ])
 
     return LaunchDescription([
         DeclareLaunchArgument(
             'version', default_value='false',
-            description='Set "true" to output version information'),
+            description='Set "true" to output version information'
+        ),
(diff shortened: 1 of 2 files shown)


gazebo_ros/launch/gzclient.launch.py
```diff
@@ -44,33 +45,43 @@ def generate_launch_description():
...
-
-    prefix = PythonExpression(['"gdb -ex run --args" if "true" == "',
-                                     LaunchConfiguration('gdb'),
-                                     '"else "valgrind" if "true" == "',
-                                     LaunchConfiguration('valgrind'),
-                                     '"else ""']),
+    env = {
+        'GAZEBO_MODEL_PATH': model,
+        'GAZEBO_PLUGIN_PATH': plugin,
+        'GAZEBO_RESOURCE_PATH': media}
```
chapulina:
```suggestion
        'GAZEBO_RESOURCE_PATH': media
    }
```

chapulina (approved):
Thanks for the cleanup! I just have one suggestion. Also, I wonder if all the `)` should also be on a new line. The guidelines say that "_The closing brace/bracket/parenthesis on multiline constructs may either line up under the first non-whitespace character of the last line of list._"

It would be great if the linter would catch these.

gazebo_ros/launch/gzclient.launch.py
ivanpauno: Done in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1110/commits/c18079fd9218ad93718fe38e0b909686e5d81a4d.

ivanpauno:
> Thanks for the cleanup! I just have one suggestion. Also, I wonder if all the `)` should also be on a new line. The guidelines say that "_The closing brace/bracket/parenthesis on multiline constructs may either line up under the first non-whitespace character of the last line of list._"

I moved all closing braces/parentheses to a new line in multi-line statements.
 
> It would be great if the linter would catch these.

I'm not sure if flake8 can be configured ....

PR #1111, opened 2020-05-27
gazebo_ros_wheel_slip: support parameters specified in launch file (backport to melodic)

scpeters:
The `gazebo_ros_wheel_slip` plugin currently ignores slip compliance values specified in a  launch file due to the check of the `level` parameter added in 3cf9f5e1bdf424d5b1866e74f3bde2e904c74298, which ensures that parameter values specified in the urdf / sdf xml parameters are not over-ridden by the default values specified in the `.cfg` file. In order to respect the values set in launch files but ignore the defaults in the `.cfg` file, change the `.cfg` default values to `-1` and adjust the callback to ignore negative values. This is fine because gazebo's WheelSlipPlugin already has default values for the slip compliance parameters, so the `.cfg` defaults are extraneous.

This also reverts 3cf9f5e1bdf424d5b1866e74f3bde2e904c74298 so that the `level` parameter is ignored again.

cc @maxx84

### DIFF
--- a/gazebo_plugins/cfg/WheelSlip.cfg
+++ b/gazebo_plugins/cfg/WheelSlip.cfg
@@ -9,7 +9,7 @@ from dynamic_reconfigure.parameter_generator_catkin import *
 gen = ParameterGenerator()
 
 #       Name                                      Type      Reconf level   Description                                                                  Default    Min   Max
-gen.add("slip_compliance_unitless_lateral",       double_t, 0,             "Unitless slip compliance (slip / friction) in the lateral direction.",      0.0,       0.0,  20.0)
-gen.add("slip_compliance_unitless_longitudinal",  double_t, 0,             "Unitless slip compliance (slip / friction) in the longitudinal direction.", 0.0,       0.0,  20.0)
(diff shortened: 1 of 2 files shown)


Maxx84:
Both the issue and this fix have been replicated. 

Behavior before this fix:
- The plugin loaded the parameters (A) specified in the urdf/sdf, even if different values (B) are set in the launch file. Although the dynamic parameters appear to be set with the values (B) specified in the launch file, the robot clearly behaves according to the parameter values (A) specified in the urdf/sdf. If new dynamic parameters (C) are manually set during runtime, the robot starts behaving according to these new parameters (C).

Behavior with this fix:
- The plugin loads the parameters (B) specified in the launch file, if available. The dynamic parameters are set with the correct values (B) and the robot behaves in the expected way.
- If no parameters are set in the launch file, then the plugin loads the parameter values (A) set in the urdf/sdf. These values (A) however are not loaded in the plugin dynamic parameters, which are just set to zero. Nevertheless, the robot behaves correctly, according to the parameters (A). If new dynamic parameters (C) are manually set during runtime, the robot starts behaving according to these new parameters (C).

scpeters:
I've created #1126 that targets this to noetic-devel. We will merge that first and then revisit this as a backport to melodic if all is well.

scpeters:
I just merged #1126 to noetic

jacobperron (approved):

Issue #1116, opened 2020-06-08
[ROS 2] gazebo_ros returns always an empty namespace


ahcorde:
I was trying to add a namespace using the `node.cpp` class. In particular I have some issue with this [line](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_ros/src/node.cpp#L58). if I change the name of the xml tag to `name_space` or others I can get the value, but using the word namespace the code is returning always "/".

Arguments, parameters and remapping are working fine

```
<plugin ...>
            <ros>
              <namespace>gazebo_ros2_control</namespace>
            </ros>
</plugin>
```

@chapulina any ideas?

ahcorde:
This is the `sdf->PrintValues()` in the first line of the `Load` function (entry point for the plugin).For some reason it's overriding the `<namespace>` tag inside `<ros>`

```
[gzserver-1] <plugin name='gazebo_ros2_control' filename='libgazebo_ros2_control.so'>
[gzserver-1]   <ros>
[gzserver-1]     <namespace>/</namespace>
[gzserver-1]     <namespace2>test_cam_16bit</namespace2>
[gzserver-1]     <remapping>test_camera_name/image_raw:=image_test_16bit</remapping>
[gzserver-1]     <remapping>camera_info:=camera_info_test_16bit</remapping>
[gzserver-1]     <argument>from:=to</argument>
[gzserver-1]   </ros>
```

chapulina:
I also ran into this issue [yesterday](https://github.com/chapulina/dolly/commit/f9b2e078ce1b645be558708b3060a4cd4f1c57d7#diff-3ea051f5bf6c350c7601fe2d685f31c7R334) using `eloquent` + `gazebo9` debs. I'm sure it used to work. I have no clue why it would stop working now.

ahcorde:
Fixed in this PR https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1117

PR #804, opened 2018-08-24
Remove node_name from <ros> SDF tag

dhood:
In progress

PR #1117, opened 2020-06-09
[ROS 2] Use "" as default in spawn_entity.py instead of self.get_namespace().

ahcorde:
Targeting this issue https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1116

if the <namespace> tag is defined then we should not override the value

### DIFF
--- a/gazebo_ros/src/gazebo_ros_factory.cpp
+++ b/gazebo_ros/src/gazebo_ros_factory.cpp
@@ -230,7 +230,7 @@ void GazeboRosFactoryPrivate::SpawnEntity(
 
   // Walk recursively through the entire file, and add <ros><namespace> tags to all plugins
   auto robot_namespace = req->robot_namespace;
-  if (!robot_namespace.empty()) {
+  if (robot_namespace.compare("/") != 0) {
     AddNamespace(entity_elem, robot_namespace);
   }
 
@@ -349,9 +349,10 @@ void GazeboRosFactoryPrivate::AddNamespace(
 
         // Then we can add the element
         ns_elem = ros_elem->AddElement("namespace");
-        // Set namespace
-        ns_elem->Set<std::string>(_robot_namespace);
       }
+
+      // Set namespace
+      ns_elem->Set<std::string>(_robot_namespace);
     }
     AddNamespace(child_elem, _robot_namespace);
   }
--- a/gazebo_ros/src/gazebo_ros_factory.cpp
+++ b/gazebo_ros/src/gazebo_ros_factory.cpp
@@ -230,7 +230,7 @@ void GazeboRosFactoryPrivate::SpawnEntity(
 
   // Walk recursively through the entire file, and add <ros><namespace> tags to all plugins
   auto robot_namespace = req->robot_namespace;
-  if (robot_namespace.compare("/") != 0) {
+  if (!robot_namespace.empty()) {
     AddNamespace(entity_elem, robot_namespace);
   }
 
--- a/gazebo_ros/src/gazebo_ros_factory.cpp
+++ b/gazebo_ros/src/gazebo_ros_factory.cpp
@@ -349,10 +349,9 @@ void GazeboRosFactoryPrivate::AddNamespace(
 
         // Then we can add the element
         ns_elem = ros_elem->AddElement("namespace");
+        // Set namespace
+        ns_elem->Set<std::string>(_robot_namespace);
       }
-
-      // Set namespace
-      ns_elem->Set<std::string>(_robot_namespace);
     }
     AddNamespace(child_elem, _robot_namespace);
   }
--- a/gazebo_ros/scripts/spawn_entity.py
+++ b/gazebo_ros/scripts/spawn_entity.py
@@ -70,7 +70,7 @@ class SpawnEntityNode(Node):
         parser.add_argument('-gazebo_namespace', type=str, default='',
                             help='ROS namespace of gazebo offered ROS interfaces. \
                             Default is without any namespace')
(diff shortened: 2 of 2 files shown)


ivanpauno (commented):
I think we support both using `<plugin><ros><namespace>...` and `<plugin><namespace>...`, right @chapulina ?
In that case, the logic doesn't seem to be completely correct.

Otherwise LGTM!

chapulina:
>  we support both using <plugin><ros><namespace>... and <plugin><namespace>

I believe you're right, but I don't think that was ever intentional. I'm trying to remember how we ended up in this situation, and I think that originally the `Node::Create` (now `Node::Get`) function expected to receive the `<ros>` element:

https://github.com/ros-simulation/gazebo_ros_pkgs/pull/781/files#diff-3398c1d107dfecdcaa399f78e78ad988

Then it was extended to also accept the parent `<plugin>` element:

https://github.com/ros-simulation/gazebo_ros_pkgs/pull/804/files#diff-d3eeabf9745d59cc4a9cf2dee81e16ba

And in the process we ended up supporting all of `<ros>`'s contents directly inside `<plugin>`.

So I think that we shouldn't break backwards compatibility and support both. But it may be good to deprecate the `<plugin>` top-level support from G-turtle.

gazebo_ros/src/gazebo_ros_factory.cpp
```diff
@@ -349,10 +349,9 @@ void GazeboRosFactoryPrivate::AddNamespace(
 
         // Then we can add the element
         ns_elem = ros_elem->AddElement("namespace");
+        // Set namespace
+        ns_elem->Set<std::string>(_robot_namespace);
       }
-
-      // Set namespace
-      ns_elem->Set<std::string>(_robot_namespace);
```
chapulina: If the user sets a `robot_namespace` on the factory request, we do want to override the values within the SDF file. So I think this was correct originally. And we should already be skipping it if the user left it empty:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/57732fe150efcb324e905f0f68537888c1eceb30/gazebo_ros/src/gazebo_ros_factory.cpp#L231-L235

ahcorde: The following condition will never be false. Because the deafult namespace is '/'

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/57732fe150efcb324e905f0f68537888c1eceb30/gazebo_ros/src/gazebo_ros_factory.cpp#L233

Here is where the default value is set 

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/57732fe150efcb324e905f0f68537888c1eceb30/gazebo_ros/scripts/spawn_entity.py#L73-L74

ahcorde: Changed the comparison to :

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/f154399e4462fda3dc4d06e842ffe280873c3043/gazebo_ros/src/gazebo_ros_factory.cpp#L233-L235

ivanpauno: I don't know why the namespace of the spawn entity node is being used as the robot namespace, it sounds weird to me.
I would use `""` as the default.

Comparing with `"/"` is not a good idea, as that is a valid namespace (`""` is not a valid namespace).

chapulina: > I don't know why the namespace of the spawn entity node is being used as the robot namespace

What we call "robot namespace" is just the ROS namespace that will be added to all topics and services for that plugin. So if you want to spawn multiple robots with diff drive controllers, each can be controlled in a different topic (i.e. `/robot1/cmd_vel`, `/robot2/cmd_vel`, etc).

> Here is where the default value is set

Nice catch, I was really puzzled about how the `/` had leaked into the request.

---

@ivanpauno / @jacobperron , going back to the conversation in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1091#discussion_r428914423, can we just set the default back to empty (invalid) and let lower layers handle the issue in case an empty namespace goes through?

ivanpauno: I'm not sure how that codes affects the spawn entity node namespace:

```cpp
  std::string ns = "/";
  ...
  // Set namespace if tag is present
  if (sdf->HasElement("namespace")) {
    ns = sdf->GetElement("namespace")->Get<std::string>();
  }
```

`self.get_namespace()` in `spawn_entity.py` will continue returning "/" besides changing `ns` to "" in the above snippet. Maybe I'm not seeing a weird interaction between both things, in that case modifying the above snippet sounds ok.

> What we call "robot namespace" is just the ROS namespace that will be added to all topics and services for that plugin. So if you want to spawn multiple robots with diff drive controllers, each can be controlled in a different topic (i.e. /robot1/cmd_vel, /robot2/cmd_vel, etc).

That sounds good, but I don't know why the namespace of the spawner should be used as the default namespace of the spawned robot.
I would use `""` as default in `spawn_entity.py` instead of `self.get_namespace()`.
I don't have much gazebo experience, so maybe I'm confusing things.

chapulina: > I would use "" as default in spawn_entity.py instead of self.get_namespace().

Good point, maybe that's all that's needed

ahcorde: Fixed and updated the title of the PR

ivanpauno (approved):
lgtm

chapulina (approved):

PR #1118, opened 2020-06-09
Depth only

anaelle-sw:
In order to get the output structure from gazebo_ros_camera plugin match our realsense d435 output, we added a `<depth_only>true</depth_only>` tag, that allows to get rid of the rgb image publication and rgb field in the pointcloud. 
This tag is set to false by default in order to be retro compatible.

Tested on Ubuntu 20.04, ROS2 Foxy, Gazebo 11.


anaelle-sw:
Few changes are now added:
* DEPTH_ONLY is no more a camera type, but is replaced by an `impl_` attribute.
* `<depth_only>true</depth_only>` tag allows to not publish `image_raw`. It is set to `false` by default, for retro compatibility.
* `<rgb_pointcloud>false</rgb_pointcloud>` tag allows to get rid of the `rgb` field in the pointcloud output. It is set to `true` by default, for retro compatibility.

Recap table (first line is by default values):
| `depth_only`  | `rgb_pointcloud` | `image_raw` publication | output pointcloud type |
| ------------- | ------------- | -------------	| ------------- |
| false | true | yes | xyz rgb |
| true | true| no | xyz rgb |
| true | false | no | xyz |
| false | false | yes | xyz |

Tested on Ubuntu 20.04, ROS2 Foxy, Gazebo 11.

PR #1121, opened 2020-06-12
Replace deprecated parameters callback API

jacobperron:
rclcpp::Node now supports multiple parameter callbacks, so we do not need to worry about overwriting an existing callback.
This change fixes compile time deprecation warnings since ROS Foxy.


jacobperron:
Linux: [Build Status](https://ci.ros2.org/job/ci_linux/11132/)

chapulina (approved):

Issue #1101, opened 2020-05-08
[ROS2] Should gazebo_ros_p3d plugin broadcast tf2 frame "base_link"?


mvoellmy:
I tried to make the gazebo_ros_p3d plugin broadcast the tf frame in addition with the odom-topic, but noticed that this was not implemented. [`gazebo_ros_p3d.cpp`](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/eloquent/gazebo_plugins/src/gazebo_ros_p3d.cpp) includes 
```
#include <tf2_ros/transform_broadcaster.h>
#include <tf2_ros/transform_listener.h>
```
but I could not find, where/if they are used, so I assume that this functionality is planned to be implemented.

I can implement the functionality and create a pull request, but wanted to first check in if this is the right way to get the tf broadcast or if there is a different method intended for this functionality.

**Environment:**

 - OS: Ubuntu 18.04
 - Gazebo version: Gazebo 9
 - ROS version: Eloquent
  - gazebo_ros_pkgs: debian install
 - glx server in use: server glx vendor string: SGI

han88:
Hi is there a way to fix the same issue in [ROS1] noetic?

PR #1124, opened 2020-06-16
[ROS2] Option for P3D plugin to broadcast tf frame

mvoellmy:
This PR refers to the questions described in #1101 

It adds the ability of broadcasting the tf transform from the specified body to the specified frame in addition to the odom topic.

We use this to visualize the pose of our rover in rviz. Furthermore the name of the tf frame broadcasted can be customized. If no tf_name is specified, then no tf frame will be broadcast.

If there is a better way of doing the described functionality, let me know. I implemented the features mainly for our specific usecase but if they can be used here, I'm happy to also add testing functionality.

**Environment:**

 - OS: Ubuntu 18.04
 - Gazebo version: Gazebo 9
 - ROS version: Eloquent
  - gazebo_ros_pkgs: debian install
 - glx server in use: server glx vendor string: SGI

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_p3d.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_p3d.hpp
@@ -44,6 +44,9 @@ class GazeboRosP3DPrivate;
       <!-- Name of the link within this model whose pose will be published -->
       <body_name>box_link</body_name>
 
+      <!-- If set, broadcasts tf frame of body frame under name specified here. -->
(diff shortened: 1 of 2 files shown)


shonigmann:
I'd also be interested in this functionality, and it seems like it would make for an intuitive addition to the p3d plugin. I was able to merge and build @mvoellmy's code in foxy after fixing one or two minor conflicts. Was there anything else blocking this PR?

han88:
Hi is there a way to fix the same issue in gazebo_ros_p3d.cpp in [ROS1] noetic gazebo_ros_pkg?

PR #931, opened 2019-05-30
Declare parameters and use overrides

chapulina:
From Dashing, parameters need to be [explicitly declared](https://index.ros.org/doc/ros2/Releases/Release-Dashing-Diademata/#declaring-parameters).

This PR changes the usage of the `<parameter>` SDF tag to match the behaviour of parameters passed by the command line through a yaml file. This means they're overriding the default value, but will have no effect unless the node has declared the parameters.

See relevant discussion on https://github.com/ros2/rclcpp/issues/730.

This PR should substitute #930.

FYI @wjwwood

PR #1125, opened 2020-06-17
[ROS 2] Set to true automatically_declare_parameters_from_overrides when creating the node

ahcorde:
Set to true automatically_declare_parameters_from_overrides in gazebo_ros

### DIFF
--- a/gazebo_ros/src/node.cpp
+++ b/gazebo_ros/src/node.cpp
@@ -96,6 +96,7 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
 
   rclcpp::NodeOptions node_options;
   node_options.arguments(arguments);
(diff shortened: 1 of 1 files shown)


ahcorde:
When parameters are declared in the URDF then these parameters should appear as param of the node, right ?.

Right now you need to explicitly declared them in the code with `declare_parameter()`. 

is this the expected behaviour ?

chapulina:
Oh these parameter overrides bring me bad memories... I think we should add some tests with the expectations we want, if we don't have any yet. Here are some references:

* https://github.com/ros2/rclcpp/issues/730
* https://github.com/ros-simulation/gazebo_ros_pkgs/pull/931

ahcorde:
Closing this PR because there was a long discussion about this, thank for the pointers

PR #1126, opened 2020-06-17
gazebo_ros_wheel_slip: support parameters specified in launch file

scpeters:
This re-targets #1111 to noetic.

The `gazebo_ros_wheel_slip` plugin currently ignores slip compliance values specified in a  launch file due to the check of the `level` parameter added in 3cf9f5e1bdf424d5b1866e74f3bde2e904c74298, which ensures that parameter values specified in the urdf / sdf xml parameters are not over-ridden by the default values specified in the `.cfg` file. In order to respect the values set in launch files but ignore the defaults in the `.cfg` file, change the `.cfg` default values to `-1` and adjust the callback to ignore negative values. This is fine because gazebo's WheelSlipPlugin already has default values for the slip compliance parameters, so the `.cfg` defaults are extraneous.

This also reverts 3cf9f5e1bdf424d5b1866e74f3bde2e904c74298 so that the `level` parameter is ignored again.

See also the comment by @maxx84 on #1111: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1111#issuecomment-642156729

### DIFF
--- a/gazebo_plugins/cfg/WheelSlip.cfg
+++ b/gazebo_plugins/cfg/WheelSlip.cfg
@@ -9,7 +9,7 @@ from dynamic_reconfigure.parameter_generator_catkin import *
 gen = ParameterGenerator()
 
 #       Name                                      Type      Reconf level   Description                                                                  Default    Min   Max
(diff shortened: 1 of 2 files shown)


jacobperron (approved):
LGTM

scpeters:
one unrelated test failure; it's probably flaky

PR #1128, opened 2020-06-19
Dashing ➡️ Eloquent

chapulina:
There aren't any important code changes, mainly just syncing changelogs and other commits that have already been ported across branches.

PR #1129, opened 2020-06-19
Eloquent ➡️ Foxy

chapulina:
Forward-port of `eloquent` to `ros2`, including #1128.


chapulina:
A lot of tests that currently fail on `ros2` will be fixed by this - that branch has 17 test failures, this one has only 6. They're different failures though, mainly because previously those tests couldn't even start.

PR #1133, opened 2020-07-01
[noetic][planar_move_controller] add cmdTimeout param

doisyg:
Adding a param to stop a robot using the **planar_move_controller** when no Twist is published for more than the specified `cmdTimeout`. Not used if negative or not specified

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_planar_move.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_planar_move.h
@@ -77,7 +77,7 @@ namespace gazebo {
       std::string robot_base_frame_;
       double odometry_rate_;
       double cmd_timeout_;
-      ros::Time last_cmd_received_time;
+      ros::Time last_cmd_received_time_;
 
       // Custom Callback Queue
       ros::CallbackQueue queue_;
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_planar_move.h
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_planar_move.h
@@ -76,6 +76,8 @@ namespace gazebo {
       std::string odometry_frame_;
       std::string robot_base_frame_;
       double odometry_rate_;
+      double cmd_timeout_;
+      ros::Time last_cmd_received_time;
 
       // Custom Callback Queue
       ros::CallbackQueue queue_;
--- a/gazebo_plugins/src/gazebo_ros_planar_move.cpp
+++ b/gazebo_plugins/src/gazebo_ros_planar_move.cpp
@@ -114,7 +114,7 @@ namespace gazebo
     cmd_timeout_ = -1;
     if (!sdf->HasElement("cmdTimeout"))
     {
-      ROS_WARN_NAMED("planar_move", "PlanarMovePlugin (ns = %s) missing <odometryRate>, "
+      ROS_WARN_NAMED("planar_move", "PlanarMovePlugin (ns = %s) missing <cmdTimeout>, "
           "defaults to %f",
           robot_namespace_.c_str(), cmd_timeout_);
     }
--- a/gazebo_plugins/src/gazebo_ros_planar_move.cpp
+++ b/gazebo_plugins/src/gazebo_ros_planar_move.cpp
@@ -111,6 +111,17 @@ namespace gazebo
     {
       odometry_rate_ = sdf->GetElement("odometryRate")->Get<double>();
     }
+    cmd_timeout_ = -1;
+    if (!sdf->HasElement("cmdTimeout"))
+    {
+      ROS_WARN_NAMED("planar_move", "PlanarMovePlugin (ns = %s) missing <odometryRate>, "
+          "defaults to %f",
+          robot_namespace_.c_str(), cmd_timeout_);
+    }
+    else
+    {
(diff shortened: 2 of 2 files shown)


gazebo_plugins/src/gazebo_ros_planar_move.cpp
```diff
@@ -111,6 +111,17 @@ namespace gazebo
     {
       odometry_rate_ = sdf->GetElement("odometryRate")->Get<double>();
     }
+    cmd_timeout_ = -1;
+    if (!sdf->HasElement("cmdTimeout"))
+    {
+      ROS_WARN_NAMED("planar_move", "PlanarMovePlugin (ns = %s) missing <odometryRate>, "
```
ahcorde: This warning message is wrong

gazebo_plugins/src/gazebo_ros_planar_move.cpp
```diff
@@ -168,6 +179,11 @@ namespace gazebo
   void GazeboRosPlanarMove::UpdateChild()
   {
     boost::mutex::scoped_lock scoped_lock(lock);
+    if((ros::Time::now()-last_cmd_received_time).toSec() > cmd_timeout_ && cmd_timeout_>=0){
```
ahcorde:
```suggestion
    if ((ros::Time::now()-last_cmd_received_time).toSec() > cmd_timeout_ && cmd_timeout_>=0)
    {
```

gazebo_plugins/src/gazebo_ros_planar_move.cpp
```diff
@@ -207,6 +223,7 @@ namespace gazebo
       const geometry_msgs::Twist::ConstPtr& cmd_msg)
   {
     boost::mutex::scoped_lock scoped_lock(lock);
+    last_cmd_received_time=ros::Time::now();
```
ahcorde:
```suggestion
    last_cmd_received_time = ros::Time::now();
```

ahcorde (changes requested):

ahcorde (approved):
LGTM

doisyg:
Is anything blocking this merge ?

gazebo_plugins/include/gazebo_plugins/gazebo_ros_planar_move.h
```diff
@@ -76,6 +76,8 @@ namespace gazebo {
       std::string odometry_frame_;
       std::string robot_base_frame_;
       double odometry_rate_;
+      double cmd_timeout_;
+      ros::Time last_cmd_received_time;
```
chapulina: I'll just point out that this breaks ABI, but I don't think it's very common to extend this plugin, so this may not affect many users. We're also not being very strict about ABI with ROS 1.

chapulina: I also couldn't fail to notice that the variables are inconsistent, the time doesn't end with `_`. But I see another variable in the same situation above, so this doesn't seem to be enforced well...

gazebo_plugins/src/gazebo_ros_planar_move.cpp
```diff
@@ -168,6 +179,12 @@ namespace gazebo
   void GazeboRosPlanarMove::UpdateChild()
   {
     boost::mutex::scoped_lock scoped_lock(lock);
+    if ((ros::Time::now()-last_cmd_received_time).toSec() > cmd_timeout_ && cmd_timeout_>=0)
```
chapulina: A small suggestion would be to check `cmd_timeout_>=0` first, so that `ros::Time::now` doesn't need to be called every iteration for all simulations where no timeout is given.

gazebo_plugins/src/gazebo_ros_planar_move.cpp
```diff
@@ -111,6 +111,17 @@ namespace gazebo
     {
       odometry_rate_ = sdf->GetElement("odometryRate")->Get<double>();
     }
+    cmd_timeout_ = -1;
+    if (!sdf->HasElement("cmdTimeout"))
+    {
+      ROS_WARN_NAMED("planar_move", "PlanarMovePlugin (ns = %s) missing <cmdTimeout>, "
```
chapulina, line 120: It feels a bit weird to warn about not setting an optional parameter. All current users will start seeing this warning as soon as they update, even though there's nothing wrong with their setup. I think this should be and `INFO`... But I see it is following the pattern from the rest of the file, so I think it's ok to leave it like this too...

chapulina (approved):
I just have some minor comments, but this looks good, thanks for the addition.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_planar_move.h
doisyg: renamed `last_cmd_received_time` `last_cmd_received_time_`

gazebo_plugins/src/gazebo_ros_planar_move.cpp
doisyg: Nice save indeed

gazebo_plugins/src/gazebo_ros_planar_move.cpp
doisyg: As you prefer

doisyg:
Conforming to comments and merged last commit from noetic

PR #1135, opened 2020-07-08
colcon.pkg: build gazebo first in colcon workspace

scpeters:
Add a colcon.pkg file to gazebo_dev with gazebo's cmake project
name "Gazebo" listed as a build dependency to support building
gazebo from source in a colcon workspace.

### DIFF
--- a/gazebo_dev/colcon.pkg
+++ b/gazebo_dev/colcon.pkg
@@ -1,0 +1,8 @@
+# Configuration file for colcon (https://colcon.readthedocs.io).
+#
+# Please see the doc for the details of the spec:
+#   - https://colcon.readthedocs.io/en/released/user/configuration.html#colcon-pkg-files
+
+{
+  "build-dependencies": ["Gazebo"],
+}
--- a/gazebo_dev/colcon.pkg
+++ b/gazebo_dev/colcon.pkg
@@ -4,5 +4,5 @@
 #   - https://colcon.readthedocs.io/en/released/user/configuration.html#colcon-pkg-files
 
 {
-  "build-dependencies": ["Gazebo"],
+  "dependencies": ["Gazebo"],
 }
--- a/gazebo_plugins/colcon.pkg
+++ b/gazebo_plugins/colcon.pkg
@@ -1,0 +1,8 @@
+# Configuration file for colcon (https://colcon.readthedocs.io).
+#
+# Please see the doc for the details of the spec:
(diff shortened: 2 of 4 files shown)


chapulina (approved):
Oh, this is nice, I didn't know about this file! And it looks like [it adds to](https://github.com/colcon/colcon.readthedocs.org/issues/44) `package.xml` in case that's also present, so it shouldn't break anything.

I think an alternative solition could be adding `Gazebo` as a rosdep key that points to the latest Gazebo debs. This way the dependency could be directly added to `package.xml`.

scpeters:
> I think an alternative solition could be adding `Gazebo` as a rosdep key that points to the latest Gazebo debs. This way the dependency could be directly added to `package.xml`.

`gazebo_dev` already has dependencies on `libgazebo9-dev` and `gazebo9`, so I think adding `Gazebo` might be redundant. This mainly just helps with building from source in a colcon workspace, so I figured a colcon.pkg file was a concise solution

chapulina:
> I think adding Gazebo might be redundant.

Oh yeah I think that it would probably substitute `gazebo9`. But it was just a thought. Something to keep in mind for the future if we don't want to maintain 2 files in other places, like Ignition.

scpeters:
I just ran into one issue with this approach, which is that unless you use the `--merge-install` parameter to `colcon build`, any dependents of `gazebo_dev` (like `gazebo_ros` and `gazebo_plugins`) are unable to `find_package(gazebo)`, I believe because `colcon` is not including `Gazebo` cmake paths for those packages. Using `--merge-install` is easy enough though.

chapulina:
Do you think it's worth it adding a note about `--merge-install` as a comment in the `colcon.pkg` file itself?

scpeters:
At the suggestion of @dirk-thomas, I switched to a regular `dependencies` declaration for `Gazebo` instead of `build-dependencies` to fix compilation without `--merge-install`. There still seems to be a runtime issue with the `PKG_CONFIG_PATH` though

edit: the issue is caused by our use of `pkg-config` to resolve the location of gazebo's setup.sh script when launching gazebo:

* https://github.com/ros-simulation/gazebo_ros_pkgs/blob/melodic-devel/gazebo_ros/scripts/gzserver#L26

scpeters:
At the suggestion of @dirk-thomas, I duplicated the `colcon.pkg` files for each package that depends on `Gazebo` in eadfadb (gazebo_ros, gazebo_plugins, gazebo_ros_control). This makes it so `--merge-install` isn't required.

chapulina (approved):

PR #1137, opened 2020-07-22
gazebo_ros_api_plugin cleanup

scpeters:
I noticed with @clalancette that `gazebo_ros_api_plugin` advertises the `/clock` topic and sets some variables in multiple places ([loadGazeboRosApiPlugin()](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/f7a19b82f449044c5d39fc1f02f2fbb4471a7457/gazebo_ros/src/gazebo_ros_api_plugin.cpp#L212) and [advertiseServices()](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/f7a19b82f449044c5d39fc1f02f2fbb4471a7457/gazebo_ros/src/gazebo_ros_api_plugin.cpp#L253)). This removes that duplicate code from `advertiseServices()` and makes sure is called before `advertiseServices()`.

There's also an unused private function that is an overload of `publishSimTime()`, so I've removed it here and added doxygen for the remaining `publishSimTime()`.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -205,6 +205,11 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
 
   // Manage clock for simulated ros time
   pub_clock_ = nh_->advertise<rosgraph_msgs::Clock>("/clock",10);
+
+  /// \brief advertise all services
+  if (enable_ros_network_)
+    advertiseServices();
+
   // set param for use_sim_time if not set by user already
   if(!(nh_->hasParam("/use_sim_time")))
     nh_->setParam("/use_sim_time", true);
@@ -215,10 +220,6 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
 #else
   last_pub_clock_time_ = world_->GetSimTime();
 #endif
-
-  /// \brief advertise all services
-  if (enable_ros_network_)
-    advertiseServices();
 
   // hooks for applying forces, publishing simtime on /clock
   wrench_update_event_ = gazebo::event::Events::ConnectWorldUpdateBegin(boost::bind(&GazeboRosApiPlugin::wrenchBodySchedulerSlot,this));
--- a/gazebo_ros/src/gazebo_ros_api_plugin.cpp
+++ b/gazebo_ros/src/gazebo_ros_api_plugin.cpp
@@ -204,7 +204,7 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
   pub_model_states_connection_count_ = 0;
 
   // Manage clock for simulated ros time
-  pub_clock_ = nh_->advertise<rosgraph_msgs::Clock>("/clock",10);
+  pub_clock_ = nh_->advertise<rosgraph_msgs::Clock>("/clock", 10);
 
   /// \brief advertise all services
   if (enable_ros_network_)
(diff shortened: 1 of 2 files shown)


gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -193,7 +193,6 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
 
   gazebonode_ = gazebo::transport::NodePtr(new gazebo::transport::Node());
   gazebonode_->Init(world_name);
-  //stat_sub_ = gazebonode_->Subscribe("~/world_stats", &GazeboRosApiPlugin::publishSimTime, this); // TODO: does not work in server plugin?
```
clalancette, line 196: Since this is commented out, we also don't need the `void GazeboRosApiPlugin::publishSimTime(const boost::shared_ptr<gazebo::msgs::WorldStatistics const> &msg)` signature below, so it makes sense to remove both.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -204,10 +203,6 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
   pub_link_states_connection_count_ = 0;
   pub_model_states_connection_count_ = 0;
 
-  /// \brief advertise all services
-  if (enable_ros_network_)
-    advertiseServices();
-
```
clalancette: By moving this from here to after we advertise the clock, we get mostly the same behavior as was duplicated in the `advertiseServices` method.  One thing I mentioned to @scpeters in person is that this slightly changes the order of operations, since before we would advertise the clock and then get the parameters later, whereas now we are doing it all at the same time.  I don't have a strong reason why that would matter, but I thought I would point it out.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -249,9 +248,6 @@ void GazeboRosApiPlugin::advertiseServices()
     return;
   }
 
-  // publish clock for simulated ros time
-  pub_clock_ = nh_->advertise<rosgraph_msgs::Clock>("/clock",10);
-
```
clalancette, line 254: And now this code is duplicate because we already do this in the main `loadGazeboRosApiPlugin`

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -204,12 +203,13 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
   pub_link_states_connection_count_ = 0;
   pub_model_states_connection_count_ = 0;
 
+  // Manage clock for simulated ros time
+  pub_clock_ = nh_->advertise<rosgraph_msgs::Clock>("/clock",10);
+
```
clalancette: By moving this above `advertiseServices`, we get exactly the same behavior as before, but without the code duplication.

clalancette (approved):
All of my comments are exactly that; comments pointing out why things are like they are.  The changes look good to me.  I'll approve, but someone else from the Gazebo team should probably take a look.

gazebo_ros/src/gazebo_ros_api_plugin.cpp
```diff
@@ -204,12 +203,13 @@ void GazeboRosApiPlugin::loadGazeboRosApiPlugin(std::string world_name)
   pub_link_states_connection_count_ = 0;
   pub_model_states_connection_count_ = 0;
 
+  // Manage clock for simulated ros time
+  pub_clock_ = nh_->advertise<rosgraph_msgs::Clock>("/clock",10);
```
ahcorde:
```suggestion
  pub_clock_ = nh_->advertise<rosgraph_msgs::Clock>("/clock", 10);
```

ahcorde (changes requested):
small nit

scpeters:

scpeters:
@ahcorde I merged your suggestion

ahcorde (approved):

PR #1144, opened 2020-08-11
[Noetic] Added directives if profiler is not installed

ahcorde:
Added directives if profiler is not installed. This will avoid compilation issues


chapulina (approved):
Ok this is a good idea, because people may try to install this without having `libignition-common3-profiler-dev` installed, which will be a new dependency from Gazebo 11.1.0

PR #1145, opened 2020-08-11
[Noetic] Bridge to republish PerformanceMetrics in ROS

ahcorde:
This PR is related to this other PR https://github.com/osrf/gazebo/pull/2819 which adds a performanceMetrics message. Then this msg will be republish from the Gazebo transport layer to the ROS network.

### DIFF
--- a/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
+++ b/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
@@ -292,9 +292,10 @@ private:
   /// \brief Unused
   void onResponse(ConstResponsePtr &response);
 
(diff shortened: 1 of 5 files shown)


ahcorde:
The CI is not going to build until Gazebo 11 it's released with this new feature.

gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
```diff
@@ -291,6 +292,9 @@ class GazeboRosApiPlugin : public SystemPlugin
   /// \brief Unused
   void onResponse(ConstResponsePtr &response);
 
+  /// \brief Subscriber callback for performance metrics. This will be send in the ROS network
+  void onPerformanceMetrics(const boost::shared_ptr<gazebo::msgs::PerformanceMetrics const> &msg);
```
chapulina, line 298: Could you add some ifdefs so this works with the current minor version of Gazebo 11? We should proceed with the assumption that the Gazebo 11 packages on https://packages.ros.org won't be upgraded, and this needs to work with that.

iche033 (approved):
works for me!

PR #1148, opened 2020-08-13
Port wheel slip plugin to ros2 (forward port from eloquent)

scpeters:
This is a forward port of #1099.

Initial port of gazebo_ros_wheel_slip_plugin to ros2.

It includes a similar change to #1111, which moved the
WheelSlipPlugin::Load call before the parameter callback
so that the values from SDF/URDF are set first.
Then the callback is changed to ignore negative values, and the
default slip parameter values are set to -1, so that the SDF/URDF
values are still preferred unless a different parameter value
is specified in a launch file.

### DIFF
--- a/gazebo_plugins/CMakeLists.txt
+++ b/gazebo_plugins/CMakeLists.txt
@@ -448,6 +448,23 @@ if(ENABLE_PROFILER)
   target_link_libraries(gazebo_ros_projector ${ignition-common3_LIBRARIES})
 endif()
 
+# gazebo_ros_wheel_slip
+if(NOT GAZEBO_VERSION VERSION_LESS 9.5)
+  add_library(gazebo_ros_wheel_slip SHARED
+    src/gazebo_ros_wheel_slip.cpp
+  )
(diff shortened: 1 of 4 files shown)


gazebo_plugins/CMakeLists.txt
```diff
@@ -448,6 +448,23 @@ if(ENABLE_PROFILER)
...
+  ament_target_dependencies(gazebo_ros_wheel_slip
+    "gazebo_dev"
+    "gazebo_ros"
+    "rclcpp"
+    "std_msgs"
+  )
+  target_link_libraries(gazebo_ros_wheel_slip
+    WheelSlipPlugin
+  )
+  ament_export_libraries(gazebo_ros_wheel_slip)
```
jacobperron, line 466: Looks like there is an additional change needed for the include directories
```suggestion
  target_include_directories(gazebo_ros_wheel_slip PUBLIC include)
  ament_export_libraries(gazebo_ros_wheel_slip)
```

jacobperron (approved):
LGTM (pending CI)

PR #1166, opened 2020-09-22
Add repos file with dependencies

jacobperron:
We can then have the installation instructions reference this file.

If this looks reasonable, I can backport to Dashing, Eloquent, and Foxy.


chapulina (approved):
Good idea :+1:

jacobperron:
Ports to Eloquent (#1167) and Dashing (#1168).

PR #1173, opened 2020-09-30
gazebo_ros_utils: don't set tf_prefix if empty

scpeters:
Similar changes were made in #1143 to address #554, but `gazebo_ros_utils` wasn't included. I'm not sure what the consequences of this change are.


jacobperron (approved):
On the face of it, this looks correct to me. It keeps the logic consistent for when no tf_prefix is provided.

It looks like `tf_prefix_` is only used in `GazeboRos::resolveTF`, which is used in the diff drive and tricycle drive plugins.

MatthijsBurgh:

Issue #1175, opened 2020-10-08
Both depth camera plugins stop publishing after upgrading gazebo9


corot:
Both depth camera plugins stop publishing after upgrading gazebo9 from amd64 9.14.0-1 to 9.15.0-2
Looks like the plugin interface gets broken...  no idea how, but the fact is that the plugin gets initialized and advertises the topics.
But the callback methods with the images (OnNewDepthFrame, OnNewRGBPointCloud, OnNewImageFrame) never get called by gazebo!

mitschron:
I have the same problem with all gazebo plugins. I am working with gazebo11 ros-melodic. 
In my case it seems to be caused by the performance metric subscriber added on this commit: 6c53c21746c1246ec8eac2a0159719ba63664e61
Changing the topic to from absolute namespace to relative seems to fix it.  ("/gazebo/performance_metrics" -> "~/performance_metrics").  
That is in line **202** of _gazebo_ros_api_plugin.cpp_

HighPriest:
@mitschron Can you turn your fix into a pull request?
Alot of people have this problem, including my team. We have found a different fix, but it's on a much higher abstraction level.

mitschron:
> @mitschron Can you turn your fix into a pull request?
> Alot of people have this problem, including my team. We have found a different fix, but it's on a much higher abstraction level.

https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1180

ahcorde:
Hello, 

can you provide more details? are you installing the deb packages? compiling from source ?

HighPriest:
@ahcorde 
The issue appears in both source and package installation of gazebo11 against ROS2 Foxy.
Building the gazebo_ros_pkgs/dashing branch against Gazebo11 "fixes" the issue with gazebo_ros_camera and gazebo_ros_depth_camera plugins.

ahcorde:
can you run it with `--verbose` ?

HighPriest:
You mean run the build or run Gazebo, or run Gazebo and spawn an object with camera plugin?
There is no meaningful output out of Gazebo either way.
As I said before, debugging the plugging shows the "publish" function executes properly, but the data does not appear on the ROS topic.

The fix made by @mitschron appears to silence the PROFILER, which makes me think that it might be something related with it.

ahcorde:
@HighPriest
I mean run Gazebo and spawn an object with camera plugin. 
```
[...]

PR #1180, opened 2020-10-16
gazebo_ros: Use relative namespace for performance metrics

mitschron:


chapulina (changes requested):
I don't think this will work. Gazebo is publishing at an exact topic without a namespace:

https://github.com/osrf/gazebo/blob/10863ffa8ec28e8e0e3d3a3a588d21cdc78055ab/gazebo/sensors/SensorManager.cc#L186

Also, the `~` will resolve to `/gazebo/worldname`, not just `/gazebo`.

Are you trying to solve some issue?

mitschron:
> I don't think this will work. Gazebo is publishing at an exact topic without a namespace:
> 
> https://github.com/osrf/gazebo/blob/10863ffa8ec28e8e0e3d3a3a588d21cdc78055ab/gazebo/sensors/SensorManager.cc#L186
> 
> Also, the `~` will resolve to `/gazebo/worldname`, not just `/gazebo`.
> 
> Are you trying to solve some issue?

Please have a look at issue #1175. Adding the performance metrics seems to be breaking the plugins.

scpeters:
> > I don't think this will work. Gazebo is publishing at an exact topic without a namespace:
> > https://github.com/osrf/gazebo/blob/10863ffa8ec28e8e0e3d3a3a588d21cdc78055ab/gazebo/sensors/SensorManager.cc#L186
> > Also, the `~` will resolve to `/gazebo/worldname`, not just `/gazebo`.
> > Are you trying to solve some issue?
> 
> Please have a look at issue #1175. Adding the performance metrics seems to be breaking the plugins.

I believe the underlying issue in #1175 is https://github.com/osrf/gazebo/issues/2902, and it can occur whenever there is a subscriber to the gazebo topic `/gazebo/performance_metrics` and gazebo's `SensorManager` has sensors in 2 or more `SensorContainer`s (ie a camera and a lidar, or a lidar and an IMU). This pull request works around the issue by breaking the gazebo subscriber so that it subscribes to an incorrect topic. I've submitted a different workaround in #1202 to `noetic-devel` (and #1203 for melodic and #1205 for `ros2`) that only subscribes to the gazebo topic if there is anyone subscribing to the corresponding ROS topic. So it will still be possible to encounter the deadlock, but it requires an extra step of subscribing to the `/performance_metrics` topic. I've opened #1209 to backport this workaround to `kinetic-devel`, so I think this can be closed once that is merged.

scpeters:
I'm going to close this since I've just merged #1209 to `kinetic-devel`, and I believe it's a suitable workaround to #1175

PR #1185, opened 2020-10-29
Use namespace of the spawn entity as the default robot namespace

jacobperron:
This is analogous to ROS 1, where launching the spawn_model script would start plugins in the same namespace as the spawn_model script in the launch file.

This should simplify some instances where the spawn_entity script is used in launch files,

For example, in situations where we might be including a launch script in a namespace we currently have to also pass the namespace for plugins independently:

```xml
<launch>
  <push-ros-namespace namespace="/my_namespace" />
  <node name="spawn_entity" pkg="gazebo_ros" exec="spawn_entity.py" output="screen"
        args="-entity foo -file foo.urdf -robot_namespace /my_namespace" />
  ...
</launch>
```

This could be reduced to:

```xml
<launch>
  <push-ros-namespace namespace="/my_namespace" />
  <node name="spawn_entity" pkg="gazebo_ros" exec="spawn_entity.py" output="screen"
        args="-entity foo -file foo.urdf" />
  ...
</launch>
```

### DIFF
--- a/gazebo_ros/scripts/spawn_entity.py
+++ b/gazebo_ros/scripts/spawn_entity.py
@@ -67,10 +67,11 @@ class SpawnEntityNode(Node):
         parser.add_argument('-reference_frame', type=str, default='',
                             help='Name of the model/body where initial pose is defined.\
                             If left empty or specified as "world", gazebo world frame is used')
-        parser.add_argument('-gazebo_namespace', type=str, default=self.get_namespace(),
+        parser.add_argument('-gazebo_namespace', type=str, default='',
                             help='ROS namespace of gazebo offered ROS interfaces. \
                             Default is without any namespace')
-        parser.add_argument('-robot_namespace', type=str, default='',
+        entity_namespace = self.get_namespace() if self.get_namespace() != '/' else ''
+        parser.add_argument('-robot_namespace', type=str, default=entity_namespace,
                             help='change ROS namespace of gazebo-plugins')
         parser.add_argument('-timeout', type=float, default=30.0,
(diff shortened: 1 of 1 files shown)


ahcorde:
I think there is/was a reason to do it. This PR I made just do the opposite https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1117

jacobperron:
> I think there is/was a reason to do it. This PR I made just do the opposite #1117

I see the problem. We could fix this by checking if the node namespace is "/" (I've updated the PR, PTAL).

I think there's a couple questions we need to answer:

1. Should we support using the spawn_entity namespace (if it has one)? I.e. should we allow the use-case motivating this PR:

```
<launch>
  <push-ros-namespace namespace="/my_namespace" />
  <node name="spawn_entity" pkg="gazebo_ros" exec="spawn_entity.py" output="screen"
        args="-entity foo -file foo.urdf" />
</launch>
```

I'm pretty sure this was possible in ROS 1.

2. If we say "yes" to 1, then what is the order of precedence for choosing the spawn_entity node namespace versus the robot_namespace argument?

E.g. in the following example, should the namespace of the model's ROS entities be `/foo` or `/bar`? 

```
<launch>
  <push-ros-namespace namespace="/foo" />
  <node name="spawn_entity" pkg="gazebo_ros" exec="spawn_entity.py" output="screen"
        args="-entity foo -file foo.urdf -robot_namespace /bar" />
</launch>
```

3. Furthermore, what should be the precedence if a plugin in `foo.urdf` contains a `<namespace>` element?

@ahcorde @chapulina Thoughts?

PR #1188, opened 2020-11-07
Fix sdf ros topic naming bug

samsv:
When using .sdfs gazebo might automatically generate invalid ros topic names (names which contain "::"). This checks if the topic name has any ":" and replaces it with a "_" to generate valid names.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_force.cpp
+++ b/gazebo_plugins/src/gazebo_ros_force.cpp
@@ -105,7 +105,9 @@ void GazeboRosForce::Load(physics::ModelPtr _model, sdf::ElementPtr _sdf)
 
   if (this->topic_name_.find(':') != std::string::npos)
   {
+    std::string _topic_name = this->topic_name_;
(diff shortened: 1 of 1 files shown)


chapulina:
Thank you for the contribution, that looks reasonable to me. Just some clarifications:

* What's the failure mode if we leave the topic unchanged? A warning? A crash?
* I think we should print a warning message to the user that this conversion is being made.
* Do you know of other invalid characters that may need to be substituted?

samsv:
It throws the following error:
```
[Err] [Model.cc:1129] Exception occured in the Load function of plugin with name[ros_force] and filename[libros_force.so]. This plugin will not run.
```
The node with the plugin doesn't spawn due to the error, but you can only know which node wasn't created by checking the output of `rostopic list`.
I don't think that there is another invalid character. The ":" is from sdf's naming conventions and something out of the user control, as far as I know.
I agree about printing a warning.

samsv:
I have added the warning message.

chapulina (approved):

PR #1196, opened 2020-11-19
Make executor thread-safe when adding and removing nodes

jacobperron:
gazebo_ros incorrectly assumes rclcpp::MultiThreadedExecutor is thread-safe.
This change adds thread-safe wrappers to methods for adding/removing nodes and spinning.

I believe this fixes https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1169.

This may have some performance impact on two fronts: 1) the locking while spinning and 2) switching from calling spin() to spin_once() (I think) removes the multi-threaded nature of the executor.

I'll see if we can improve performance before taking this out of draft state.

### DIFF
--- a/gazebo_ros/include/gazebo_ros/executor.hpp
+++ b/gazebo_ros/include/gazebo_ros/executor.hpp
@@ -18,6 +18,10 @@
 #include <rclcpp/rclcpp.hpp>
 #include <gazebo/common/Events.hh>
 
+#include <memory>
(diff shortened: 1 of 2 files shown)


jacobperron:
I've opened an alternative to this PR that retains the multi-threaded part of the executor: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1197

PR #1202, opened 2020-12-12
Only subscribe to /gazebo/performance_metrics when necessary

scpeters:


jacobperron (approved):
LGTM

chapulina (approved):
I haven't tried running it but it looks like a reasonable workaround :+1:

PR #1203, opened 2020-12-15
Backport #1202 to melodic-devel: fix performance metrics

scpeters:
Backport of #1202 to `melodic-devel`:

### DIFF
--- a/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
+++ b/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
@@ -391,8 +391,7 @@ private:
   ros::Publisher     pub_performance_metrics_;
   int                pub_link_states_connection_count_;
   int                pub_model_states_connection_count_;
-  // use static variable to avoid ABI break
-  static int         pub_performance_metrics_connection_count_;
+  int                pub_performance_metrics_connection_count_;
 
   // ROS comm
   boost::shared_ptr<ros::AsyncSpinner> async_ros_spin_;
--- a/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
+++ b/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
@@ -138,16 +138,16 @@ public:
   /// \brief advertise services
   void advertiseServices();
 
-  /// \brief
+  /// \brief Callback for a subscriber connecting to LinkStates ros topic.
   void onLinkStatesConnect();
 
-  /// \brief
+  /// \brief Callback for a subscriber connecting to ModelStates ros topic.
   void onModelStatesConnect();
 
-  /// \brief
+  /// \brief Callback for a subscriber disconnecting from LinkStates ros topic.
   void onLinkStatesDisconnect();
 
-  /// \brief
+  /// \brief Callback for a subscriber disconnecting from ModelStates ros topic.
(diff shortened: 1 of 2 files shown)


gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
```diff
@@ -376,6 +391,7 @@ class GazeboRosApiPlugin : public SystemPlugin
   ros::Publisher     pub_performance_metrics_;
   int                pub_link_states_connection_count_;
   int                pub_model_states_connection_count_;
+  int                pub_performance_metrics_connection_count_;
```
jacobperron, line 394: Adding a new member is not ABI compatible.

jacobperron (commented):
The change LGTM, though I'm not sure if we care about maintaining ABI compatibility in this case. @chapulina ?

gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
scpeters: do we need to revert that change from `noetic-devel` too?

jacobperron: I didn't think about it before. Noetic is an interesting case since it is the last ROS 1 release; maybe an exception can be made.

scpeters: In terms of the impact of an ABI break: because this is a plugin, I think breaking its ABI will only cause a problems for binaries of a plugin that links against `GazeboRosApiPlugin`. I don't know of any plugins that do this.

In terms of working around the ABI break: because this is a `SystemPlugin`, I think there will only ever be one instance open at a time. So I think we could safely make this a `static` variable. I'll check with @iche033 to see if he concurs.

scpeters: I made it a `static` variable in 043e5dd, which I think will work. I can port this fix to `noetic-devel` if merge it here

chapulina: We haven't been strict about ABI compatibility in ROS 1, right @j-rivero ?

Especially with this plugin, I don't think many users may be extending it, probably we shouldn't be installing the header... I wouldn't block on the ABI here...

> I think there will only ever be one instance open at a time.

Yeah I think that's the usual use case for this plugin, even though I think technically it can be added more than once, but that could cause other issues with duplicate topics and services...

scpeters: so it sounds like we could merge this either way I think

jacobperron: If we're okay with ABI breaks, it might be nice to keep the change the same as the one on `noetic-devel` for consistency.

I don't mind either way.

jacobperron (approved):

gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h
scpeters: ok, I'll revert 043e5dd to maintain consistency with noetic

scpeters:
I think the CI test failures are unrelated. I'm going to merge

PR #1205, opened 2020-12-22
ros2: Only subscribe to /gazebo/performance_metrics when necessary

scpeters:
This is similar to #1202 and #1203 targeting `ros2` with polling instead of advertiser callbacks.

We are currently subscribing to the `/gazebo/performance_metrics` topic even if there are no subscribers to the ROS topic forwarding this data. This changes `gazebo_ros_init` to only subscribe to the gazebo topic if there are any subscribers to the corresponding ROS topic. While advertiser callbacks are used in ROS 1 but are not yet in ROS2, here we use polling in the `GazeboRosInitPrivate::PublishSimTime` callback to check for subscribers since it is called for each Gazebo
time step.

I've also added the `GAZEBO_ROS_HAS_PERFORMANCE_METRICS` macro to de-duplicate the gazebo version checking logic.


gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -33,6 +33,13 @@
 #include <memory>
 #include <string>
 
+#ifndef GAZEBO_ROS_HAS_PERFORMANCE_METRICS
+#if (GAZEBO_MAJOR_VERSION == 11 && GAZEBO_MINOR_VERSION > 1) || \
+    (GAZEBO_MAJOR_VERSION == 9 && GAZEBO_MINOR_VERSION > 14)
```
scpeters: uncrustify doesn't like my indentation here

~~~
--- src/gazebo_ros_init.cpp
+++ src/gazebo_ros_init.cpp.uncrustify
@@ -38 +38 @@
-    (GAZEBO_MAJOR_VERSION == 9 && GAZEBO_MINOR_VERSION > 14)
+  (GAZEBO_MAJOR_VERSION == 9 && GAZEBO_MINOR_VERSION > 14)
~~~

I think it looks better like this, but it's not a big deal. Is it allowed to tell uncrustify to ignore this line, or should I just make the change?

ahcorde (approved):
I vote for the uncrustify version.

gazebo_ros/src/gazebo_ros_init.cpp
jacobperron: For indentation, I think you can use

```c
// *INDENT-OFF*
#if (GAZEBO_MAJOR_VERSION == 11 && GAZEBO_MINOR_VERSION > 1) || \
    (GAZEBO_MAJOR_VERSION == 9 && GAZEBO_MINOR_VERSION > 14)
// *INDENT-ON*
```

But subjectively, I would just do what uncrustify says to avoid the extra "INDENT" directives. I don't feel strongly one way or the other.

jacobperron (approved):
LGTM with green CI

gazebo_ros/src/gazebo_ros_init.cpp
scpeters: I ended up going with 3a875ea

jacobperron (approved):

scpeters:
just one test failure that looks like the flaky test in #1169

I'll go ahead and merge

PR #1206, opened 2020-12-22
Update test worlds to use the correct element names for xyzOffset and rpyOffset

jacobperron:
The test worlds add an extra 's' character, but the real names are singular.

The names actually used:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/a63566be22361fa1f02ebcca4a9857d233e1c2ac/gazebo_plugins/src/gazebo_ros_p3d.cpp#L94

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/a63566be22361fa1f02ebcca4a9857d233e1c2ac/gazebo_plugins/src/gazebo_ros_p3d.cpp#L102

---

Side note, **this changed in ROS 2 to be plural**:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/a73abf58afde5b4e9d2d8c8085a20a99e0773923/gazebo_plugins/src/gazebo_ros_p3d.cpp#L132

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/a73abf58afde5b4e9d2d8c8085a20a99e0773923/gazebo_plugins/src/gazebo_ros_p3d.cpp#L138

And the migration guide mixes both plural and singular :confused: https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-P3D

I think it makes sense to continue to use singular names in ROS 2, `xyz_offset` and `rpy_offset`. I'll open a PR proposing that change.

### DIFF
--- a/gazebo_plugins/test/p3d_test/worlds/3_double_pendulums.world
+++ b/gazebo_plugins/test/p3d_test/worlds/3_double_pendulums.world
@@ -150,8 +150,8 @@
           <bodyName>link_1</bodyName>
           <topicName>/model_1/link_1/pose</topicName>
           <frameName>world</frameName>
-          <xyzOffsets>0 0 -2.1</xyzOffsets>
-          <rpyOffsets>0 0  0.0</rpyOffsets>
+          <xyzOffset>0 0 -2.1</xyzOffset>
+          <rpyOffset>0 0  0.0</rpyOffset>
       </plugin>
       <plugin name="p3d_link_2_controller" filename="libgazebo_ros_p3d.so">
           <alwaysOn>true</alwaysOn>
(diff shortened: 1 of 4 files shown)


jacobperron:
I guess it's worth backporting this change to melodic and kinetic as well.

jacobperron:
Rather than change the ROS 2 implementation, I've just corrected the [migration guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-P3D). Though I have a subjective preference towards the singular form, changing it at this point seems too disruptive.

scpeters (approved):

scpeters:
> Rather than change the ROS 2 implementation, I've just corrected the [migration guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-P3D). Though I have a subjective preference towards the singular form, changing it at this point seems too disruptive.

the Imu plugins also use the singular form. I think we should update the ros2 plugins to accept both and deprecate the plural form

jacobperron:
> the Imu plugins also use the singular form. I think we should update the ros2 plugins to accept both and deprecate the plural form

Sounds good to me. I can open a PR deprecating the plural form.

Correct me if I'm wrong, but it doesn't look like the IMU plugin in ROS 2 has an equivalent SDF element.

jacobperron:
Here's the patch for ROS 2, deprecating plural names: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1210

PR #1209, opened 2020-12-22
Backport #1202 to kinetic: fix performance metrics

scpeters:
Backport of #1202 to `kinetic-devel`, an alternative to #1180:


ahcorde (approved):
LGTM with green CI

PR #1210, opened 2020-12-23
Make p3d offset element names singular

jacobperron:
- `<xyz_offsets>` is renamed to `<xyz_offset>`
- `<rpy_offsets>` is renamed to `<rpy_offset>`

The old names can still be used, but are deprecated.
This is more consistent with the naming convention used in ROS 1 versions.

I've also added a test to ensure the deprecated elements continue to work.


jacobperron:
I've updated the migration guide to reflect this change (adding a note that it applies since Galactic): https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-P3D

jacobperron:

chapulina (approved):
:+1:

jacobperron:
`test_node` is failing in both CI builds. This is a known failure #1169

PR #1197, opened 2020-11-20
Replace executor with a thread safe implementation

jacobperron:
The rclcpp::MultiThreadedExecutor is not thread safe, and so it is possible to hit races that cause gazebo_ros to crash. Specifically, there's a race when adding and removing nodes, which can happen when node are created and destroyed during runtime.

This is a copy of rclcpp::Executor with some modifications to make it thread safe. A lock was added to the spin implementation and also to the add_node and remove_node methods. In order to avoid deadlock, we must have a timeout in the spin loop while waiting for work.

Fixes #1169

Replaces https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1196

PR #1212, opened 2021-01-13
Fix executor to avoid random exceptions when shutting down

ivanpauno:
This PR together with https://github.com/ros2/rclcpp/pull/1505 seems to fix `test_node` failures.

Fixes https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1169.
Replaces https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1197.

### DIFF
--- a/gazebo_ros/src/executor.cpp
+++ b/gazebo_ros/src/executor.cpp
@@ -27,7 +27,7 @@ Executor::Executor()
   while (!this->spinning) {
     // TODO(ivanpauno): WARN Terrible hack here!!!!
     // We cannot call rclcpp::shutdown asynchronously, because it generates exceptions that
-    // cannot be caught properly
+    // cannot be caught properly (see https://github.com/ros2/rclcpp/issues/1139).
     // Executor::cancel() doesn't cause this problem, but it has a race.
     // Wait until the launched thread starts spinning to avoid the race ...
     std::this_thread::sleep_for(100ms);
--- a/gazebo_ros/src/executor.cpp
+++ b/gazebo_ros/src/executor.cpp
@@ -22,15 +22,22 @@ namespace gazebo_ros
 Executor::Executor()
 : spin_thread_(std::bind(&Executor::run, this))
 {
+  using namespace std::chrono_literals;
   sigint_handle_ = gazebo::event::Events::ConnectSigInt(std::bind(&Executor::shutdown, this));
+  while (!this->spinning) {
+    // TODO(ivanpauno): WARN Terrible hack here!!!!
+    // We cannot call rclcpp::shutdown asynchronously, because it generates exceptions that
+    // cannot be caught properly
+    // Executor::cancel() doesn't cause this problem, but it has a race.
+    // Wait until the launched thread starts spinning to avoid the race ...
+    std::this_thread::sleep_for(100ms);
+  }
 }
 
 Executor::~Executor()
 {
   // If ros was not already shutdown by SIGINT handler, do it now
-  if (rclcpp::ok()) {
-    rclcpp::shutdown();
-  }
+  this->shutdown();
   spin_thread_.join();
 }
 
@@ -41,7 +48,9 @@ void Executor::run()
 
 void Executor::shutdown()
 {
-  rclcpp::shutdown();
+  if (this->spinning) {
+    this->cancel();
+  }
 }
 
 }  // namespace gazebo_ros


ivanpauno:
As a reference of the issue of using `rclcpp::shutdown`, see https://github.com/ros2/rclcpp/issues/1139.

ivanpauno:
@jacobperron can you review this?

gazebo_ros/src/executor.cpp
```diff
@@ -22,15 +22,22 @@ namespace gazebo_ros
...
 : spin_thread_(std::bind(&Executor::run, this))
 {
+  using namespace std::chrono_literals;
   sigint_handle_ = gazebo::event::Events::ConnectSigInt(std::bind(&Executor::shutdown, this));
+  while (!this->spinning) {
+    // TODO(ivanpauno): WARN Terrible hack here!!!!
+    // We cannot call rclcpp::shutdown asynchronously, because it generates exceptions that
+    // cannot be caught properly
+    // Executor::cancel() doesn't cause this problem, but it has a race.
+    // Wait until the launched thread starts spinning to avoid the race ...
```
jacobperron, line 32: Cool, I didn't know about `cancel()`. I guess the race is if the executor is constructed and quickly destructed. Instead of sleeping here, alternatively, we could block in the destructor using an additional flag (we are already blocking there with `spin_thread_.join()`), e.g.

```c++
Executor::~Executor()
{
  while (!this->thread_done_) {
    this->shutdown();
    // Or maybe std::condition_variable would be slightly better?
    std::this_thread::sleep_for(100ms);
  }
  spin_thread_.join();
}

void Executor::run()
{
  spin();
  this->thread_done_ = true;
}
```

I don't think we have to worry about asynchronous calls to `Executor::shutdown()` since it is private.

Either way, it's still a not a great hack.

gazebo_ros/src/executor.cpp
```diff
@@ -22,15 +22,22 @@ namespace gazebo_ros
 Executor::Executor()
 : spin_thread_(std::bind(&Executor::run, this))
 {
+  using namespace std::chrono_literals;
   sigint_handle_ = gazebo::event::Events::ConnectSigInt(std::bind(&Executor::shutdown, this));
+  while (!this->spinning) {
+    // TODO(ivanpauno): WARN Terrible hack here!!!!
+    // We cannot call rclcpp::shutdown asynchronously, because it generates exceptions that
+    // cannot be caught properly
```
jacobperron: It'd be nice to reference the rclcpp issue here in the comment.

jacobperron (commented):
I haven't tried it out yet, but makes sense to me.

gazebo_ros/src/executor.cpp
ivanpauno: >     // Or maybe std::condition_variable would be slightly better?

If you use a condition variable, you still have a race because of the time that happens between you trigger the condition variable and you call spin.

>  Instead of sleeping here, alternatively, we could block in the destructor using an additional flag (we are already blocking there with spin_thread_.join()), e.g.

Sounds good to me

ivanpauno: >> Instead of sleeping here, alternatively, we could block in the destructor using an additional flag (we are already blocking there with spin_thread_.join()), e.g.
>
> Sounds good to me

oh wait, the private `shutdown` method can be called from a signal handler, and calling sleep there doesn't sound valid (and we want the signal handler to cancel the executor).

So, I guess that I will keep blocking in the constructor (which sounds like safe to do).

jacobperron: > If you use a condition variable, you still have a race because of the time that happens between you trigger the condition variable and you call spin.

I imagined the trigger happens after `spin()`, e.g.

```
void Executor::run()
{
  spin();
  this->cv_.notify_all();
}
```

And then we condition the while-loop in the destructor on the condition variable.

gazebo_ros/src/executor.cpp
ivanpauno: See https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1212/commits/018c260a248f04876774189da6491425b4ede6da

gazebo_ros/src/executor.cpp
ivanpauno: > I imagined the trigger happens after spin(), e.g.

`spin` is blocking, so that wouldn't block.
You could try to create a "one time timer" and trigger the condition variable there, but really that's a lot of boilerplate and I'm not sure if it has some caveats or not.

jacobperron: :thinking: I was imagining the trigger happens at the end of the spin thread. It seems like it would work to me, or maybe I'm misunderstanding something about condition variables.

`spin()` blocks until the executor is canceled.
Then the spin thread notifies the main thread via the condition variable.

Meanwhile, the main thread is blocking in the destructor until the spin thread is done.

jacobperron: err, nevermind, I got it backwards. I see the issue :upside_down_face:

jacobperron (approved):
LGTM, seems like the best we can do until some refactoring happens to the rclcpp executor.

jacobperron:

ivanpauno:
@jacobperron feel free to merge when this is ready, as I don't have write access here.

jacobperron:

scpeters:
the devel job on build.ros2.org has been having lots of test failures since this was merged:

* https://build.ros2.org/job/Rdev__gazebo_ros_pkgs__ubuntu_focal_amd64/13/

any thoughts?

gazebo_ros/src/executor.cpp
```diff
@@ -41,7 +48,9 @@ void Executor::run()
 
 void Executor::shutdown()
 {
-  rclcpp::shutdown();
```
jacobperron, line 44: I suspect the issue is that we're no longer properly shutting down the context (only the executor itself). I can take a closer look.

jacobperron: Reverting this PR didn't seem to have any positive effect: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1227

ivanpauno: > Reverting this PR didn't seem to have any positive effect: #1227

I see the revert was finally closed, is the problem fixed?
I can take another look to this if not

scpeters: it was fixed by #1228

Issue #1211, opened 2021-01-13
Increase /clock frequency


iluetkeb:
**Description**

The current default frequency for /clock is 10Hz, which gives 100ms clock resolution for the "now()" call. This is *very* coarse and represents a huge difference to most real-world robots, where the usable clock precision is in the microsecond range, and also to ROS1, where it used to be 1ms, IIRC. Since all nodes need the sim_time, increasing the frequency has a cost, of course, therefore a balance needs to be found between precision and message sending overhead.

There is a parameter for changing this, but I would say that the default should be reasonable, and in my considered opinion, 10Hz is not reasonable for the reasons outlined above.

**Expected behavior**

I would suggest to increase the default frequency to 1000, to mimick ROS1 behavior.

**Environment (please complete the following information):**

  - gazebo_ros_pkgs: up to master

musamarcusso:
Trying the instructions in the wiki also seem to not work.
I created a `params.yaml` file 
```yaml
gazebo:
    ros__parameters:
        publish_rate: 50.0
```
and called as instructed [here](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-ROS-Clocks-and-sim-time). Gazebo seems to think the file is a `world` file.

```
gazebo --verbose -s libgazebo_ros_init.so __params:=params.yaml
Gazebo multi-robot simulator, version 11.3.0
Copyright (C) 2012 Open Source Robotics Foundation.
Released under the Apache 2 License.
http://gazebosim.org

[Msg] Waiting for master.
Gazebo multi-robot simulator, version 11.3.0
Copyright (C) 2012 Open Source Robotics Foundation.
Released under the Apache 2 License.
http://gazebosim.org

[Err] [Plugin.hh:178] Failed to load plugin libgazebo_ros_init.so: libgazebo_ros_init.so: cannot open shared object file: No such file or directory
[Msg] Waiting for master.
[Msg] Connected to gazebo master @ http://127.0.0.1:11345
[Msg] Publicized address: 192.168.178.25
[Msg] Connected to gazebo master @ http://127.0.0.1:11345
[Msg] Publicized address: 192.168.178.25
[Wrn] [SystemPaths.cc:459] File or path does not exist [""] [__params:=params.yaml]
[Err] [Server.cc:444] Could not open file[__params:=params.yaml]
[Wrn] [Server.cc:359] Falling back on worlds/empty.world
[...]

PR #1214, opened 2021-01-25
revert to default update as in ROS1

iluetkeb:
This reverts the clock update frequency back to the ROS 1 value of 1kHz, in response to #1211.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_init.cpp
+++ b/gazebo_ros/src/gazebo_ros_init.cpp
@@ -132,7 +132,7 @@ public:
   gazebo::transport::NodePtr gz_node_;
 
   /// Default frequency for clock publisher.
-  static constexpr double DEFAULT_PUBLISH_FREQUENCY = 10.;
+  static constexpr double DEFAULT_PUBLISH_FREQUENCY = 1000.;
 };
 
 GazeboRosInit::GazeboRosInit()
--- a/gazebo_ros/src/gazebo_ros_init.cpp
+++ b/gazebo_ros/src/gazebo_ros_init.cpp
@@ -132,7 +132,7 @@ public:
   gazebo::transport::NodePtr gz_node_;
 
   /// Default frequency for clock publisher.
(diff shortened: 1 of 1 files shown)


kamille6:
Has it been determined why the check is failing? Is there some way I could help move this along? I am in need of a faster /clock and would like to eventually not have to build gazebo_ros_pkgs from source for my project just to get a faster /clock. Thank you!

robin-read:
When trying to open the link for the failed test I get a 404 error which is a bit concerning.

This is a small and easy change/win that will certainly make the migration to ROS2 much cleaner for the community at large. To echo the comment from @kamille6, having to clone this repo to get a one line change is a bit cumbersome.

To the maintainers, please could you have a look at this PR? It would be much appreciated. 

Thanks!

bednarhonza:
Hello guys, any update on this? It would be very welcomed!

Thanks!

iluetkeb:
We did some more testing on the meantime, and anything beyond 250Hz causes simulation slowdown, so I would probably modify this a bit by now.
Also, we found that you actually _can_ pass this as an argument, by giving it on the command-line (in contrast to adding it to the models, which doesn't work). Just stick it into a YAML file and give that as a --ros-args argument when starting gzserver.

bednarhonza:
I have tried to set the rate at 250 Hz, but my simulation does not go above 160 Hz. My CPU should be able to handle this, I have the real-time factor 1.0 as well.

I'm trying to launch it through default gazebo.launch.py. Could you help me with setting the parameter via the launch file?

iluetkeb:
You can set the launch variable `extra_gazebo_args` to `--ros-args --params-file <your-yaml>`

bednarhonza:
Thanks, that works. I have set the rate to 100 Hz, but the /clock runs at 83 Hz. Is there any other reason for that?

jacobperron:
Change the default frequency sounds reasonable to me; it would be good to know the reason for the discrepancy between the desired and actual publish rate.. though I suppose we can ticket that separately.

Issue #1218, opened 2021-02-01
QoS mismatch between gazebo_ros_camera and gazebo_ros_video


BrettRD:
A QoS mismatch between camera and video plugins prevents an image topic from the camera being projected onto a surface.
The choice of QoS by the camera plugin also causes issues with rviz, rqt, and just about every other subscriber I've tried.

Pull request https://github.com/ros-simulation/gazebo_ros_pkgs/pull/926 would lead me to believe the camera plugin is the only topic in ROS that is using the correct QoS for its type, so I think this is a bug with gazebo_ros_video

**Description**
When using gazebo_ros_camera and after https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1052, the camera plugin produces a message with QoS `RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT` which is expected,
but the video pluigin subscribes with QoS `RMW_QOS_POLICY_RELIABILITY_RELIABLE`
As a result, the subscription cannot receive any messages because of the QoS incompatibility.

**To Reproduce**
Create a SDF with camera and video plugins, remap them to the same topic
```          
... one link
<plugin name="plugin_name" filename="libgazebo_ros_camera.so">
  <ros>
    <remapping>camera/image_raw:=cam_raw</remapping>
    <!-- the default topic name defies ~/ remapping but that's a different issue -->
  </ros>
... defaults
</plugin>

... another link
<plugin name="display" filename="libgazebo_ros_video.so">
  <ros>
    <remapping>~/image_raw:=cam_raw</remapping>
  </ros>
</plugin>
```
launch the world with
```
        launch.actions.ExecuteProcess(
            cmd=['gazebo', '--verbose', world]
        ),
```
inspect the topic with verbose output
```
$ ros2 topic info -v /cam_raw
Publisher count: 1
...
  Reliability: RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT
...
Subscription count: 2
...
  Reliability: RMW_QOS_POLICY_RELIABILITY_RELIABLE
...
  Reliability: RMW_QOS_POLICY_RELIABILITY_RELIABLE
```

The camera output is visible to `ros2 topic echo`, but it is not visible to rqt, and it is not displayed on the texture.

**Expected behavior**
The camera output should be visible in the texture rendered by gazebo_ros_video.

[...]

PR #1219, opened 2021-02-03
use SystemDefaultsQoS

BrettRD:
change QoS of camera and video plugins to honour system defaults unless explicitly overridden
https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1218
Should I include other plugins while I'm at it?

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
+++ b/gazebo_plugins/src/gazebo_ros_ackermann_drive.cpp
@@ -361,7 +361,7 @@ void GazeboRosAckermannDrive::Load(gazebo::physics::ModelPtr _model, sdf::Elemen
   impl_->last_update_time_ = _model->GetWorld()->SimTime();
 
   impl_->cmd_vel_sub_ = impl_->ros_node_->create_subscription<geometry_msgs::msg::Twist>(
-    "cmd_vel", qos.get_subscription_qos("cmd_vel", rclcpp::QoS(1)),
+    "cmd_vel", qos.get_subscription_qos("cmd_vel", rclcpp::SystemDefaultsQoS()),
     std::bind(&GazeboRosAckermannDrivePrivate::OnCmdVel, impl_.get(), std::placeholders::_1));
 
   RCLCPP_INFO(
@@ -375,7 +375,7 @@ void GazeboRosAckermannDrive::Load(gazebo::physics::ModelPtr _model, sdf::Elemen
   impl_->publish_odom_ = _sdf->Get<bool>("publish_odom", false).first;
   if (impl_->publish_odom_) {
     impl_->odometry_pub_ = impl_->ros_node_->create_publisher<nav_msgs::msg::Odometry>(
-      "odom", qos.get_publisher_qos("odom", rclcpp::QoS(1)));
+      "odom", qos.get_publisher_qos("odom", rclcpp::SystemDefaultsQoS()));
 
     RCLCPP_INFO(
       impl_->ros_node_->get_logger(), "Advertise odometry on [%s]",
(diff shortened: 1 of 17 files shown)


BrettRD:
I've changed every instance of `QoS(1)` and `SensorDataQoS()` to `SystemDefaultsQoS()`.
In the default foxy install, this change should have no effect on behaviour, except to relieve QoS compatibility issues elsewhere in the ROS ecosystem.

Notably, joint state publishers are using `QoS(1000)` which is clearly not default, I've left them custom.
Being reliable publishers, they shouldn't cause any QoS compatibility issues.

BrettRD:
There's a couple of failed tests that seem unrelated.

`joint_state_publisher` which is unchanged in this pull request, has an initialisation issue in its test, probably to do with discarded history when someone set its QoS to `SensorDataQoS`.
https://github.com/ros-simulation/gazebo_ros_pkgs/blob/foxy/gazebo_plugins/test/test_gazebo_ros_joint_state_publisher.cpp#L65
The subscription happens after the executor would have passed the messages on, and it won't receive old messages unless the QoS is transient_local.

`test_node` seems to be suffering from generic neglect; deprecated SDF version, and broken ros init

jacobperron (commented):
I was actually imagining we only change instances of `SensorDataQoS` to `QoS(5)` (the ROS default with custom depth), instead of the *system* default QoS profile. The system default profile will inherit the default settings of the underlying RMW. This doesn't guarantee that the QoS settings will match other subs/pubs using the ROS default.

For best out-of-the box experience, I think it would be better to stick with the ROS default (e.g. `rclcpp::QoS(5)` as [the depth of `SensorDataQoS` is 5](https://github.com/ros2/rmw/blob/09c8fc3d07485ef8cc47caebd4be2561d82b7b08/rmw/include/rmw/qos_profiles.h#L28)).

jacobperron:
The `test_node` failure is known and should be fixed by https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1212 (pending backport to foxy).

BrettRD:
> I was actually imagining we only change instances of `SensorDataQoS` to `QoS(5)` (the ROS default with custom depth), instead of the _system_ default QoS profile. The system default profile will inherit the default settings of the underlying RMW. This doesn't guarantee that the QoS settings will match other subs/pubs using the ROS default.
> 
> For best out-of-the box experience, I think it would be better to stick with the ROS default (e.g. `rclcpp::QoS(5)` as [the depth of `SensorDataQoS` is 5](https://github.com/ros2/rmw/blob/09c8fc3d07485ef8cc47caebd4be2561d82b7b08/rmw/include/rmw/qos_profiles.h#L28)).

That sounds a lot like we're making an explicit choice to deprecate most of the tuning features exposed in the RMWs.
https://github.com/ros2/rmw_fastrtps/blob/master/README.md

If we use system defaults in nodes, then foxy ships with a built-in per-topic performance tuning system.
Fast-dds ships out of the box with defaults equivalent to `QoS(1)`.

I agree that it would be better to have something vendor vendor agnostic, but I feel that's better achieved by requiring that the RMW provide ROS2's (or the user's) preferred defaults to the DDS layer.

BrettRD:
related:
https://github.com/ros2/rclcpp/pull/1508

This line configures ROS2 to systematically ignore DDS XML level defaults.
https://github.com/ros2/rclcpp/blob/foxy/rclcpp/include/rclcpp/qos.hpp#L67
changing `rmw_qos_profile_default` to `rmw_qos_profile_system_default` allows most of ROS to honour the XML feature with `QoS(N)` while still correctly setting keep_last and depth. users on stock Fastdds won't notice.

jacobperron:
It was a conscious design decision in the early development of ROS 2 to not use the system default QoS. [This design article](http://design.ros2.org/articles/qos.html) claims the reason was to provide a default that most resembles ROS 1's behavior. I suspect there may be other motivation behind the decision, but if not, then I agree that we should consider changing the default profile. However, I think such a change should 1) happen on rolling, and 2) happen in the client libraries (ie. rclcpp and rclpy).

My reasoning for (1) is to avoid inadvertently breaking users in an already released distro, for example, not everyone is using the same RMW and it's difficult to guarantee that changing the QoS to "system default" will go smoothly for all. I'm less hesitant about changing from `SensorDataQoS` to `QoS(10)` since it is only making the settings more compatible (best effort -> reliable).
For (2), it seems reasonable to first get the proposal for changing the default accepted and then follow suite in other packages (like gazebo_ros).

BrettRD:
> changing from `SensorDataQoS` to `QoS(10)` since it is only making the settings more compatible (best effort -> reliable).

In that case, since `SensorDataQoS` is the recommended QoS in the design docs, changing to `rclcpp::SensorDataQoS().relaible()` for only publishers might be a little more open-textbook for new users than going back to `QoS(N)`.
It would also yield the desired behaviour when the only subscribers to the topic are using `SensorDataQoS`, and it'll make it easier for rosbag and other loggers to get guarantees over data completeness.

I'll make a new pull request.

BrettRD:
Resolved in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1224

PR #1221, opened 2021-02-04
gazebo_ros: use lxml in spawn_entity.py

scpeters:
The python `xml.etree.ElementTree` library does not handle xml namespaces well, replacing namespaces of prefixed elements
(like `<gazebo:custom_tag/>`) with `ns0`, `ns1`, etc. (like `<ns0:custom_tag/>`). I noticed this problem while debugging a model with custom elements spawned using `spawn_entity.py`.

Based on a [hint from stackoverflow](https://stackoverflow.com/a/14853417), this switches to using lxml instead, which has the same syntax and is already used by other ros2 packages (such as [urdf_parser_py](https://github.com/ros/urdf_parser_py/blob/ros2/package.xml)).


ahcorde (approved):

scpeters:
this has an approval and CI is now passing, so I'll merge

PR #1228, opened 2021-02-10
Fix tests for cyclonedds

scpeters:
There are some QoS incompatibilities in some tests that use
SystemDefaultsQoS, so this changes them to use KeepLast(1)
with transient_local instead. This fixes some of the test
failures but not all.

See [get_message_or_timeout in testing_utils.hpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_ros/include/gazebo_ros/testing_utils.hpp#L135) to see what QoS parameters the test subscribers are using.

### DIFF
--- a/gazebo_ros/test/plugins/args_init.cpp
+++ b/gazebo_ros/test/plugins/args_init.cpp
@@ -56,8 +56,9 @@ void ProperInit::Load(int argc, char ** argv)
   auto node = gazebo_ros::Node::Get();
 
   // Create a publisher
-  auto pub = node->create_publisher<std_msgs::msg::String>("test",
-      rclcpp::QoS(rclcpp::KeepLast(1)).transient_local());
+  auto pub = node->create_publisher<std_msgs::msg::String>(
+    "test",
+    rclcpp::QoS(rclcpp::KeepLast(1)).transient_local());
 
   // Run lambda every 1 second
   using namespace std::chrono_literals;
--- a/gazebo_ros/test/plugins/args_init.cpp
+++ b/gazebo_ros/test/plugins/args_init.cpp
@@ -56,7 +56,8 @@ void ProperInit::Load(int argc, char ** argv)
   auto node = gazebo_ros::Node::Get();
 
   // Create a publisher
-  auto pub = node->create_publisher<std_msgs::msg::String>("test", rclcpp::SystemDefaultsQoS());
+  auto pub = node->create_publisher<std_msgs::msg::String>("test",
+      rclcpp::QoS(rclcpp::KeepLast(1)).transient_local());
 
   // Run lambda every 1 second
   using namespace std::chrono_literals;
--- a/gazebo_plugins/test/test_gazebo_ros_joint_state_publisher.cpp
+++ b/gazebo_plugins/test/test_gazebo_ros_joint_state_publisher.cpp
@@ -56,14 +56,16 @@ TEST_F(GazeboRosJointStatePublisherTest, Publishing)
   executor.add_node(node);
 
   // Step a bit before starting
-  world->Step(500);
-  executor.spin_once(500ms);
+  for (unsigned int i = 0; i < 10; ++i) {
+    world->Step(1);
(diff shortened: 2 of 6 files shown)


jacobperron (approved):
LGTM. Even though `foxy` is not suffering from the same failures, it still seems like a good idea to backport this fix.

I think these test failures started happening because the default RMW changed for ROS 2 recently. I think this is the relevant release: https://github.com/ros/rosdistro/pull/28205

gazebo_ros/test/plugins/args_init.cpp
```diff
@@ -56,7 +56,8 @@ void ProperInit::Load(int argc, char ** argv)
   auto node = gazebo_ros::Node::Get();
 
   // Create a publisher
-  auto pub = node->create_publisher<std_msgs::msg::String>("test", rclcpp::SystemDefaultsQoS());
+  auto pub = node->create_publisher<std_msgs::msg::String>("test",
+      rclcpp::QoS(rclcpp::KeepLast(1)).transient_local());
```
jacobperron: I think uncrustify prefers breaking at the open parenthesis:
```suggestion
  auto pub = node->create_publisher<std_msgs::msg::String>(
    "test",
    rclcpp::QoS(rclcpp::KeepLast(1)).transient_local());
```

Same for other lines in this PR.

scpeters: thanks: f31628b

scpeters:
I think I've fixed the remaining tests; let's see how it goes in CI

scpeters:
there is just one test failure left in CI: [test_gazebo_ros_joint_state_publisher](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/test/test_gazebo_ros_joint_state_publisher.cpp)

it's a bit flaky because it sometimes passes on my machine but fails pretty consistently. Sometimes it fails to receive a message, which fails an assertion and immediately ends the test. I'm suspicious of the [subscriber using SensorDataQoS](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/test/test_gazebo_ros_joint_state_publisher.cpp#L66). After instrumenting with a print statement in the callback, it seems that when it fails, it fails to receive the first message. So the initialization code could probably be improved as well.

scpeters:
the CI failure looks like an infrastructure problem, but the other job passed

scpeters:
@jacobperron can you re-review my more recent changes?

scpeters:

scpeters:
I think the latest CI failures on drogon are related to an underlying bug in gazebo:

~~~
terminate called after throwing an instance of 'boost::wrapexcept<boost::thread_resource_error>'
  what():  boost thread: trying joining itself: Resource deadlock avoided
~~~

[Build Status](https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/41/) https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/41/

I'll open an issue in this repository to track these flaky tests and another one upstream in osrf/gazebo

jacobperron (approved):
Changes look okay to me

scpeters:
> I think the latest CI failures on drogon are related to an underlying bug in gazebo:
> 
> ```
> terminate called after throwing an instance of 'boost::wrapexcept<boost::thread_resource_error>'
>   what():  boost thread: trying joining itself: Resource deadlock avoided
> ```
> 
> [Build Status](https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/41/) https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/41/
> 
> I'll open an issue in this repository to track these flaky tests and another one upstream in osrf/gazebo

reported in https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1232

PR #1230, opened 2021-02-12
gazebo_ros_diff_drive: lock wheels when idle

dandedrick:
Previously the wheels were left to free-wheel when idle. This could
cause the robot to drift around. In order to prevent this the
enableIdleBrake can be enabled and then the wheels will be locked in
position when we are not actively moving them. The default for
enableIdleBrake is false so that existing code will continue to behave
as expected.


dandedrick:
Gazebo 9 support was added and reopened as https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1231

PR #1238, opened 2021-02-20
Deprecate -spawn_service_timeout option

jacobperron:
The spawn_entity.py script has two options for setting service timeouts:

1. '-timeout': Sets the duration to wait for all services to be ready
2. '-spawn_service_timeout': Sets the duration to wait for the spawn_entity service to be ready

(1) is available in all ROS distros since Dashing (except Eloquent for some reason).
(2) was introduced in Eloquent, and replaced (1) in Foxy as the way to control the timeout specifically for the spawn_entity service.

This change deprecates (2) and assumes (1) is used for all service timeouts again.

This change also adds smoke tests for the two options.

---

For reference, `-timeout` was added in  #1072 and #1073, and `-spawn_service_timeout` was added in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1090.

### DIFF
--- a/gazebo_ros/scripts/spawn_entity.py
+++ b/gazebo_ros/scripts/spawn_entity.py
@@ -80,7 +80,7 @@ class SpawnEntityNode(Node):
         parser.add_argument('-wait', type=str, metavar='ENTITY_NAME',
                             help='Wait for entity to exist')
         parser.add_argument('-spawn_service_timeout', type=float, metavar='TIMEOUT',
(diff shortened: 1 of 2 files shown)


jacobperron:
@Karsten1987 and @chapulina, let me know what you think about this proposal to deprecate `-spawn_service_timeout`.

ivanpauno (approved):

jacobperron:

Karsten1987 (approved):

PR #1241, opened 2021-02-26
Reverse order of rotation offset (rpyOffset) calculation in p3d and imu plugins

iche033:
See [this comment](https://github.com/osrf/sdformat/pull/500#issuecomment-785302490) in sdformat. Ran into an issue with calculation of `rpyOffset` in urdf to sdf conversion, which seems to suggest that the `rpyOffset` should be post multiplied in the gazebo ros plugins.

There isn't really much documentation on the `xyzOffset` and `rpyOffset` params and what frames they are in. All the example I found online have been using 0s for `rpyOffset`. The changes here will however affect user's code if they are manually specifying `rpyOffset`.

Signed-off-by: Ian Chen <ichen@osrfoundation.org>

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_imu.cpp
+++ b/gazebo_plugins/src/gazebo_ros_imu.cpp
@@ -20,8 +20,6 @@
  * Date: 1 June 2008
  */
 
-#include <sdf/sdf_config.h>
-
 #include <gazebo_plugins/gazebo_ros_imu.h>
 #include <ignition/math/Rand.hh>
 
@@ -116,6 +114,9 @@ void GazeboRosIMU::LoadThread()
   }
   else
     this->offset_.Rot() = ignition::math::Quaterniond(this->sdf->Get<ignition::math::Vector3d>("rpyOffset"));
+
+  if (!this->sdf->HasElement("ignition::corrected_offsets"))
+    this->correctedOffsets_ = this->sdf->Get<bool>("ignition::corrected_offsets");
 
   if (!this->sdf->HasElement("updateRate"))
   {
@@ -248,11 +249,11 @@ void GazeboRosIMU::UpdateChild()
     // apply rpy offsets
     // rotation calculation needs to be reversed for sdformat versions > 6.2.0,
     // see https://github.com/osrf/sdformat/pull/500
-#if SDF_MAJOR_VERSION < 6 || (SDF_MAJOR_VERSION == 6 && SDF_MINOR_VERSION <= 2)
-    rot = this->offset_.Rot()*rot;
-#else
+  if (this->correctedOffsets_)
     rot = rot*this->offset_.Rot();
-#endif
+  else
+    rot = this->offset_.Rot()*rot;
+
     rot.Normalize();
 
     // get Rates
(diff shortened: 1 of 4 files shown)


gazebo_plugins/src/gazebo_ros_imu.cpp
```diff
@@ -244,7 +244,7 @@ void GazeboRosIMU::UpdateChild()
     rot = pose.Rot();
 
     // apply rpy offsets
-    rot = this->offset_.Rot()*rot;
+    rot = rot*this->offset_.Rot();
```
azeey, line 253: Would it be possible to condition this based on what version of libsdformat is being used?

iche033: added version check in fca594d. It assumes that the release will be made in sdformat 6.3.0 and forward ported though. So we'll need to merge and release the [sdformat pull request](https://github.com/osrf/sdformat/pull/500) first

azeey (approved):
Looks like there's another IMU plugin (GazeboRosImuSensor) that also uses rpyOffset https://github.com/ros-simulation/gazebo_ros_pkgs/blob/fca594d99d253707cde3185f5d595f7965944d45/gazebo_plugins/src/gazebo_ros_imu_sensor.cpp#L86

but from what I can tell, the `rpyOffset` that would be used in that sensor is manually set, not automatically generated by the fixed joint reduction code in libsdformat. So it seems okay to leave it as is.

gazebo_plugins/src/gazebo_ros_imu.cpp
```diff
@@ -244,7 +246,13 @@ void GazeboRosIMU::UpdateChild()
     rot = pose.Rot();
 
     // apply rpy offsets
+    // rotation calculation needs to be reversed for sdformat versions > 6.2.0,
+    // see https://github.com/osrf/sdformat/pull/500
+#if SDF_MAJOR_VERSION < 6 || (SDF_MAJOR_VERSION == 6 && SDF_MINOR_VERSION <= 2)
```
azeey: hmm, thinking on this some more, if this runs on currently released versions of libsdformat that are greater than 6.2 (eg. v9.0), it'll do the wrong thing. It might be an ugly hack, but maybe we should have a flag in libsdformat that we check instead of the version?

iche033:
> but from what I can tell, the rpyOffset that would be used in that sensor is manually set, not automatically generated by the fixed joint reduction code in libsdformat. So it seems okay to leave it as is.

I think for p3d and the imu plugin, the `rpyOffset` can also be manually set. If not set, fixed joint reduction will generate them. If they are set, fixed joint reduction will read in the manually specified rpyOffset values and apply additional transformation on them.

hmm does that mean we should also not be touching these equations?

gazebo_plugins/src/gazebo_ros_imu.cpp
scpeters: I think injecting an SDFormat element into the plugin block would be much simpler than trying to track which versions have the fix, especially once we start forward-porting:

* https://github.com/ignitionrobotics/sdformat/pull/500#issuecomment-880180999

iche033: updated to use the injected corrected_offsets element

iche033:
as of ef708a3, the equation will only be flipped if it detects that the `ignition::corrected_offsets` tag was injected by sdformat. So this change should not break existing behavior. We also don't need to wait for an sdformat release before merging this.

gazebo_plugins/src/gazebo_ros_imu.cpp
iche033: @scpeters are you ok with the latest changes?

PR #1246, opened 2021-03-05
gazebo_ros_camera: Added accessor methods for camera properties

kjeppesen1:
Added accessor methods to gazebo_ros_camera for subclass access to camera properties.

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
@@ -22,6 +22,7 @@
 
 #include <memory>
 #include <string>
+#include <vector>
 
 namespace gazebo_plugins
 {
@@ -163,10 +164,10 @@ protected:
   void SetCameraEnabled(const bool _enabled);
 
   // Get pointer to the cameras
-  const std::vector<gazebo::rendering::CameraPtr> GetCameras() const;
+  std::vector<gazebo::rendering::CameraPtr> GetCameras() const;
 
   // Get camera name
-  const std::string GetCameraName() const;
+  std::string GetCameraName() const;
 
   // Get number of cameras
   uint64_t GetNumCameras() const;
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
@@ -18,7 +18,6 @@
 #include <gazebo/plugins/CameraPlugin.hh>
 #include <gazebo/plugins/DepthCameraPlugin.hh>
 #include <gazebo_plugins/multi_camera_plugin.hpp>
-#include <gazebo_ros/node.hpp>
 #include <std_msgs/msg/empty.hpp>
 
 #include <memory>
@@ -165,9 +164,6 @@ protected:
 
   // Get pointer to the cameras
   const std::vector<gazebo::rendering::CameraPtr> GetCameras() const;
-
-  // Get pointer to the ros node
-  const gazebo_ros::Node::SharedPtr GetRosNode() const;
 
   // Get camera name
   const std::string GetCameraName() const;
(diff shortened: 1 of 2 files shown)


kjeppesen1:
@scpeters @jacobperron let me know if you have any questions or suggestions

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -838,6 +838,26 @@ void GazeboRosCamera::OnTrigger(const std_msgs::msg::Empty::SharedPtr)
...
 }
 
+const std::vector<gazebo::rendering::CameraPtr> GazeboRosCamera::GetCameras() const
+{
+  return impl_->camera_;
+}
+
+const gazebo_ros::Node::SharedPtr GazeboRosCamera::GetRosNode() const
+{
+  return impl_->ros_node_;
```
scpeters: just curious, why do you need the ROS node pointer? are you planning to create extra publishers?

kjeppesen1: Yes my subclass creates a subscriber to toggle the lens caps. Do you have a better solution in mind?

jacobperron: I'm not sure if it works, but you might be able to call `gazebo_ros::Node::Get(_sdf);` in the subclasses `Load` method. Then you'll also be able to parse any custom SDF tags specific to your plugin.

kjeppesen1: @scpeters @jacobperron Thanks for the suggestion. I can confirm that `gazebo_ros::Node::Get(_sdf);` does work. I have updated the PR to remove the ros node accessor.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
```diff
@@ -162,6 +162,15 @@ class GazeboRosCamera
   /// param[in] _enabled True to enable.
   void SetCameraEnabled(const bool _enabled);
 
+  // Get pointer to the cameras
+  const std::vector<gazebo::rendering::CameraPtr> GetCameras() const;
+
+  // Get camera name
+  const std::string GetCameraName() const;
```
jacobperron: What advantage is there in returning const values here? I guess returning const references might save on some copying, but I think it's fine to return non-const values.

kjeppesen1: That was to convey the design intent that these values are not meant to modified, but rather just be retrieved for reference. Let me know if you see no value in that and I can delete that specifier.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
```diff
@@ -162,6 +162,15 @@ class GazeboRosCamera
   /// param[in] _enabled True to enable.
   void SetCameraEnabled(const bool _enabled);
 
+  // Get pointer to the cameras
+  const std::vector<gazebo::rendering::CameraPtr> GetCameras() const;
```
scpeters: cpplint wants an `#include <vector>` at the top of this file due to this addition

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
jacobperron: IIUC, these are being returned by value, so there's nothing wrong with (or preventing) the caller modifying their local copy.

Also, I think it's potentially more efficient to remove the const so the compiler can ~~treat the return value as an rvalue~~ take advantage of move semantics.

scpeters: I think he meant that you could return `const std::string&`?

jacobperron: After re-educating myself about move semantics, I think it's generally considered a bad idea to return reference types. I would simply remove the const from the return types.

kjeppesen1: Removed const from the return types

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
kjeppesen1: Added

jacobperron (approved):

Issue #1232, opened 2021-02-12
Flaky tests fail with boost::thread_resource_error exceptions


scpeters:
**Description**

There are sometimes flaky test failures involving gazebo with `boost::thread_resource_error` exceptions. I see these most consistently with the [drogon node](https://build.osrfoundation.org/computer/linux-aws_drogon.nv.xenial/) on build.osrfoundation.org

**To Reproduce**

It's flaky, so it's not easy to reproduce. I've observed it in the `test_gazebo_ros_diff_drive` and `test_gazebo_ros_harness` in the following job:

* https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/41/testReport/junit/(root)/gazebo_plugins/test_gazebo_ros_diff_drive_gtest_missing_result/
* https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/41/testReport/junit/(root)/gazebo_plugins/test_gazebo_ros_harness_gtest_missing_result/

**Expected behavior**

The tests should pass.

**Environment (please complete the following information):**

 - OS: 20.04 in CI
 - Gazebo version: 11.3.0
 - ROS version: rolling
  - gazebo_ros_pkgs: CI build
 - glx server in use: not sure, it's on drogon

**Additional context**

~~~
9: [ RUN      ] GazeboRosDiffDrive/GazeboRosDiffDriveTest.Publishing/0
9: [INFO] [1613153117.538449261] [test.diff_drive]: Wheel pair 1 separation set to [1.250000m]
9: [INFO] [1613153117.538679648] [test.diff_drive]: Wheel pair 1 diameter set to [0.600000m]
9: [INFO] [1613153117.552771257] [test.diff_drive]: Subscribed to [/test/cmd_test]
9: [INFO] [1613153117.557785013] [test.diff_drive]: Advertise odometry on [/test/odom_test]
9: [INFO] [1613153117.564436068] [test.diff_drive]: Publishing odom transforms between [odom_frame_test] and [chassis]
9: [INFO] [1613153117.564487945] [test.diff_drive]: Publishing wheel transforms between [chassis], [left_wheel_joint] and [right_wheel_joint]
9: [Msg] Waiting for master.
9: [Msg] Connected to gazebo master @ http://127.0.0.1:11345
9: [Msg] Publicized address: 172.17.0.2
[...]

PR #1249, opened 2021-03-10
Fix line length in gazebo_ros/test/CMakeLists.txt

scpeters:
I noticed a complaint about line length in `gazebo_ros/test/CMakeLists.txt` in CI for #1246. Let's see if CI is happy with this.


ahcorde (approved):

jacobperron (approved):
A bug was recently fixed in ament_lint so that line lengths in CMake are actually enforced now (https://github.com/ament/ament_lint/pull/236). I suspect this is why we are seeing this now.

scpeters:
flaky test looks like https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1232; I think this is ok to merge

PR #1258, opened 2021-03-20
Avoid rejecting QoS overrides parameters

jacobperron:
Some parameters are declared by the node and rejecting them can cause the gazebo plugin to crash.

See general discussion about best practice in https://github.com/ros2/rclcpp/issues/1587
IMO, we don't need to guard against unknown parameters, since they must be explicitly declared in order for us to get a callback in the first place.


ivanpauno (approved):

jacobperron:
The test failure seems unrelated. See https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1259 for a proposed fix.

PR #1276, opened 2021-05-21
[JointStatePublisher] remove unused tf_prefix_

yoshito-okada:
In the joint state publisher plugin, the member variable tf_prefix_ exists but never used.  I simply remove it.


ahcorde (approved):

yoshito-okada:
I also cleaned up include files because there was <tf/transform_*.h>

Issue #1268, opened 2021-04-26
[gazebo_ros]: 'Step control' of ROS-Gazebo for ROS-based Reinforcement Learning


alikureishy:
**Description**
This is a feature request for supporting step control of Gazebo, via ROS. Gazebo independently offers step control via the gazebo::msgs::WorldControl message, but that functionality is not presently exposed to ROS nodes via the gazebo_ros_api plugin.

**Motivation**:
This is mostly in the context of ROS-based machine-learning systems (specifically reinforcement learning, but applicable generally to ML as well). In the case of RL, AI agents must be able to associate actions with their consequent observations and rewards, as part of the reward-based feedback loop. However, since Gazebo and ROS (indeed all nodes within ROS) operate asynchronously with respect to each other, it is presently not possible to draw time-synchronized or time-quantized action-state-reward associations, which impacts both training and inference stages of the RL system.

alikureishy:
Issue was brought up earlier in 2016 (https://github.com/ros-simulation/gazebo_ros_pkgs/issues/489). A potential solution was provided as a workaround in the comments, but I feel this can be fixed in the gazebo_ros plugin directly. Also made the feature request a bit more comprehensive with this ticket.

alikureishy:
If this is an acceptable feature, checking it into ROS1 might allow it to be picked up automatically during any pending [ROS2 port of gazebo_ros_api_plugin.hpp/cpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/.ros1_unported/gazebo_ros/include/gazebo_ros/gazebo_ros_api_plugin.h) as well.

alikureishy:
For anyone who would like to see this feature, please upvote on the original description above, or add your thoughts via a comment.

safdark:
Feature [implementation](https://github.com/safdark/gazebo_ros_pkgs/tree/issue_1268). Will submit a pull request when ready.

alikureishy:
Implemented a workaround [PR](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1280) to unblock myself. Posting here in case others find it helpful too.

Kettenhoax:
[...]

PR #1280, opened 2021-05-31
[gazebo_ros_api_plugin]: 'Step control' of ROS-Gazebo for ROS-based Reinforcement Learning (Issue # 1268)

alikureishy:
This is the implementation for issue # 1268 that I had logged a month ago.

**Description**
This PR adds support for step control of Gazebo, via ROS. Gazebo independently offers step control via the gazebo::msgs::WorldControl interface, but that functionality was not previously exposed to ROS nodes via the gazebo_ros_api plugin. This PR closes that gap.

**Motivation**
Mostly in the context of ROS-based machine-learning systems (specifically reinforcement learning, but applicable generally to ML as well). In the case of RL, AI agents must be able to associate actions with their consequent observations and rewards, as part of the reward-based feedback loop. However, since Gazebo and ROS (indeed all nodes within ROS) operate asynchronously wrt each other, it is presently not possible to draw time-synchronized or time-quantized action-state-reward associations, which impacts both training and inference stages of the RL system. This 'step control' feature, invokable directly by agents running as ROS nodes (or scripts/actions/services), addresses this need. An ML agent (running as a ROS node, for example) can now collect and accurately associate action-state-reward tuples across each step (or set of steps).

**Proposed fix**
- gazebo_ros_api_plugin.h exposes a new ROS-facing service ('step_control') with corresponding [StepControl.srv](https://github.com/safdark/gazebo_ros_pkgs/blob/600290fe4d72262a6baaff2c7842ec828106812a/gazebo_msgs/srv/StepControl.srv) message.
- A request for >0 steps pauses the simulation and advances the simulator by the specified # of steps. A request of 0 steps unpauses the simulator.
- The implementation utilizes the [WorldControl](https://github.com/osrf/gazebo/blob/c09a427cc6432de3a2cf761f09f25fb16dd0b0ae/gazebo/msgs/world_control.proto) interface in gazebo that pauses and steps the gazebo simulator.
- Effective step duration is determined by the 'max_step_size' parameter in gazebo's Physics properties (Default 0.001 seconds).


alikureishy:
In reference to the CI failure below ("Npr__gazebo_ros_pkgs__ubuntu_focal_amd64"), that's actually happening on the noetic-devel branch itself, and is unrelated to this PR.

alikureishy:
All: This is my first PR submission, so plz overlook any faux pas below.

@davetcoleman @j-rivero @scpeters @jbohren @adityapande-1995 : My previous comments about the step-control [issue](https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1268) from April 2021 (and _this_ PR in May 2021) have not received any responses (>3 months). Though I followed the contribution guidelines, is there something else I need to do to move this PR forward for merge consideration?

(Note: The failing "check" has been there since prior to this PR)

alikureishy:
https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1268#issuecomment-954911830

Haitong-Wang:
This is exactly what I need now, thanks for your work @alikureishy !

When I use your code, I found a small issue. After calling the service of step control, it would be better to wait until those steps are actually taken by Gazebo before collecting robot observations and computing rewards (and any other computation). Not sure what is the proper way to make sure those steps are taken, my solution is that after calling step control service, keep checking current ros time in a loop until it does not change.

alikureishy:
> When I use your code, I found a small issue. After calling the service of step control, it would be better to wait until those steps are actually taken by Gazebo before collecting robot observations and computing rewards (and any other computation). Not sure what is the proper way to make sure those steps are taken, my solution is that after calling step control service, keep checking current ros time in a loop until it does not change.

Glad it's what you need, @Haitong-Wang ! The step control invocation will only return ***after*** gazebo has taken those requested steps. The ```WorldControl``` feature in Gazebo, which is invoked by the ```StepControl``` wrapper API, advances Gazebo by the requested # of steps and then pauses gazebo.

[Correction]
So, any impacts from the stepped time will have been posted on the relevant topics between the instant the  ```StepControl``` call is invoked, until the next ```StepControl``` invocation is made by your implementation (which depends on your agent implementation). More behavior details are described at the end of the opening comment of this PR.

Does that answer your question?

Haitong-Wang:
@alikureishy Thanks for your reply. When I tried to get ROS time before and after calling `StepControl` service, I found that the two time are basically the same. However, the interval (difference between the time after and before calling `StepControl` service) was supposed to equal the step size that I set if Gazebo had taken those steps. That's why I felt that my program should wait for Gazebo taking those steps.

alikureishy:
Have you enabled ```use_sim_time=1``` in your node confirguration? And,
have you tried using a higher step size in the StepControl call? Gazebo
itself might not update the clock that often..so you might have to advance
it several steps before it sends another clock update. Try larger step
sizes (maybe 500) and see if it updates the time? Alternatively, change the
clock publish frequency for gazebo so it updates at the right frequency (
/gazebo/pub_clock_frequency).

Daviddeer2:
> @alikureishy Thanks for your reply. When I tried to get ROS time before and after calling `StepControl` service, I found that the two time are basically the same. However, the interval (difference between the time after and before calling `StepControl` service) was supposed to equal the step size that I set if Gazebo had taken those steps. That's why I felt that my program should wait for Gazebo taking those steps.

Hi, I meet the same problem. It seems service "step_control" response immediately after calling, without waiting simulation. Have you solve it yet?

PR #1282, opened 2021-06-01
Fix warnings when building against the latest sources.

clalancette:
Post-Galactic, we have deprecated the tf2_geometry_msgs.h
header file in preference to tf2_geometry_msgs.hpp.  While
we are in here, also remove a command-line option that gcc
does not understand.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

Note that we are going to need to branch off `galactic` to merge this, as this is not compatible with Galactic.  Also FYI @karsten1987, this is partially undoing an earlier commit from you.

### DIFF
--- a/gazebo_plugins/CMakeLists.txt
+++ b/gazebo_plugins/CMakeLists.txt
@@ -8,7 +8,7 @@ endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   # we dont use add_compile_options with pedantic in message packages
   # because the Python C extensions dont comply with it
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic -Wno-inconsistent-missing-override")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic")
 endif()
 
 if(WIN32)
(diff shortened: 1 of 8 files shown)


ahcorde (approved):

clalancette:
Closing and reopening to reset the CI jobs.  They should be much more likely to build now (at least on Rolling, which is the only place this should be merged to).

clalancette:
So things are building now, but the tests are failing.  It looks like some of the tests are failing with heap corruption (`malloc(): corrupted top size`).  I ran a couple of them through `valgrind` locally.  While there are some warnings shown by valgrind (mostly around `Conditional jump or move depends on uninitialized value(s)` from gazebo_rendering), I wasn't able to see any of the heap corruption or reproduce the problem.  I'm not quite sure what is going on with those tests, though I do think they are unlikely to be related to this PR.

How should we proceed here?

jacobperron:
There haven't been any recent changes to gazebo_ros_pkgs that would explain the failures. My best guess is that there was an update to Gazebo that has started to cause these failures.

@j-rivero Any ideas?

clalancette:

clalancette:
The heap problem was an issue in rclcpp, which is now fixed (at least in Rolling; we are still working on the backport to Galactic).  If you look at https://build.ros2.org/job/Rpr__gazebo_ros_pkgs__ubuntu_focal_amd64/98/ , you can see that almost all of the tests pass now, with the exception of one.  While that failure is a real failure, I doubt it is because of this PR.  Separately, I don't know why it isn't updating the checks here; something seems to be wonky with the Github hooks.

Regardless, I think that is sufficient to merge this in, as it is a clear improvement.  I'm going to go ahead and do that.

Issue #1272, opened 2021-05-04
spawn_entity.py cannot publish URDF meshes to Gazebo11


newcanopies:
UR10 urdf xacro is spawned as a model, but meshes are not displayed:
![image](https://user-images.githubusercontent.com/65022111/117045957-67f31d00-ad10-11eb-99bd-21ddf5da5440.png)
Missing ur10 meshes
![image](https://user-images.githubusercontent.com/65022111/117045997-72adb200-ad10-11eb-9e63-eea1d56945de.png)

.
.
.
.

FelipeGdM:
Hi @newcanopies!

Could you print the value of `GAZEBO_MODEL_PATH` and `GAZEBO_RESOURCE_PATH` environment variables?

They store the list of directories where Gazebo will search for files to build your model (see the [guide](http://gazebosim.org/tutorials?tut=components&cat=get_started#EnvironmentVariables) for more). 

> [Wrn] [SystemPaths.cc:459] File or path does not exist [""] [model://ur10_gazebo/models/ur10/meshes/dae/wrist3.dae]
[Err] [MeshShape.cc:64] Failed to find mesh file [model://ur10_gazebo/models/ur10/meshes/dae/wrist3.dae]
[Err] [Plugin.hh:178] Failed to load plugin libgazebo_ros_control.so: 
libgazebo_ros_control.so: cannot open shared object file: No such file or directory
[Err] [Visual.cc:2956] No mesh specified
[Wrn] [SystemPaths.cc:459] File or path does not exist [""] [model://ur10_gazebo/models/ur10/meshes/dae/upperarm.dae]

My initial guess is that this URI `model://ur10_gazebo/models/ur10/meshes/dae/wrist3.dae` is bad formatted or `GAZEBO_MODEL_PATH` was not updated accordingly. Usually, there is no `models/` folder within a model folder, it seems that a `package://` was translated to `model://` (maybe during urdf > sdf convertion?). Strictly speaking, it is not wrong, but it defines how `GAZEBO_MODEL_PATH` should be updated

> [Err] [Visual.cc:2956] No mesh specified
[Wrn] [SystemPaths.cc:459] File or path does not exist [""] [$(find ur10)/models/ur10/meshes/stl/upperarm.stl]

This is strange, because Gazebo should not "see" the `$(find ur10)` directive, since it is a `.xacro` command. Have you updated the `.xacro` file after it was parsed?

newcanopies:
[...]

Issue #1236, opened 2021-02-18
[ROS2] Could not spawn URDF model


atinfinity:
**Description**

I tried to spawn URDF model using gazebo_ros.
But, I could not spawn URDF model.

**To Reproduce**

I used <https://github.com/ros/urdf_tutorial/blob/master/urdf/05-visual.urdf>.

```
ros2 launch gazebo_ros gazebo.launch.py
ros2 run gazebo_ros spawn_entity.py -file $HOME/dev_ws/src/urdf_tutorial/urdf/05-visual.urdf -entity 05
```

The message of this command is as follows.

```
$ ros2 run gazebo_ros spawn_entity.py -file $HOME/dev_ws/src/urdf_tutorial/urdf/05-visual.urdf -entity 05
[INFO] [1613667021.959447709] [spawn_entity]: Spawn Entity started
[INFO] [1613667021.960142006] [spawn_entity]: Loading entity XML from file /home/dandelion/dev_ws/src/urdf_tutorial/urdf/05-visual.urdf
[INFO] [1613667021.962618085] [spawn_entity]: Waiting for service /spawn_entity, timeout = 5
[INFO] [1613667021.963103341] [spawn_entity]: Waiting for service /spawn_entity
[INFO] [1613667021.967460842] [spawn_entity]: Calling service /spawn_entity
[INFO] [1613667022.131465118] [spawn_entity]: Spawn status: SpawnEntity: Successfully spawned entity [05]
```

**Expected behavior**

I expected that `05-visual.urdf` spawns in Gazebo world.

**Screenshots**

**Environment (please complete the following information):**

 - OS: Ubuntu 20.04
 - Gazebo version: Gazebo 11.3.0
 - ROS version: ROS2 foxy
  - gazebo_ros_pkgs: debian install
 - glx server in use: output from `glxinfo | grep 'server glx vendor'` (glxinfo is in mesa-utils package for Debian/Ubuntu)

**Additional context**

I found related issue.

- https://github.com/ros2/demos/pull/349
- https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1191

newcanopies:
I have this same issue on the same Environment. 

~/ws_ROS2/src/urdf_tutorial$ ros2 run gazebo_ros spawn_entity.py -file $HOME/ws_ROS2/src/urdf_tutorial/urdf/05-visual.urdf -entity 05

```
[...]

Issue #1191, opened 2020-11-11
Cannot spawn mesh data in XML file using Spawn_Entity.py


Drojas251:
**Description**
I am trying to spawn a robot model, described in a urdf, into gazebo 11 from a ros2 python launch file by using the spawn_entity.py node. I am publishing the xml data onto a topic with robot_state_publisher and reading that data from the topic with the spawn_entity node. I am able to spawn the robot successfully if the urdf is made up of built in geomotires ( box, cylinder..), however when I use a mesh to build my robot, spawn_entity node does now spawn those links with a mesh file in gazebo. I have tried multiple files types ( dae, stl, obj) and tried different urdf's.  

**To Reproduce**
I have used the launch file from gazebo_ros2_control_demos exactly. The only changes I made was just adding an additional link and joint to the urdf file that was used in that demo, and used a mesh for that link. I have provided a screen shot below. 
`

    xacro_file = os.path.join(gazebo_ros2_control_demos_path,
                              'urdf',
                              'test_cart_position.xacro.urdf')

    doc = xacro.parse(open(xacro_file))
    xacro.process_doc(doc)
    params = {'robot_description': doc.toxml()}

    node_robot_state_publisher = Node(
        package='robot_state_publisher',
        executable='robot_state_publisher',
        output='screen',
        parameters=[params]
    )

    spawn_entity = Node(package='gazebo_ros', executable='spawn_entity.py',
                        arguments=['-topic', 'robot_description',
                                   '-entity', 'cartpole'],
                        output='screen')

`

**Expected behavior**

I have made sure my files are in the shared folder and that everything is linked properly. If I can spawn a urdf with basic shapes, then I know that my launch file and the node is working as it should. However, I suspect that the mesh information is not completely being transferred from the robot_state_publisher topic to the spawn_entity node. 

**Screenshots**

Here is the portion of the URDF that includes the mesh. I have simply added this link and joint to the test_cart_position.xacro.urdf file in gazebo_ros2_control_demo. The result in gazebo does not display the mesh model. As I said before, I tried multiple file types. 

![Screenshot from 2020-11-10 17-14-10](https://user-images.githubusercontent.com/52464724/98753001-38bead00-2378-11eb-9ccd-923c5ba5214d.png)

**Environment (please complete the following information):**

 - OS: [Ubuntu 20.04]
 - Gazebo version: [Gazebo 11]
 - ROS version: [foxy ]
[...]

Issue #1112, opened 2020-05-29
spawn_entity.py  & libgazebo_ros_factory.so cannot find models


JaehyunShim:
Describe your issue here and explain how to reproduce it.
```sh
1. Prepare your urdf file with <mesh filename="package://~"> lines
2. gazebo --verbose -s libgazebo_ros_factory.so
3. ros2 run gazebo_ros spawn_entity.py -entity myentity -x 0 -y 0 -z 0 -file /path_to_urdf_file/myrobot.urdf
4. Then you will see error codes with "Unable to download model[model://~]". 
I didn't see this problem when I read my urdf file and displayed it on rviz2 but only when I tried to spawn it on Gazebo.
```

A clear and concise description of what the bug or feature is.

```sh
Reading urdf file with <mesh filename="package://~"> and spawning the robot model on gazebo does not work with libgazebo_ros_factory.so.
```

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**

 - OS: Ubuntu 18.04
 - Gazebo version:  Gazebo 9
 - ROS version: Dashing
  - gazebo_ros_pkgs: debian install
 - glx server in use: output from `glxinfo | grep 'server glx vendor'` (glxinfo is in mesa-utils package for Debian/Ubuntu)

**Additional context**

Add any other context here.

joe28965:
So this has to do with the difference of setup between ROS1 and ROS2. Gazebo will look in the paths it has available for the files. You basically need to tell it where to find it. In the link below it explains how to do just that.

https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Gazebo-ROS-Paths

Just to give you a longer explanation, you locate where your meshes are. Let's assume your package looks something like this:

```
my_package
   meshes
      my_mesh.stl
   package.xml
```

In your package, you need to add the following:

```
  <export>
    <build_type>ament_cmake</build_type>
    <gazebo_ros gazebo_model_path="${prefix}/meshes"/>
  </export>
```

The `${prefix}` points towards the prefix of package.xml, so by adding `/meshes` do you point towards the path of your mesh. When you run the entity spawner now, it should be able to find your file.

If this solves your problem, please let us know so we can close the issue.

JaehyunShim:
@joe28965 

It still does not work and now I have two questions.
[...]

PR #1284, opened 2021-06-12
[gazebo_ros_factory]: Resolve package:// urls

bjsowa:
In ROS2 the `SpawnEntity` method let's the SDF library handle the URDF conversion, which is good in my opinion, but introduces a problem. The `sdf::readString` method changes all `package://` urls into `model://` so they can be resolved by Gazebo. This, in turn, makes it very problematic to spawn URDF model with `package://` references without maintaining a separate SDF version of the model.
In contrast, the `gazebo_ros_api_plugin` from ROS1, [replaces the `package://` references](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/231a7219b36b8a6cdd100b59f66a3df2955df787/gazebo_ros/src/gazebo_ros_api_plugin.cpp#L664-L693) with full resolved paths when converting the model.

This PR somewhat restores this behavior by doing the following steps before `sdf::readString`:
1. Convert the XML String into XML Document (using tinyxml2).
2. Iterate through all elements of the document and resolve attributes that start with `package://` into full paths.
3. Convert the XML Document back to String

### Related issues
#1272 
#1236
#1191 
#1112
LeoRover/leo_simulator#5

### DIFF
--- a/gazebo_ros/src/gazebo_ros_factory.cpp
+++ b/gazebo_ros/src/gazebo_ros_factory.cpp
@@ -56,12 +56,6 @@ public:
     gazebo_msgs::srv::GetModelList::Request::SharedPtr,
     gazebo_msgs::srv::GetModelList::Response::SharedPtr res);
 
-  /// Recursively iterate over elements of XML document and substitute all attribute values
-  /// that start with package:// with the resolved absolute path to the file.
-  /// \param[in] elem The root element of the document
-  /// \returns True on success, false on error.
-  bool SubstPackage(tinyxml2::XMLElement * elem);
-
   /// \brief Function for inserting an entity into Gazebo from ROS Service Call.
   /// Supported formats are SDF and URDF and works with both models and lights.
   /// \param[in] req Request
@@ -76,6 +70,15 @@ public:
   void DeleteEntity(
     gazebo_msgs::srv::DeleteEntity::Request::SharedPtr req,
     gazebo_msgs::srv::DeleteEntity::Response::SharedPtr res);
+
+  /// Recursively iterate over elements of XML document and substitute all attribute and
+  /// text values that start with package:// with the resolved absolute path to the file.
+  /// \param[in] elem The root element of the document
+  void SubstPackage(tinyxml2::XMLElement * elem);
+
+  /// \brief Resolve single package:// url.
+  /// \returns A new text with the resolved url.
+  std::string ResolvePackage(std::string text);
 
   /// Iterate over all child elements and add <ros><namespace> tags to the <plugin> tags.
   /// \param[out] _elem SDF element.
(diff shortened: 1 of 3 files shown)


bjsowa:
I also added resolving for text values (apart from attributes), so now it should be also possible to use package:// urls for `uri` tags in SDF.

bjsowa:
Can someone review this PR? It solves a common issue with spawning URDF in Gazebo. I can write some tests to show the use cases.

p0rys:
I had exactly the mentioned issus when simulating my robot. Finally a fix and it is working like a charm, thanks you!

LGTM. I would really appriciate when this commit is merged and released with the next sync (foxy in my case)

fmauch:
Is this being addressed? We ran into that issue setting up packages for UniversalRobots in https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/11. Supporting the `package://` notation seems quite necessary, given that it is the way `urdf` suggests specifying mesh paths.

shonigmann:
I'm also in support of adding this functionality, for improved consistency with ROS1 and as a necessary solution for pure URDF models. 

I just want to add a workaround that I've been using in my Xacro files that others might find useful while we wait for this functionality to be added.

Instead of specifying URIs using `package://package_name/urdf/my_file.urdf` notation, you can use `$(find package_name)/urdf/my_file.urdf`, which will evaluate to the full path when xacro generates the URDF.

fmauch:
> Instead of specifying URIs using `package://package_name/urdf/my_file.urdf` notation, you can use `$(find package_name)/urdf/my_file.urdf`, which will evaluate to the full path when xacro generates the URDF.

Just a quick addition: As discussed in the PR mentioned [above](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1284#issuecomment-1023674241) this effectively destroys the ability to use the URDF on a remote machine, e.g. by starting RViz on a remote machine to view the robot model. As the `$(find ...` syntax translates the meshes to absolute paths inside the URDF, those meshes will most probably not be found on the remote machine.

bjsowa:
> this effectively destroys the ability to use the URDF on a remote machine, e.g. by starting RViz on a remote machine to view the robot model. As the `$(find ...` syntax translates the meshes to absolute paths inside the URDF, those meshes will most probably not be found on the remote machine.

This is exactly why I decided to write this PR. I really don't like the idea of having to keep separate versions of the URDF for simulation and for the real robot.
I hope the maintainers will notice it at some point even though it's been almost 8 months without any feedback.

jlack1987:
any movement here? I just got to using ros2 + gazebo and fixing this would be a huge help

joe28965:
So I want to point something out here. I am not going to comment on the work and whether or not it's needed. That's a discussion for somebody else.

However, replacing `package://PACKAGE_NAME/meshes/mesh.stl` with `${find PACKAGE_NAME}/meshes/mesh.stl` works in both RViz2 and Gazebo. Small tip for those coming across this PR and need a fix.

p0rys:
> So I want to point something out here. I am not going to comment on the work and whether or not it's needed. That's a discussion for somebody else.
> 
> However, replacing `package://PACKAGE_NAME/meshes/mesh.stl` with `${find PACKAGE_NAME}/meshes/mesh.stl` works in both RViz2 and Gazebo. Small tip for those coming across this PR and need a fix.

@joe28965 did you read any of the comments above? This "solution" was already mentioned and fmauch made clear while this will not fix  the problem for everyone.

joe28965:
Oh damn, you're right, I might need to visit the optician apparently...

daisukes:
I want this to be merged! I just started to migrate my ROS project to ROS2 then got the same issue here.
Is `ros2` branch of this repo maintained by @jacobperron?
I just wanted to make sure the right person is aware of this PR. Thanks.

gazebo_ros/src/gazebo_ros_factory.cpp
```diff
@@ -344,6 +379,49 @@ void GazeboRosFactoryPrivate::DeleteEntity(
...
+void GazeboRosFactoryPrivate::SubstPackage(tinyxml2::XMLElement * element)
+{
+  if (!element) {
+    return;
+  }
+
+  for (auto attribute = element->FirstAttribute(); attribute != nullptr;
+    attribute = attribute->Next())
+  {
+    element->SetAttribute(attribute->Name(), ResolvePackage(attribute->Value()).c_str());
```
daisukes, line 391: `ResolvePackage` can throw an exception when the package cannot be found or its URL is invalid.
This will stop spawning at the first exception, but I think just providing an error log would be more friendly for the users to find what's missing. Also, it is compatible with the current behavior (the gazebo shows what it can show).

gazebo_ros/src/gazebo_ros_factory.cpp
```diff
@@ -344,6 +379,49 @@ void GazeboRosFactoryPrivate::DeleteEntity(
...
+  }
+
+  const char * text = element->GetText();
+  if (text) {
+    element->SetText(ResolvePackage(text).c_str());
```
daisukes, line 396: Here as well

daisukes (commented):
@bjsowa, I tried the code and this works well! 
I just left some comments. Please give me your thought. Thanks!

francocipollone:
Hey there! It seems like a great fix to this issue. Is there anything I can do from my end to keep pushing this PR?

PR #1286, opened 2021-06-16
Use real time

threeal:
Make plugin to use real time for the header stamp if the `use_sim_time` parameter is set to be false.


threeal:
hi could someone check why the build is failed?

PR #1291, opened 2021-06-29
gazebo_ros[2]: Add missing exec dependencies

mateus-amarante:
The libraries imported by the `gazebo_ros` [scripts](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/foxy/gazebo_ros/scripts) are not declared as dependencies. I noticed this issue when I got the following error while testing a package depending on it:
```bash
...
File "/opt/ros/foxy/lib/python3.8/site-packages/scripts/gazebo_ros_paths.py", line 33, in <module>
  from ros2pkg.api import get_package_names
ModuleNotFoundError: No module named 'ros2pkg'
```
I added the `ros2pkg` and all the other ROS packages imported in the scripts as dependencies.

I was not sure which distro I should target. I'm pushing to the oldest distro with support, which is the one I'm using.

### DIFF
--- a/gazebo_ros/package.xml
+++ b/gazebo_ros/package.xml
@@ -32,7 +32,12 @@
   <depend>std_srvs</depend>
   <depend>tinyxml_vendor</depend>
 
+  <exec_depend>ament_index_python</exec_depend>
+  <exec_depend>geometry_msgs</exec_depend>
   <exec_depend>launch_ros</exec_depend>
+  <exec_depend>python3-catkin-pkg</exec_depend>
+  <exec_depend>ros2pkg</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
 
   <build_export_depend>geometry_msgs</build_export_depend>
   <build_export_depend>sensor_msgs</build_export_depend>


gazebo_ros/package.xml
```diff
@@ -32,7 +32,12 @@
   <depend>std_srvs</depend>
   <depend>tinyxml_vendor</depend>
 
+  <exec_depend>ament_index_python</exec_depend>
+  <exec_depend>geometry_msgs</exec_depend>
   <exec_depend>launch_ros</exec_depend>
+  <exec_depend>python3-catkin-pkg</exec_depend>
```
scpeters, line 38: are we sure we want `python3-catkin-pkg` in the `foxy` branch?

mateus-amarante: It is imported here:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/3f18465d3fac30c729f9406eff0fda67ab606502/gazebo_ros/scripts/gazebo_ros_paths.py#L32

jacobperron: I suppose this is fine, though I think we could replace this logic with equivalent ament API calls in a follow-up.

jacobperron:

jacobperron (approved):

jacobperron:

jacobperron:

gazebo_ros/package.xml
scpeters: oh, I guess it's even used in ros2pkg (https://github.com/ros2/ros2cli/pull/94):

* https://github.com/ros2/ros2cli/blob/5c99e5f777ea3355ad8760663e3ee486872e2600/ros2pkg/ros2pkg/verb/create.py#L23-L26

scpeters (approved):
thanks for the fix and your patience with explaining it all

PR #1295, opened 2021-07-08
Added ROS parameter to toggle performance metrics in gazebo_ros

adityapande-1995:


gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -206,7 +210,12 @@ void GazeboRosInitPrivate::onPerformanceMetrics(
     msg_ros.sensors.push_back(sensor_msgs);
   }
 
-  performance_metrics_pub_->publish(msg_ros);
+  // Check if parameter was enabled
+  bool check_enable_performance_metrics ;
+  this->ros_node_->get_parameter("enable_performance_metrics", check_enable_performance_metrics);
+  if (check_enable_performance_metrics){
```
jacobperron: IIUC, we don't need to execute the rest of the logic in `onPerformanceMetrics` above. I suggest we check this parameter at the beginning of the method and return early if the parameter value is `false`.

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -176,6 +176,10 @@ void GazeboRosInit::Load(int argc, char ** argv)
     rclcpp::ParameterValue(GazeboRosInitPrivate::DEFAULT_PUBLISH_FREQUENCY));
   impl_->throttler_ = Throttler(rate_param.get<double>());
 
+  // PerformanceMetrics parameter
+  impl_->ros_node_->declare_parameter<bool>("enable_performance_metrics", true);
+
+
```
jacobperron, line 185: nitpick: remove this extra blank line (I thought a linter might complain, but perhaps not).

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -176,6 +176,10 @@ void GazeboRosInit::Load(int argc, char ** argv)
     rclcpp::ParameterValue(GazeboRosInitPrivate::DEFAULT_PUBLISH_FREQUENCY));
   impl_->throttler_ = Throttler(rate_param.get<double>());
 
+  // PerformanceMetrics parameter
+  impl_->ros_node_->declare_parameter<bool>("enable_performance_metrics", true);
```
jacobperron: I don't think the performance metrics feature is really documented anywhere. It might be nice to pass a [ParameterDescriptor](https://github.com/ros2/rclcpp/blob/0034929eef9f6626530d8207030b791e121a5415/rclcpp/include/rclcpp/node.hpp#L434) here with the `description` field set to document what this parameter does (e.g. "if set to true, performance metrics are published to the topic 'performance_metrics'").

jacobperron (commented):
:+1: 

I've left some minor comments

gazebo_ros/src/gazebo_ros_init.cpp
adityapande-1995: Done

gazebo_ros/src/gazebo_ros_init.cpp
adityapande-1995: Done

gazebo_ros/src/gazebo_ros_init.cpp
adityapande-1995: Done

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -188,25 +195,29 @@ void GazeboRosInit::Load(int argc, char ** argv)
...
-    sensor_msgs.name = sensor.name();
-
-    if (sensor.has_fps()) {
-      sensor_msgs.fps = sensor.fps();
-    } else {
-      sensor_msgs.fps = -1;
+  // Check if performance metrics parameter was enabled
+  bool check_enable_performance_metrics;
+  this->ros_node_->get_parameter("enable_performance_metrics", check_enable_performance_metrics);
+  if (check_enable_performance_metrics) {
```
jacobperron: Prefer returning early, this cuts down on the level of indentation (and is more readable IMO). E.g.

```c++
if (!check_enable_performance_metrics) {
  return;
}

Comment (adityapande-1995) [reply gazebo_ros/src/gazebo_ros_init.cpp]: Done

**Review by jacobperron — APPROVED** [review]
LGTM!

It'd be nice to sort out what's up with CI before merging though.

**Comment by jacobperron** [comment]

File: gazebo_ros/test/perf_metrics_subscriber.py [hunk 5]
```diff
@@ -0,0 +1,55 @@
...
+
+class RunTest:
+    def __init__(self):
+        self.msg_count = 0
+
+    # Echo the messages from /performance_metrics topic, let it run for a while
+    def read_msgs(self):
+        print("Reading the performance_metrics topic...")
+        os.environ['PYTHONUNBUFFERED'] = '1'
+        self.read_msgs_process = subprocess.Popen(["ros2", "topic", "echo", "/performance_metrics"], stdout=subprocess.PIPE)
```
jacobperron: Rather than depending on ros2topic and starting a subprocess, we can instead create a subscription to /performance_metrics directly in this process. This way we avoid potential bugs (or changes in output) in ros2topic effecting our test or issues with clean shutdown of this test, for example.

gazebo_ros/test/perf_metrics_subscriber.py
```diff
@@ -0,0 +1,55 @@
...
+        start = time.time()
+        output = []
+        while 1:
+            if poller.poll(0):
+                line = self.read_msgs_process.stdout.readline()
+                output.append(line)
+
+            now = time.time()
+            if now - start > 5:
+                break
```
jacobperron: I suppose it's a matter of preference, but if we make this the condition of the while-loop it makes it obvious under what circumstances the loop ends, e.g.

```py
while time.time() - start < 5:
    ...
```

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+            sigterm_timeout='20',
+            sigkill_timeout='20'
+        ),
+        launch_testing.actions.ReadyToTest()
+    ])
+
+
+class MyTestingClass(unittest.TestCase):
+    def test_routine(self):
+        rclpy.init()
```
jacobperron: I suggest doing the rclpy.init() and rclpy.shutdown() in [`setUpClass`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass) and [`tearDownClass`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDownClass) methods respectively. This should help remove a lot the if-else blocks below.

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+        node = MakeTestNode()
+        node_found = node.wait_for_gazebo_node()
+
+        if node_found:
```
jacobperron: We could more simply write:

```python
assert node.wait_for_gazebo_node(), 'Parameter could not be set'
```

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+import launch_testing
+import launch_testing.asserts
+import pytest
+import os
+
+import rclpy
+from gazebo_msgs.msg import PerformanceMetrics
+from rclpy.node import Node
+from rcl_interfaces.msg import Parameter, ParameterValue
+from ros2param.api import call_set_parameters
```
jacobperron: If we can help it, I think we should avoid depending on `ros2param` for similar reasons to depending on `ros2topic`. I'd rather just add our own implementation of calling the set_parameters service to this file (e.g. as part of the the test fixture).

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+class MyTestingClass(unittest.TestCase):
+    def test_routine(self):
+        rclpy.init()
+        node = MakeTestNode()
+        node_found = node.wait_for_gazebo_node()
+
+        if node_found:
+            response = node.set_parameter(False)
+            if response.successful:
+                print("enable_performance_metrics successfully set to false !")
```
jacobperron: Style: prefer single quotes for strings (see [style guide](https://docs.ros.org/en/rolling/Contributing/Code-Style-Language-Versions.html#python)).

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+                time.sleep(2.5)
+
+                if node.msg_count > 0:
+                    print("Received {} messages, test failed".format(node.msg_count))
+                    rclpy.shutdown()
+                    assert False
+                else:
+                    print("No messages were published")
+                    rclpy.shutdown()
+                    assert True
```
jacobperron:
```suggestion
                assert node.msg_count == 0, f'Received {node.msg_count} messages'
```

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+
+class MyTestingClass(unittest.TestCase):
+    def test_routine(self):
+        rclpy.init()
+        node = MakeTestNode()
+        node_found = node.wait_for_gazebo_node()
+
+        if node_found:
+            response = node.set_parameter(False)
+            if response.successful:
```
jacobperron: Alternatively,

```python
assert response.successful
```

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Set enable_performance_metrics parameter to false
+    def set_parameter(self, state=False):
+        parameter = Parameter()
+        parameter.name = 'enable_performance_metrics'
+
+        temp_value = ParameterValue()
+        temp_value.bool_value = state
+        temp_value.type = 1
+        parameter.value = temp_value
```
jacobperron: I think we can avoid the temporary. Also, prefer using the named constants defined in the message instead of literal numbers:
```suggestion
        parameter.value.bool_value = state
        parameter.value.type = ParameterType.PARAMETER_BOOL
```

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+                                    parameters=[parameter]).results[0]
+
+    # Start listening on messages
+    def start_subscriber(self):
+        self.subscription = self.create_subscription(
+            PerformanceMetrics,
+            'performance_metrics',
+            self.subscriber_callback,
+            10)
+        self.subscription
```
jacobperron:
```suggestion
```

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+            output='screen',
+            sigterm_timeout='20',
+            sigkill_timeout='20'
+        ),
+        launch_testing.actions.ReadyToTest()
+    ])
+
+
+class MyTestingClass(unittest.TestCase):
+    def test_routine(self):
```
jacobperron: Let's pick a less generic name for our test fixture, like `TestPerformanceMetricsParam`. Same goes for the test method (e.g. `test_disable_performance_metrics` makes it more clear what this test is about). Adding docstrings is a good habit too :slightly_smiling_face:

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def wait_for_gazebo_node(self):
+        start = time.time()
+        print("Waiting to find Gazebo Node...")
+        while time.time() - start < 5 and not self.flag_gazebo_node:
+            active_nodes = self.get_node_names()
+            if 'gazebo' in active_nodes:
+                self.flag_gazebo_node = True
+                print("Found the gazebo node !")
+                time.sleep(0.1)
```
jacobperron:
```suggestion
            self.flag_gazebo_node = 'gazebo' in active_nodes
            time.sleep(0.1)
```

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+            assert False
+
+
+class MakeTestNode(Node):
+    def __init__(self):
+        super().__init__('testing_node')
+        self.flag_gazebo_node = False
+        self.msg_count = 0
+
+    def wait_for_gazebo_node(self):
```
jacobperron: It might be nice to pass the timeout as a parameter, e.g.
```suggestion
    def wait_for_gazebo_node(self, timeout_sec=5.0):
```

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+            if 'gazebo' in active_nodes:
+                self.flag_gazebo_node = True
+                print("Found the gazebo node !")
+                time.sleep(0.1)
+
+        if not self.flag_gazebo_node:
+            print("Gazebo node was not found !")
+        return self.flag_gazebo_node
+
+    # Set enable_performance_metrics parameter to false
```
jacobperron: This comment isn't quite correct, since the value depends on the argument to this method. Consider adding a docstring :)

gazebo_ros/test/disable_perf_metrics.test.py
```diff
@@ -0,0 +1,126 @@
...
+                assert False
+        else:
+            print("Could not find Gazebo node, exiting")
+            rclpy.shutdown()
+            assert False
+
+
+class MakeTestNode(Node):
+    def __init__(self):
+        super().__init__('testing_node')
```
jacobperron: Consider a more descriptive name for this test node (e.g. `test_perf_metrics_node`). In case someone notices the node on the ROS graph they'll know roughly where it's coming from.

jacobperron (commented):
Nice, I like the direction the tests are headed. I think we can do a bit of refactoring to clean it up (see comments below). Also, for completeness we could add a second test method for setting the performance metrics parameter to `true`.

adityapande-1995:
Changed made:
* Added doc-strings, fixed incorrect comments
* Changed if-else blocks to assertions
* Added timeout parameter to wait_for_gazebo()
* Meaningful / descriptive names for node, file, classes
* Double quotes changed to single quotes, except in doc-strings
* rclpy.init() and shutdown() in ```@classmethod ```
* Shorter logic in wait_for_gazebo() method
* Test case added to enable the parameter and ensure messages are received
*  ```ParameterType.PARAMETER_BOOL``` added instead of hard coding the numerical value
* Unnecessary line ```self.subscription``` removed

TODO:
* Remove dependency on ros2params API
* Linting, uncrustify check

adityapande-1995:
set_parameter is based on [ros2param call_set_parameter function](https://github.com/ros2/ros2cli/blob/6cbe07025506233120703bc88e6a2123615e2e97/ros2param/ros2param/api/__init__.py#L189) rather than using the function itself.

adityapande-1995:
TODO:
* Reflect the changes made in documentation

gazebo_ros/test/test_disable_performance_metrics.test.py
```diff
@@ -0,0 +1,172 @@
...
+from rcl_interfaces.msg import Parameter, ParameterType
+from rcl_interfaces.srv import SetParameters
+import time
+from threading import Thread
+
+
+@pytest.mark.launch_test
+def generate_test_description():
+    """Launch a gzserver node."""
+    os.chdir(os.path.dirname(__file__))
```
jacobperron: I guess this line isn't necessary?

gazebo_ros/test/test_disable_performance_metrics.test.py
```diff
@@ -0,0 +1,172 @@
...
+class TestPerformanceMetricsParam_enable(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        rclpy.init()
+
+    @classmethod
+    def tearDownClass(cls):
+        rclpy.shutdown()
+
+    def test_parameter_enable(self):
```
jacobperron: For completeness, it would be nice to add a second test for disabling the parameter.

gazebo_ros/test/test_disable_performance_metrics.test.py
```diff
@@ -0,0 +1,172 @@
...
+            'Gazebo node not found, parameter could not be set'
+
+        # Set enable_performance_metrics parameter to True and listen for messages
+        response = node.set_parameter(state=True)
+        assert response.successful, 'Parameter could not be set to True'
+
+        node.start_subscriber()
+        time.sleep(2.5)
+        assert node.msg_count != 0, 'Did not receive any messages after\
+            setting enable_performance_metrics to True, test failed'
```
jacobperron: We can do better than blindly sleeping. We could poll with a timeout for `node.msg_count` to be greater than zero. Or even better, use a Python [`Event`](https://docs.python.org/3/library/threading.html#event-objects) object to signal from the spin thread to this thread.

jacobperron (commented):
A few more comments about the tests. Generally looks good to me :slightly_smiling_face:

gazebo_ros/test/test_disable_performance_metrics.test.py
adityapande-1995: The ``TestPerformanceMetricsParam_disable`` block disables the parameter (when it is originally enabled) and makes sure no msgs are received. After that ``TestPerformanceMetricsParam_enable`` test runs which enables the parameter (which is currently disabled) and checks that messages are received successfully.

gazebo_ros/test/test_disable_performance_metrics.test.py
adityapande-1995: Added ``Event()`` based quick return logic

gazebo_ros/test/test_disable_performance_metrics.test.py
```diff
@@ -0,0 +1,190 @@
...
+        assert response.successful, 'Parameter could not be set to True'
+
+        node.start_subscriber()
+        msgs_received_flag = node.msg_event_object.wait(timeout=5.0)
+        global msg_interval
+        assert msgs_received_flag, 'Did not receive any messages after\
+            setting enable_performance_metrics to True, test failed'
+
+
+class TestPerformanceMetricsParam_2(unittest.TestCase):
```
jacobperron: I don't think we need a separate test fixture for the second test; we can keep both tests associated with the same fixture,  ie.

```py
class TestPerformanceMetricsParam(unittest.TestCase):
    
    @classmethod
    def setUpClass(cls):
        rclpy.init()

    @classmethod
    def tearDownClass(cls):
        rclpy.shutdown()

    def test_parameter_enable(self):
        ...
    
    def test_parameter_disable(self):
        ...
```

adityapande-1995: Done. However, ```rclpy.shutdown()`` has to be called after the end of both of them to stop the spin thread, so that the next node can spin.

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

jacobperron:
I think CI is blocked by https://github.com/ros/rosdistro/pull/30344

jacobperron (approved):
Tests look good

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

adityapande-1995:
The only error in CI currently is ```what():  boost thread: trying joining itself: Resource deadlock avoided``` and it seems flaky.

PR #1298, opened 2021-07-19
Added runtime warnings, need to check hand_of_god demo

adityapande-1995:


jacobperron (commented):
Thanks for working on this!

To help reduce code duplication, I imagined we could perform this check in a common place (like [gazebo_ros::Node](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/84a088be9bfcba4f8dda29eb52265c112dd808d2/gazebo_ros/src/node.cpp)). Maybe we also want to register a callback with [add_on_set_parameters_callback](https://github.com/ros2/rclcpp/blob/0d6d9e67782bcd1de06af875459e4abb2b91ab86/rclcpp/include/rclcpp/node.hpp#L926) and emit a warning in the callback. This way users will be warned if they try to set `use_sim_time` during runtime as well as at startup.

Also, I'm not sure it makes sense to set `use_sim_time` to `true` in the event that the check fails. I think we want to warn the user simply because the value of `use_sim_time` shouldn't affect any of the default plugins.

adityapande-1995:
Current status : parameter change callback implemented in Node.cpp, if it detects change in use_sim_time, just prints a warning saying that it will be ignored. hand_of_god demo changed to use gazebo time, need to check if it works fine. All demos which have a timestamp in the message headers now use gazebo time. 

TODO : Update documentation once the PR is merged

gazebo_ros/src/node.cpp
```diff
@@ -108,6 +108,15 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
   // Parse the qos tag
   node->qos_ = gazebo_ros::QoS(sdf, name, ns, node_options);
 
+  // Generate warning on start up if use_sim_time parameter is set
+  bool check_sim_time;
+  node->get_parameter("use_sim_time", check_sim_time);
+  if (!check_sim_time) {
+    RCLCPP_WARN(
+      node->get_logger(), "Startup warning: use_sim_time will be ignored and messages "
+      "will continue to use simulation timestamps");
```
jacobperron:
```suggestion
      node->get_logger(), "Startup warning: use_sim_time parameter will be ignored by "
      "default plugins and ROS messages will continue to use simulation timestamps");
```

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -177,6 +181,27 @@ Node::SharedPtr Node::CreateWithArgs(Args && ... args)
...
+      auto result = rcl_interfaces::msg::SetParametersResult();
+      result.successful = true;
+
+      for (const auto & parameter : parameters) {
+        auto param_name = parameter.get_name();
+        if (param_name == "use_sim_time") {
+          RCLCPP_WARN(
+            node->get_logger(),
+            "use_sim_time will be ignored and messages will "
+            "continue to use simulation timestamps");
```
jacobperron: Some rewording suggestions to make it clear this warning is with respect to a ROS parameter. Also, since we're just logging a warning, this statement is **not** necessarily true for third-party plugins.

```suggestion
            "use_sim_time parameter will be ignored by default plugins "
            "and ROS messages will continue to use simulation timestamps");
```

Alternatively, we could reject changes to `use_sim_time`, then any plugins using a ROS Clock associated with a `gazebo_ros::Node` will always use simulation time provided by the `/clock` topic. This might be a bit drastic though; I think warning is good improvement.

gazebo_ros/src/node.cpp
```diff
@@ -108,6 +108,15 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
   // Parse the qos tag
   node->qos_ = gazebo_ros::QoS(sdf, name, ns, node_options);
 
+  // Generate warning on start up if use_sim_time parameter is set
```
jacobperron: It's probably better to move this logic into `CreateWithArgs`, then it will also cover the case where plugins use `Get()` (without an sdf argument).

jacobperron (commented):
Looks reasonable to me. I've left a couple minor comments.

It'd be good to get @chapulina (or someone familiar with the "hand of god" plugin) to also take a look.

adityapande-1995:

gazebo_ros/include/gazebo_ros/node.hpp
adityapande-1995: Reworded the warning, and currently accepting ```use_sim_time``` changes with a warning.

jacobperron (approved):

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

jacobperron:
I think the bots are offline. I think someone is looking into it.

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

adityapande-1995:
## TODO 
- ``build/gazebo_ros/test/test_node ``  fails when expecting different pointer addresses

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -129,6 +130,9 @@ class GAZEBO_ROS_NODE_PUBLIC Node : public rclcpp::Node
   /// Inherit constructor
   using rclcpp::Node::Node;
 
+  // A handler for the param change callback.
+  rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr param_change_callback_handler;
```
jacobperron: style nit: Use an underscore suffix like the other member variables
```suggestion
  rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr param_change_callback_handler_;
```

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -174,6 +178,36 @@ Node::SharedPtr Node::CreateWithArgs(Args && ... args)
...
+  node->get_parameter("use_sim_time", check_sim_time);
+  if (!check_sim_time) {
+    RCLCPP_WARN(
+      node->get_logger(), "Startup warning: use_sim_time parameter will be ignored "
+      "by default plugins and ROS messages will continue to use simulation timestamps");
+  }
+
+  // Parameter change callback
+  auto param_change_callback =
+    [&node](std::vector<rclcpp::Parameter> parameters) {
```
jacobperron: I'm a little worried that it's still possible for `node` to be destructed and then we try to access it asynchronously in this callback. I'm not sure if it's likely to happen (depending on how the executor handles a destructed node), but it's probably better to be safe and use a weak_ptr here instead. This way we can actually check if the node has gone out of scope or not.

adityapande-1995: Done

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

adityapande-1995:
@ros-simulation/pull-request-builder-bot retest this please

jacobperron (approved):

adityapande-1995:
There is a known flaky failure (https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1304) where the ``test_sim_time.cpp`` test fails when expecting different time stamps from ```node->now()``` but receives the same ones. This seems to be unrelated to this PR, as the test creates a ```rclcpp::Node``` instead of a gazebo_ros Node and this PR only prints out a warning.

Issue #1304, opened 2021-08-06
Flaky test failure in test_sim_time.cpp


adityapande-1995:
The test in [test_sim_time.cpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/42f47707100ee0948fd63dab1fe6c92e7d65f266/gazebo_ros/test/test_sim_time.cpp) fails sometimes when expecting different timestamps on [this line](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/42f47707100ee0948fd63dab1fe6c92e7d65f266/gazebo_ros/test/test_sim_time.cpp#L91). The code expects ```node->now()``` time to be strictly increasing, and the test fails when the timestamps are equal. 

The exact point of failure in one of the CI tests: [link](https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/106/consoleFull#45593342ed30c675-ba23-4c35-b655-f5c948f97581)
```
19: /var/lib/jenkins/workspace/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/ws/src/gazebo_ros_pkgs/gazebo_ros/test/test_sim_time.cpp:91: Failure
19: Expected: (times[i]) > (times[i - 1]), actual: 24-byte object <08-41 82-CA 00-7F 00-00 80-15 86-11 00-00 00-00 01-00 00-00 FF-7F 00-00> vs 24-byte object <08-41 82-CA 00-7F 00-00 80-15 86-11 00-00 00-00 01-00 00-00 FF-7F 00-00>
19: 294000000    294000000
19: [Msg] Waiting for master.
19: [Msg] Connected to gazebo master @ http://127.0.0.1:11345
19: [Msg] Publicized address: 172.17.0.2
19: [Msg] Loading world file [/usr/share/gazebo-11/worlds/empty.world]
19: [  FAILED  ] TestSimTime.TestClock (1908 ms)
19: [----------] 1 test from TestSimTime (1908 ms total)
19: 
19: [----------] Global test environment tear-down
19: [==========] 1 test from 1 test suite ran. (1908 ms total)
19: [  PASSED  ] 0 tests.
19: [  FAILED  ] 1 test, listed below:
19: [  FAILED  ] TestSimTime.TestClock
```

[...]

PR #1308, opened 2021-09-08
Default slip values fix for wheel slip plugin

adityapande-1995:
This PR makes sure that the default values for ``slip_compliance_unitless_lateral`` and ``slip_compliance_unitless_longitudinal`` in ``GazeboRosWheelSlip`` plugin are taken from the supplied sdf file, and not simply initialized to  ``-1``  always.

Thus, one can query the parameters using ``ros2 param get`` and restore them later if needed

* If the plugin has more than one wheel, how should the slip parameters be handled ? Currently the each plugin instance has one pair of lateral, longitudinal slip values and it is assumed that all wheels in it would have have the same values. For e.g. 
```
<plugin name="wheel_slip_rear" filename="libgazebo_ros_wheel_slip.so">
        <ros>
          <namespace>trisphere_cycle_slip1</namespace>
       </ros>

        <wheel link_name="wheel_rear_left">
          <slip_compliance_lateral>1</slip_compliance_lateral>
          <slip_compliance_longitudinal>1</slip_compliance_longitudinal>
          <wheel_normal_force>32</wheel_normal_force>
        </wheel>

        <wheel link_name="wheel_rear_right">
          <slip_compliance_lateral>1</slip_compliance_lateral>
          <slip_compliance_longitudinal>1</slip_compliance_longitudinal>
          <wheel_normal_force>32</wheel_normal_force>
        </wheel>
      </plugin>
```

This will be handled in a separate PR to upstream ``WheelSlipPlugin``

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
+++ b/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
@@ -86,6 +86,7 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr 
             this->SetSlipComplianceLateral(slip);
           } else {
             result.successful = false;
+            result.reason = "Slip compliance values cannot be negative";
           }
         } else if (param_name == "slip_compliance_unitless_longitudinal") {
           double slip = parameter.as_double();
@@ -96,6 +97,7 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr 
             this->SetSlipComplianceLongitudinal(slip);
           } else {
             result.successful = false;
+            result.reason = "Slip compliance values cannot be negative";
           }
         }
       }
--- a/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
+++ b/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
@@ -50,6 +50,20 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr 
   // Initialize the WheelSlipPlugin first so its values are preferred unless the ros
   // parameters are overridden by a launch file.
   WheelSlipPlugin::Load(_model, _sdf);
+  double slip_lateral_default = -1.0;
+  double slip_longitudinal_default = -1.0;
+
+  std::cout << "Plugin name: " << _sdf->Get<std::string>("name") << std::endl;
+
+  if (_sdf->HasElement("wheel")) {
+    auto wheel_element = _sdf->GetElement("wheel");
+    while (wheel_element) {
+      // auto wheel_name = wheel_element->Get<std::string>("link_name");
+      slip_lateral_default = wheel_element->Get<double>("slip_compliance_lateral");
+      slip_longitudinal_default = wheel_element->Get<double>("slip_compliance_longitudinal");
+      wheel_element = wheel_element->GetNextElement("wheel");
+    }
+  }
 
   // Initialize ROS node
   impl_->ros_node_ = gazebo_ros::Node::Get(_sdf);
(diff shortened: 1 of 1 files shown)


scpeters:

scpeters (approved):

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -50,6 +50,23 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
   // Initialize the WheelSlipPlugin first so its values are preferred unless the ros
   // parameters are overridden by a launch file.
   WheelSlipPlugin::Load(_model, _sdf);
+  double slip_lateral_default = -1.0;
+  double slip_longitudinal_default = -1.0;
```
jacobperron: IIUC, if these are not set in SDF, then the parent plugin defaults to zero values.
```suggestion
  double slip_lateral_default = 0.0;
  double slip_longitudinal_default = 0.0;
```

jacobperron (commented):
Now that we are reporting the actual wheel slip values, I think we should reject parameter changes for negative values. I.e. if the conditions on L82 or L90 are false, then set `result.successful = false` and return.

For posterity, it might make sense to add getters for the wheel slip values in the parent plugin. Then we can avoid re-parsing the SDF here for instance. I think it's low-priority, but we could go ahead and add this to Gazebo (classic) and Ignition Gazebo. @scpeters What do you think?

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -67,6 +84,8 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
               impl_->ros_node_->get_logger(),
               "New lateral slip compliance: %.3e", slip);
             this->SetSlipComplianceLateral(slip);
+          } else {
+            result.successful = false;
```
jacobperron, line 88: We can also set the [reason](https://github.com/ros2/rcl_interfaces/blob/9a44202bbaf76e19fde21da0e90d8afbe89bd263/rcl_interfaces/msg/SetParametersResult.msg#L7) to something informative.

jacobperron (approved):
LGTM

The test failing is a known issue (https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1304).

scpeters:
> Now that we are reporting the actual wheel slip values, I think we should reject parameter changes for negative values. I.e. if the conditions on L82 or L90 are false, then set `result.successful = false` and return.

+1

> For posterity, it might make sense to add getters for the wheel slip values in the parent plugin. Then we can avoid re-parsing the SDF here for instance. I think it's low-priority, but we could go ahead and add this to Gazebo (classic) and Ignition Gazebo. @scpeters What do you think?

makes sense to me. @adityapande-1995 and I discussed this yesterday, in fact. I can advise on making these changes

adityapande-1995:
Changed default slip values to 0.0, Added rejection message if user tries to set parameter to < 0.0. Should I merge this one ?

scpeters:
> Changed default slip values to 0.0, Added rejection message if user tries to set parameter to < 0.0. Should I merge this one ?

let's wait for CI to finish

scpeters:
this test failure is an upstream gazebo issue: https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/118/testReport/junit/(root)/gazebo_plugins/test_gazebo_ros_planar_move_gtest_missing_result/

https://github.com/osrf/gazebo/issues/2019

adityapande-1995:
Not able to reproduce this one locally : https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/118/testReport/junit/(root)/gazebo_plugins/test_gazebo_ros_harness_gtest_missing_result/

scpeters:
> Not able to reproduce this one locally : https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-focal-amd64/118/testReport/junit/(root)/gazebo_plugins/test_gazebo_ros_harness_gtest_missing_result/

yeah, there's nothing useful in that console output. we'd need a backtrace to see how it failed. I think this is fine to merge now

PR #1309, opened 2021-09-08
gazebo_ros_camera: set LOD bias parameter via xml

scpeters:
Level of Detail (LOD) is a feature of OGRE that can improve performance when rendering large scenes. Parameters can be set for an entire scene using gazebo's [HeightmapLODPlugin](https://github.com/osrf/gazebo/blob/gazebo11/plugins/HeightmapLODPlugin.cc) for example. In some cases, it may be useful for the LOD behavior to vary for different cameras  in a scene. For example, you may wish to adjust LOD to reduce the computational demands of the user camera in `gzclient`, while preferring greater detail in a triggered camera sensor. The LOD bias parameter allows tuning individual cameras in this way. This pull request adds a `<lod_bias>` parameter to `gazebo_ros_camera` that will be passed to any attached cameras. Special logic for the `WideAngleCamera` was added to ensure that this parameter is applied to all its env cameras.

cc @mogumbo @ kjeppesen1 @williamlewww

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera.cpp
@@ -15,7 +15,9 @@
 #include <sdf/Element.hh>
 #include <gazebo/common/Events.hh>
 #include <gazebo/rendering/Distortion.hh>
+#include <gazebo/rendering/WideAngleCamera.hh>
 #include <gazebo/sensors/SensorTypes.hh>
+#include <gazebo/sensors/WideAngleCameraSensor.hh>
 #ifdef IGN_PROFILER_ENABLE
 #include <ignition/common/Profiler.hh>
 #endif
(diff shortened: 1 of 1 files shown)


jacobperron (approved):
I haven't tested this, but the code LGTM.

The LOD SDF logic seems like something that would live in the upstream CameraSensor (and WideAngleCameraSensor). Is there a reason to add this to gazebo_ros instead of Gazebo directly?

mogumbo:
> I haven't tested this, but the code LGTM.
> 
> The LOD SDF logic seems like something that would live in the upstream CameraSensor (and WideAngleCameraSensor). Is there a reason to add this to gazebo_ros instead of Gazebo directly?

I tested this and it appears to work as expected. If I'm not mistaken, this doesn't give us a way to set LOD bias in gzclient, and that would be the main advantage of adding this feature to Gazebo instead of gazebo_ros_pkgs.

scpeters:
@iche033 just suggested a better approach, so I will close this

PR #1312, opened 2021-09-17
Adds slip values for individual wheels

adityapande-1995:
This PR aims to add a functionality to add slip values to individual wheels
Upstream PR: https://github.com/osrf/gazebo/pull/3082

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
@@ -57,8 +57,8 @@ class GazeboRosWheelSlipPrivate;
 /// Slip compliance values can be declared in 3 ways:
 ///  1. SDF parameters, e.g. :
 ///       <wheel link_name="wheel_front">
-///         <slip_compliance_lateral>1</slip_compliance_lateral>
-///         <slip_compliance_longitudinal>2</slip_compliance_longitudinal>
+///         <slip_compliance_lateral>0.1</slip_compliance_lateral>
+///         <slip_compliance_longitudinal>0.2</slip_compliance_longitudinal>
 ///         <wheel_normal_force>77</wheel_normal_force>
 ///       </wheel>
 ///
@@ -67,9 +67,9 @@ class GazeboRosWheelSlipPrivate;
 ///
 ///  2. ROS parameters for each wheel, e.g :
 ///       <ros>
-///         <parameter name="slip_compliance_unitless_lateral/wheel_front" type="double">10.0
+///         <parameter name="slip_compliance_unitless_lateral/wheel_front" type="double">0.1
 ///           </parameter>
-///         <parameter name="slip_compliance_unitless_longitudinal/wheel_front" type="double">11.0
+///         <parameter name="slip_compliance_unitless_longitudinal/wheel_front" type="double">0.2
 ///           </parameter>
 ///       </ros>
 ///
@@ -78,8 +78,8 @@ class GazeboRosWheelSlipPrivate;
 ///
 ///  3. ROS parameters for all wheels, e.g :
 ///      <ros>
-///         <parameter name="slip_compliance_unitless_lateral" type="double">10.0</parameter>
-///         <parameter name="slip_compliance_unitless_longitudinal" type="double">11.0</parameter>
+///         <parameter name="slip_compliance_unitless_lateral" type="double">0.1</parameter>
+///         <parameter name="slip_compliance_unitless_longitudinal" type="double">0.2</parameter>
 ///      </ros>
 ///
 ///     If not specified, these will default to the last values set in SDF tags.
@@ -98,10 +98,10 @@ class GazeboRosWheelSlipPrivate;
     <plugin name="projector" filename="libgazebo_ros_wheel_speed.so">
       <ros>
         <namespace>wheel_slip_front</namespace>
-        <parameter name="slip_compliance_unitless_lateral/wheel_front_left" type="double">10.0</parameter>
-        <parameter name="slip_compliance_unitless_longitudinal/wheel_front_left" type="double">11.0</parameter>
-        <parameter name="slip_compliance_unitless_lateral" type="double">10.0</parameter>
-        <parameter name="slip_compliance_unitless_longitudinal" type="double">11.0</parameter>
+        <parameter name="slip_compliance_unitless_lateral/wheel_front_left" type="double">0.1</parameter>
+        <parameter name="slip_compliance_unitless_longitudinal/wheel_front_left" type="double">0.2</parameter>
+        <parameter name="slip_compliance_unitless_lateral" type="double">0.3</parameter>
+        <parameter name="slip_compliance_unitless_longitudinal" type="double">0.4</parameter>
       </ros>
       <wheel link_name="wheel_front_left">
         <slip_compliance_lateral>0</slip_compliance_lateral>
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
@@ -53,41 +53,45 @@ class GazeboRosWheelSlipPrivate;
 ///      - Description: Unitless slip compliance (slip / friction) in the
 ///           longitudinal direction. This value is applied to the wheel named 'wheel_name'
 ///           declared in the WheelSlipPlugin.
-///  
+///
 /// Precedence order and deafult values:
 /// -----------------------------------
 /// Slip compliance values can be declared in 3 ways:
-///  1. SDF parameters, for e.g. : 
+///  1. SDF parameters, for e.g. :
 ///       <wheel link_name="wheel_front">
 ///         <slip_compliance_lateral>1</slip_compliance_lateral>
 ///         <slip_compliance_longitudinal>2</slip_compliance_longitudinal>
 ///         <wheel_normal_force>77</wheel_normal_force>
 ///       </wheel>
-///      
-///      These if not specified, will default to 0.0. Any negative values will ignored and set to 0.0
-///      
+///
+///      These if not specified, will default to 0.0. Any negative values will ignored and
+//       set to 0.0
+///
 ///  2. ROS parameters for each wheel, for e.g :
 ///       <ros>
-///         <parameter name="slip_compliance_unitless_lateral/wheel_front" type="double">10.0</parameter>
-///         <parameter name="slip_compliance_unitless_longitudinal/wheel_front" type="double">11.0</parameter>
+///         <parameter name="slip_compliance_unitless_lateral/wheel_front" type="double">10.0
+///           </parameter>
+///         <parameter name="slip_compliance_unitless_longitudinal/wheel_front" type="double">11.0
+///           </parameter>
 ///       </ros>
 ///
-///      These if not specified, default to SDF parameters set for each wheel. If specified, they will override the 
-///      ones set as SDF parameters.
-///     
-///  3. ROS parameters for all wheels, for e.g :  
+///      These if not specified, default to SDF parameters set for each wheel. If specified, they
+///      will override the ones set as SDF parameters.
+///
+///  3. ROS parameters for all wheels, for e.g :
 ///      <ros>
 ///         <parameter name="slip_compliance_unitless_lateral" type="double">10.0</parameter>
 ///         <parameter name="slip_compliance_unitless_longitudinal" type="double">11.0</parameter>
 ///      </ros>
 ///
-///     These if not specified will default to the last ones set in the SDF tags. In this case, they won't override anything.
+///     These if not specified will default to the last ones set in the SDF tags. In this case,
+///     they won't override anything.
 ///     If these are specified, they override both SDF and the ROS parameters for each wheel.
 ///
 ///   Precedence order :
 ///   ROS Params for all wheels > ROS params for individual wheels > SDF parameters
 ///   Check out the test cases for more information and expected behaviour.
-///  
+///
 /// See the WheelSlipPlugin documentation at the following location for more details:
 /// http://osrf-distributions.s3.amazonaws.com/gazebo/api/11.0.0/classgazebo_1_1WheelSlipPlugin.html#details
 /**
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
@@ -24,12 +24,10 @@ namespace gazebo_plugins
 {
 class GazeboRosWheelSlipPrivate;
 
-/// Description:
-/// ------------
 /// A plugin for adjusting wheel slip parameters in gazebo.
 /// The plugin can set separate longitudinal and lateral wheel slip compliance
-/// parameters for separate wheel links. THe following ROS parameters can be specified in
-/// the world file (These are optional, you may not may not declare them):
+/// parameters for separate wheel links.
+/// The following ROS parameters can optionally be specified in the world file:
 ///  1. slip_compliance_unitless_lateral
 ///      - Type: double
 ///      - Description: Unitless slip compliance (slip / friction) in the
@@ -46,7 +44,7 @@ class GazeboRosWheelSlipPrivate;
 ///      - Type: double
 ///      - Description: Unitless slip compliance (slip / friction) in the
 ///           lateral direction. This value is applied to the wheel named 'wheel_name'
-///           declaredin in the WheelSlipPlugin.
+///           declared in in the WheelSlipPlugin.
 ///
 ///  4. slip_compliance_unitless_longitudinal/wheel_name
 ///      - Type: double
@@ -54,20 +52,20 @@ class GazeboRosWheelSlipPrivate;
 ///           longitudinal direction. This value is applied to the wheel named 'wheel_name'
 ///           declared in the WheelSlipPlugin.
 ///
-/// Precedence order and deafult values:
-/// -----------------------------------
+/// Precedence order and default values:
+/// ------------------------------------
 /// Slip compliance values can be declared in 3 ways:
-///  1. SDF parameters, for e.g. :
+///  1. SDF parameters, e.g. :
 ///       <wheel link_name="wheel_front">
 ///         <slip_compliance_lateral>1</slip_compliance_lateral>
 ///         <slip_compliance_longitudinal>2</slip_compliance_longitudinal>
 ///         <wheel_normal_force>77</wheel_normal_force>
 ///       </wheel>
 ///
-///      These if not specified, will default to 0.0. Any negative values will ignored and
-//       set to 0.0
+///      If not specified, these will default to 0.0.
+///      Any negative values will ignored and set to 0.0.
 ///
-///  2. ROS parameters for each wheel, for e.g :
+///  2. ROS parameters for each wheel, e.g :
 ///       <ros>
 ///         <parameter name="slip_compliance_unitless_lateral/wheel_front" type="double">10.0
 ///           </parameter>
@@ -75,17 +73,16 @@ class GazeboRosWheelSlipPrivate;
 ///           </parameter>
 ///       </ros>
 ///
-///      These if not specified, default to SDF parameters set for each wheel. If specified, they
-///      will override the ones set as SDF parameters.
+///      If not specified, these default to SDF parameters set for each wheel.
+///      If specified, they will override the values set as SDF parameters.
 ///
-///  3. ROS parameters for all wheels, for e.g :
+///  3. ROS parameters for all wheels, e.g :
 ///      <ros>
 ///         <parameter name="slip_compliance_unitless_lateral" type="double">10.0</parameter>
 ///         <parameter name="slip_compliance_unitless_longitudinal" type="double">11.0</parameter>
 ///      </ros>
 ///
-///     These if not specified will default to the last ones set in the SDF tags. In this case,
-///     they won't override anything.
+///     If not specified, these will default to the last values set in SDF tags.
 ///     If these are specified, they override both SDF and the ROS parameters for each wheel.
 ///
 ///   Precedence order :
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
@@ -24,9 +24,12 @@ namespace gazebo_plugins
 {
 class GazeboRosWheelSlipPrivate;
 
+/// Description:
+/// ------------
 /// A plugin for adjusting wheel slip parameters in gazebo.
 /// The plugin can set separate longitudinal and lateral wheel slip compliance
-/// parameters for separate wheel links.
+/// parameters for separate wheel links. THe following ROS parameters can be specified in
+/// the world file (These are optional, you may not may not declare them):
 ///  1. slip_compliance_unitless_lateral
 ///      - Type: double
 ///      - Description: Unitless slip compliance (slip / friction) in the
@@ -38,6 +41,53 @@ class GazeboRosWheelSlipPrivate;
 ///      - Description: Unitless slip compliance (slip / friction) in the
 ///           longitudinal direction. This value is applied to all wheels declared
 ///           in the WheelSlipPlugin.
+///
+///  3. slip_compliance_unitless_lateral/wheel_name
+///      - Type: double
+///      - Description: Unitless slip compliance (slip / friction) in the
+///           lateral direction. This value is applied to the wheel named 'wheel_name'
+///           declaredin in the WheelSlipPlugin.
+///
+///  4. slip_compliance_unitless_longitudinal/wheel_name
+///      - Type: double
+///      - Description: Unitless slip compliance (slip / friction) in the
+///           longitudinal direction. This value is applied to the wheel named 'wheel_name'
+///           declared in the WheelSlipPlugin.
+///  
+/// Precedence order and deafult values:
+/// -----------------------------------
+/// Slip compliance values can be declared in 3 ways:
+///  1. SDF parameters, for e.g. : 
+///       <wheel link_name="wheel_front">
+///         <slip_compliance_lateral>1</slip_compliance_lateral>
+///         <slip_compliance_longitudinal>2</slip_compliance_longitudinal>
+///         <wheel_normal_force>77</wheel_normal_force>
+///       </wheel>
+///      
+///      These if not specified, will default to 0.0. Any negative values will ignored and set to 0.0
+///      
+///  2. ROS parameters for each wheel, for e.g :
+///       <ros>
+///         <parameter name="slip_compliance_unitless_lateral/wheel_front" type="double">10.0</parameter>
+///         <parameter name="slip_compliance_unitless_longitudinal/wheel_front" type="double">11.0</parameter>
+///       </ros>
+///
+///      These if not specified, default to SDF parameters set for each wheel. If specified, they will override the 
+///      ones set as SDF parameters.
+///     
+///  3. ROS parameters for all wheels, for e.g :  
+///      <ros>
+///         <parameter name="slip_compliance_unitless_lateral" type="double">10.0</parameter>
+///         <parameter name="slip_compliance_unitless_longitudinal" type="double">11.0</parameter>
+///      </ros>
+///
+///     These if not specified will default to the last ones set in the SDF tags. In this case, they won't override anything.
+///     If these are specified, they override both SDF and the ROS parameters for each wheel.
+///
+///   Precedence order :
+///   ROS Params for all wheels > ROS params for individual wheels > SDF parameters
+///   Check out the test cases for more information and expected behaviour.
+///  
 /// See the WheelSlipPlugin documentation at the following location for more details:
 /// http://osrf-distributions.s3.amazonaws.com/gazebo/api/11.0.0/classgazebo_1_1WheelSlipPlugin.html#details
 /**
@@ -47,6 +97,10 @@ class GazeboRosWheelSlipPrivate;
     <plugin name="projector" filename="libgazebo_ros_wheel_speed.so">
       <ros>
         <namespace>wheel_slip_front</namespace>
+        <parameter name="slip_compliance_unitless_lateral/wheel_front_left" type="double">10.0</parameter>
+        <parameter name="slip_compliance_unitless_longitudinal/wheel_front_left" type="double">11.0</parameter>
+        <parameter name="slip_compliance_unitless_lateral" type="double">10.0</parameter>
+        <parameter name="slip_compliance_unitless_longitudinal" type="double">11.0</parameter>
       </ros>
       <wheel link_name="wheel_front_left">
         <slip_compliance_lateral>0</slip_compliance_lateral>
--- a/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
+++ b/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
@@ -241,6 +241,29 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr 
       wheel_parameter.second
     );
   }
+
+  // If the world file contains global wheelslip ros parameters, these should be set now
+  if (_sdf->HasElement("ros")) {
+    auto ros_element = _sdf->GetElement("ros");
+    if (ros_element->HasElement("parameter")) {
+      // Iterate over parameter tags
+      for (auto param_element = ros_element->GetElement("parameter"); param_element;
+        param_element = param_element->GetNextElement("parameter"))
+      {
+        auto param_name = param_element->Get<std::string>("name");
+        if (param_name == "slip_compliance_unitless_lateral" ||
+          param_name == "slip_compliance_unitless_longitudinal")
+        {
+          // Global wheelslip ros parameters were declared
+          double value = param_element->Get<double>();
+          this->impl_->ros_node_->set_parameter(rclcpp::Parameter(param_name, value));
+          RCLCPP_INFO(
+            this->impl_->ros_node_->get_logger(),
+            "Global wheelslip parameter found in world file, overriding sdf parameters..");
+        }
+      }
+    }
+  }
 }
 
 GZ_REGISTER_MODEL_PLUGIN(GazeboRosWheelSlip)
--- a/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
+++ b/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
@@ -20,6 +20,8 @@
 #include <gazebo_ros/node.hpp>
 #include <std_msgs/msg/bool.hpp>
 
+#include <algorithm>
+#include <map>
 #include <memory>
 #include <string>
 #include <vector>
@@ -34,6 +36,11 @@ public:
 
   /// Handle to parameters callback
   rclcpp::Node::OnSetParametersCallbackHandle::SharedPtr on_set_parameters_callback_handle_;
+
+  // Container to hold default values of slip parameters
+  std::map <std::string, double> mapSlipLateralDefault;
+  std::map <std::string, double> mapSlipLongitudinalDefault;
+
 };
 
 GazeboRosWheelSlip::GazeboRosWheelSlip()
@@ -50,21 +57,20 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr 
   // Initialize the WheelSlipPlugin first so its values are preferred unless the ros
   // parameters are overridden by a launch file.
   WheelSlipPlugin::Load(_model, _sdf);
-  double slip_lateral_default = 0.0;
-  double slip_longitudinal_default = 0.0;
 
+  // Read slip values from the sdf file and set them as default
   if (_sdf->HasElement("wheel")) {
-    auto wheel_element = _sdf->GetElement("wheel");
-    while (wheel_element) {
+    for (auto wheel_element = _sdf->GetElement("wheel"); wheel_element;
+        wheel_element = wheel_element->GetNextElement("wheel"))
+    {
+      auto wheel_name = wheel_element->Get<std::string>("link_name");
       double slip_lateral = wheel_element->Get<double>("slip_compliance_lateral");
-      if (slip_lateral >= 0.) {
-        slip_lateral_default = slip_lateral;
-      }
+      double default_slip_lateral = std::max(slip_lateral, 0.);
+      this->impl_->mapSlipLateralDefault[wheel_name] = default_slip_lateral;
+
       double slip_longitudinal = wheel_element->Get<double>("slip_compliance_longitudinal");
-      if (slip_longitudinal >= 0.) {
-        slip_longitudinal_default = slip_longitudinal;
-      }
-      wheel_element = wheel_element->GetNextElement("wheel");
+      double default_slip_longitudinal = std::max(slip_longitudinal, 0.);
+      this->impl_->mapSlipLongitudinalDefault[wheel_name] = default_slip_longitudinal;
     }
   }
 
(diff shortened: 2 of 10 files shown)


gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -75,31 +81,44 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+              RCLCPP_INFO(
+                impl_->ros_node_->get_logger(),
+                "New lateral slip compliance: %.3e", slip);
+                this->SetSlipComplianceLateral(wheel_name, slip);
+              } else {
+                result.successful = false;
+                result.reason = "Slip compliance values cannot be negative";
+              }
+            }
         }
```
jacobperron: nit: indentation looks inconsistent here.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -75,31 +81,44 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+              RCLCPP_INFO(
+                impl_->ros_node_->get_logger(),
+                "New longitudinal slip compliance : %.3e", slip);
+                this->SetSlipComplianceLongitudinal(wheel_name, slip);
+              } else {
+                result.successful = false;
+                result.reason = "Slip compliance values cannot be negative";
+              }
+            }
+        }
```
jacobperron: nit: inconsistent indentation

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -110,10 +129,21 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
   // Set negative values by default, which are ignored by the callback.
   // This approach allows values specified in a launch file to override the SDF/URDF values.
-  impl_->ros_node_->declare_parameter("slip_compliance_unitless_lateral", slip_lateral_default);
-  impl_->ros_node_->declare_parameter(
-    "slip_compliance_unitless_longitudinal",
-    slip_longitudinal_default);
+  for (auto &wheel_parameter : impl_->mapSlipLateralDefault)
+  {
+    impl_->ros_node_->declare_parameter(
+        "slip_compliance_unitless_lateral/" + wheel_parameter.first,
```
jacobperron: My only concern with this PR is that it creates a hard API break wrt the names of the ROS parameters. It would be nice if users didn't have to update their code, and could continue relying on the old behavior. Maybe we can keep the original parameters for updating slip values for all the wheels in addition to the new, per-wheel, parameters? I'm open to other ideas.

adityapande-1995: Yup that makes sense. I can keep the original parameters as well which apply to all the wheels.

adityapande-1995:
## Issue to be addressed: 
If we have the original ``slip_compliance_unitless_lateral`` as well as the new set of parameters ``slip_compliance_lateral/wheel_*``, this can lead to some confusing situations.

 * Suppose the slip values are set initially using  ``slip_compliance_lateral/wheel_*`` and then later updated for all wheels using ``slip_compliance_unitless_lateral``. If the user queries the slip parameter values using ``ros2 param dump`` the ``slip_compliance_lateral/wheel_*`` parameters would still show the original values, and  ``slip_compliance_unitless_lateral`` would show the new ones, so there is no way to know what the current parameters are. The values will be updated in the upstream Gazebo plugin, but querying them can lead to confusion.
* We cannot have a solution where updating the global parameter to set wheel slip values triggers the parameters for individual wheels as updating other parameters in a callback is not allowed by ``rclcpp``

After discussing with @jacobperron , the current solution can be to have a parameter which would make the node generate only global or only individual slip parameters when the node is created, thus avoiding the situation mentioned above altogether. I'm open to other suggestions though.

adityapande-1995:
This PR is put on hold until ``rclcpp`` supports setting other parameters in a validation callback.

adityapande-1995:
Current implementation uses ``ParameterEventHandler`` to trigger individual wheel slip parameters when the global ``set_slip_compliance`` values are set.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -50,21 +63,22 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
-    while (wheel_element) {
+    for (auto wheel_element = _sdf->GetElement("wheel"); wheel_element;
+      wheel_element = wheel_element->GetNextElement("wheel"))
+    {
+      auto wheel_name = wheel_element->Get<std::string>("link_name");
       double slip_lateral = wheel_element->Get<double>("slip_compliance_lateral");
-      if (slip_lateral >= 0.) {
-        slip_lateral_default = slip_lateral;
-      }
+      double default_slip_lateral = std::max(slip_lateral, 0.);
```
jacobperron: Maybe we should log a warning if a user provides a negative value; noting that we're clamping to zero.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -34,6 +36,17 @@ class GazeboRosWheelSlipPrivate
 
   /// Handle to parameters callback
   rclcpp::Node::OnSetParametersCallbackHandle::SharedPtr on_set_parameters_callback_handle_;
+
+  // Containers to hold default values of slip parameters
+  std::map<std::string, double> mapSlipLateralDefault;
```
jacobperron: nitpick: use snake_case for consistency

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -34,6 +36,17 @@ class GazeboRosWheelSlipPrivate
...
+  double default_slip_lateral;
+  std::map<std::string, double> mapSlipLongitudinalDefault;
```
jacobperron: Use snake_case

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -34,6 +36,17 @@ class GazeboRosWheelSlipPrivate
...
+  double default_slip_longitudinal;
+
+  // Event handler to set slip compliance values for individual wheels based
+  std::shared_ptr<rclcpp::ParameterEventHandler> parameter_subscriber;
+  rclcpp::ParameterCallbackHandle::SharedPtr global_parameter_callback_handle_lateral;
+  rclcpp::ParameterCallbackHandle::SharedPtr global_parameter_callback_handle_longitudinal;
```
jacobperron: Add an underscore suffix to all member variable names for style consistency.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
jacobperron: nitpick: prefer `std::unordered_map`

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -34,6 +36,17 @@ class GazeboRosWheelSlipPrivate
...
   rclcpp::Node::OnSetParametersCallbackHandle::SharedPtr on_set_parameters_callback_handle_;
+
+  // Containers to hold default values of slip parameters
+  std::map<std::string, double> mapSlipLateralDefault;
+  double default_slip_lateral;
+  std::map<std::string, double> mapSlipLongitudinalDefault;
+  double default_slip_longitudinal;
+
+  // Event handler to set slip compliance values for individual wheels based
+  std::shared_ptr<rclcpp::ParameterEventHandler> parameter_subscriber;
```
jacobperron: nitpick: For clarity, I would name the variable to reflect the type, e.g. `parameter_event_handler_`

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -75,25 +89,63 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+          }
+        }
+
         if (param_name == "slip_compliance_unitless_lateral") {
           double slip = parameter.as_double();
           if (slip >= 0.) {
             RCLCPP_INFO(
               impl_->ros_node_->get_logger(),
-              "New lateral slip compliance: %.3e", slip);
+              "New lateral slip compliance for all wheels: %.3e", slip);
```
jacobperron: I guess this log statement is redundant with the RCLCPP_INFO in the ParameterEventHandler callback. Maybe we can remove one of them.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -75,25 +89,63 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
-        } else if (param_name == "slip_compliance_unitless_longitudinal") {
+        }
+
+        if (param_name == "slip_compliance_unitless_longitudinal") {
           double slip = parameter.as_double();
           if (slip >= 0.) {
             RCLCPP_INFO(
               impl_->ros_node_->get_logger(),
-              "New longitudinal slip compliance: %.3e", slip);
+              "New longitudinal slip compliance for all wheels: %.3e", slip);
```
jacobperron: Same here. Maybe remove this log or the other one in the ParameterEventHandler callback.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -75,25 +89,63 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+        if (param_name.find("slip_compliance_unitless_lateral/") != std::string::npos) {
+          auto wheel_name = param_name.substr(param_name.find("/") + 1);
+          if (impl_->mapSlipLateralDefault.count(wheel_name) == 1) {
+            // The parameter has a valid name
+            double slip = parameter.as_double();
+            if (slip >= 0.) {
+              RCLCPP_INFO(
+                impl_->ros_node_->get_logger(),
+                "New lateral slip compliance for %s: %.3e", wheel_name.c_str(), slip);
+              this->SetSlipComplianceLateral(wheel_name, slip);
```
jacobperron, line 184: Since this parameters callback is meant for validation, I suggest we do the actual setting of slip compliance values in the ParameterEventHandler callback instead. I.e. let's avoid side-effects inside this callback. Consider using [this method](https://github.com/ros2/rclcpp/blob/d5ec2580805c3abd8b6fecaf3d09040e00e9f15e/rclcpp/include/rclcpp/parameter_event_handler.hpp#L192-L194) to subscribe to all parameter events and filter inside the callback like we are doing here.

adityapande-1995: Moved all actions to ``ParameterEventHandler`` callback

adityapande-1995:
**TODO** : Write a test case :heavy_check_mark: 
Ready for next round of review

adityapande-1995:
CI will probably need a release of ``gazebo11``

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,116 @@
...
+    {"slip_compliance_unitless_lateral/wheel_rear_right", 5},
+    {"slip_compliance_unitless_longitudinal", 6},
+    {"slip_compliance_unitless_longitudinal/wheel_rear_left", 4},
+    {"slip_compliance_unitless_longitudinal/wheel_rear_right", 6}};
+
+  int counter = 0;
+  while (!parameters_client_rear->wait_for_service(1s)) {
+    if (!rclcpp::ok()) {
+      rclcpp::shutdown();
+    }
```
jacobperron: I'm confused by this logic. Maybe you meant to call `rclcpp::init()` instead? If this is necessary I would suggest doing it in the test fixture `setUpTestCase` method.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,116 @@
...
+  // Set global slip compliance parameters, which should override the ones for wheels
+  std::vector<rclcpp::Parameter> change_params_global = {
+    rclcpp::Parameter("slip_compliance_unitless_lateral", 10.0),
+    rclcpp::Parameter("slip_compliance_unitless_longitudinal", 11.0)
+  };
+  result = parameters_client_rear->set_parameters(change_params_global);
+  ASSERT_TRUE(result[0].successful);
+  ASSERT_TRUE(result[1].successful);
+
+  std::this_thread::sleep_for(2s);
```
jacobperron, line 250: I presume this sleep is to give the plugin time to update its parameters since we're setting individual wheel slip parameters _after_ successfully setting the global parameters. It's unfortunate that we have to do this, but I don't see a way around it. It would be good to add a comment here explaining why we need to sleep.

gazebo_plugins/test/worlds/gazebo_ros_wheelslip.world
```diff
@@ -0,0 +1,80 @@
...
+
+    <include>
+      <uri>model://trisphere_cycle</uri>
+      <name>trisphere_cycle_slip0</name>
+      <pose>0 0 0  0 0 0</pose>
+      <plugin name="wheel_slip_front" filename="libgazebo_ros_wheel_slip.so">
+
+        <ros>
+          <namespace>trisphere_cycle_slip0</namespace>
+          <parameter name="slip_compliance_parameters_global" value="1" type="bool"/>
```
jacobperron: This doesn't appear to be a ROS parameter offered by the wheel slip plugin. Though, I'd be interested to know what happens if we set one of the global parameters here (ie. if it overrides the individual parameters or is ignored).

jacobperron (commented):
Tests look good! A few comments below.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
adityapande-1995: Yup, if I remove the sleep, the test fails as the individual wheels parameters take some time to update. Couldn't think of a way around it.

adityapande-1995:
Issues / test cases to tackle : 
* What happens if we specify slip compliance parameters in sdf file both as ros parameters and sdf tags ?
* What if some wheel element doesn't have the slip compliance tags ?

adityapande-1995:
## Current behaviour and edge cases: 
### Case 1 : No ROS params in world file, wheel params not specified in world file

World file : 
```
<plugin name="wheel_slip_front" filename="libgazebo_ros_wheel_slip.so">
        <ros>
          <namespace>trisphere_cycle_slip0</namespace>
        </ros>
        <wheel link_name="wheel_front">
          <wheel_normal_force>77</wheel_normal_force>
        </wheel>
      </plugin>
```
Result : (sdf parameters default to zero)
```
    slip_compliance_unitless_lateral: 0.0
    slip_compliance_unitless_lateral/wheel_front: 0.0
    slip_compliance_unitless_longitudinal: 0.0
    slip_compliance_unitless_longitudinal/wheel_front: 0.0

```
---
###  Case 2 : Global ROS params specified in world file + individual wheels params as sdf params

World file : 
```
<plugin name="wheel_slip_front" filename="libgazebo_ros_wheel_slip.so">
        <ros>
          <namespace>trisphere_cycle_slip0</namespace>
          <parameter name="slip_compliance_unitless_lateral" value="10.0" type="double"/>
          <parameter name="slip_compliance_unitless_longitudinal" value="11.0" type="double"/>
        </ros>
        <wheel link_name="wheel_front">
          <slip_compliance_lateral>1</slip_compliance_lateral>
          <slip_compliance_longitudinal>2</slip_compliance_longitudinal>
          <wheel_normal_force>77</wheel_normal_force>
        </wheel>
      </plugin>
```
Result : 
```
    slip_compliance_unitless_lateral: 0.0
    slip_compliance_unitless_lateral/wheel_front: 1.0
    slip_compliance_unitless_longitudinal: 0.0
    slip_compliance_unitless_longitudinal/wheel_front: 2.0
```
---
###  Case 3 : Global ROS params specified in world file + NO individual wheels params as sdf params

World file: 
```
<plugin name="wheel_slip_front" filename="libgazebo_ros_wheel_slip.so">
        <ros>
          <namespace>trisphere_cycle_slip0</namespace>
          <parameter name="slip_compliance_unitless_lateral" value="10.0" type="double"/>
          <parameter name="slip_compliance_unitless_longitudinal" value="11.0" type="double"/>
        </ros>
        <wheel link_name="wheel_front">
          <wheel_normal_force>77</wheel_normal_force>
        </wheel>
      </plugin>
```
Result : 
```
    slip_compliance_unitless_lateral: 0.0
    slip_compliance_unitless_lateral/wheel_front: 0.0
    slip_compliance_unitless_longitudinal: 0.0
    slip_compliance_unitless_longitudinal/wheel_front: 0.0
```
---
###  Case 4 :  Individual wheel ROS params specified in world file + individual wheels params as sdf params

World file : 
```
<plugin name="wheel_slip_front" filename="libgazebo_ros_wheel_slip.so">
        <ros>
          <namespace>trisphere_cycle_slip0</namespace>
          <parameter name="slip_compliance_unitless_lateral/wheel_front" value="10.0" type="double"/>
          <parameter name="slip_compliance_unitless_longitudinal/wheel_front" value="11.0" type="double"/>
        </ros>

        <wheel link_name="wheel_front">
          <slip_compliance_lateral>1</slip_compliance_lateral>
          <slip_compliance_longitudinal>2</slip_compliance_longitudinal>
          <wheel_normal_force>77</wheel_normal_force>
        </wheel>
      </plugin>
```

Result : 
```
    slip_compliance_unitless_lateral: 1.0
    slip_compliance_unitless_lateral/wheel_front: 0.0
    slip_compliance_unitless_longitudinal: 2.0
    slip_compliance_unitless_longitudinal/wheel_front: 0.0
```
---
## Inference : 
* wheel slip parameters if not specified in the sdf, default to zero with no parsing errors. :green_circle: 
* Global parameter set action using ros (in world file) is ignored, and set to zero as default :red_circle: 
* Individual wheel parameter set action using ros (in world file) is ignored, and set to zero as default :red_circle: 

## Possible solutions : 
* Remove ROS params (from sdf) while creating the node, do a set parameters action towards the very end after all callbacks have been registered.
* Do not allow ROS parameters at all inside the world file, ignore them all. Filter them before creating the node.

jacobperron:
@adityapande-1995 Thanks for the detailing the possible scenarios!  Good fodder for tests.

Regarding possible solutions to the odd behavior, I'm hesitant about ignoring/removing ROS parameters in the sdf. For instance, if a user wants to extend the class and use ROS parameters, we don't want to prevent them from doing so.

I'm starting to wonder if we should revisit the idea of adding a ROS parameter for toggling the new behavior (ie. if the parameter is `true`, then declare parameters for individual wheels). I think it might be less confusing for a user if we avoid having the "global" wheel slip parameters accessible at the same time as individual wheel slip parameters.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -105,15 +124,123 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+
+        // Iterate over parameters
+        for (auto & parameter : param_list) {
+          auto temp_param = this->impl_->ros_node_->get_parameter(parameter.name);
+
+          // Propagate global slip compliance lateral parameter to all wheels
+          // This does not set the actual parameter in gazebo
+          if (parameter.name == "slip_compliance_unitless_lateral") {
+            double slip = temp_param.as_double();
+            if (slip >= 0.) {
```
jacobperron: I think this is handled by the validation callback and we can remove this case.

adityapande-1995: Done

adityapande-1995:
## Resolved all precedence issues in world file

### Current order of priority is (in the world file) :

ROS global wheelslip params (if not specified, default to last wheel values in sdf) > 
ROS individual wheelslip params (if not specified, default to sdf params for each wheel ) > 
sdf wheelslip params (default to 0 if not specified)

gazebo_plugins/test/CMakeLists.txt
```diff
@@ -9,6 +9,11 @@ foreach(world ${worlds})
   # Use configure_file so this is rerun each time make is invoked (as opposed to just cmake)
   configure_file(${world} ${world} COPYONLY)
 endforeach()
+file(GLOB worlds RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "worlds/wheelslip_worlds/*.world")
+foreach(world ${worlds})
+  # Use configure_file so this is rerun each time make is invoked (as opposed to just cmake)
+  configure_file(${world} ${world} COPYONLY)
+endforeach()
```
jacobperron: Rather than repeating this logic for the `wheelslip_worlds` subdirectory, I think we can use the [`GLOB_RECURSE`](https://cmake.org/cmake/help/latest/command/file.html#glob-recurse) option at L7.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include <string>
+#include <vector>
+
+using namespace std::chrono_literals;
+
+/// Tests the gazebo_ros_wheelslip plugin
+class GazeboRosWheelSlipTest : public gazebo::ServerFixture
+{
+public:
+  void TestParameters(
```
jacobperron: I think we generally follow the ROS style guide for gazebo_ros_pkgs (though the link in the contributing guide should be updated for ROS 2, see https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1333). So, I would use snake_case for method names.
```suggestion
  void test_parameters(
```

gazebo_plugins/test/worlds/wheelslip_worlds/gazebo_ros_wheelslip_2.world
```diff
@@ -0,0 +1,84 @@
...
+          <slip_compliance_lateral>-3</slip_compliance_lateral>
+          <slip_compliance_longitudinal>4</slip_compliance_longitudinal>
+          <wheel_normal_force>32</wheel_normal_force>
+        </wheel>
+        <wheel link_name="wheel_rear_right">
+          <slip_compliance_lateral>5</slip_compliance_lateral>
+          <slip_compliance_longitudinal>6</slip_compliance_longitudinal>
+          <wheel_normal_force>32</wheel_normal_force>
+        </wheel>
+      </plugin>
```
jacobperron: It doesn't appear that we're concerned with this instance of the plugin in the tests; can it be removed here?

gazebo_plugins/test/worlds/wheelslip_worlds/gazebo_ros_wheelslip_2.world
```diff
@@ -0,0 +1,84 @@
...
+      <plugin name="joint_control" filename="libJointControlPlugin.so">
```
jacobperron: Is this joint controller necessary for the tests or can we remove it from the world file?

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,213 @@
...
+      counter++;
+      if (counter > 5) {FAIL();}
+    }
+
+    // Verify the parameters
+    std::vector<std::string> parameter_names;
+    for (auto & element : parameter_pairs) {
+      parameter_names.push_back(element.first);
+    }
+    std::this_thread::sleep_for(1s);
```
jacobperron: I don't think this sleep should be necessary. We're already waiting up to 5 seconds above for the service to be ready.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,213 @@
...
+    for (auto & parameter : parameters_received) {
+      ASSERT_EQ(parameter_pairs[parameter.get_name()], parameter.as_double());
+    }
+  }
+};
+
+
+TEST_F(GazeboRosWheelSlipTest, TestWorldFile_2)
+{
+  // We have global ROS parameters and sdf parameters in the world file
```
jacobperron: I think the terms "global" and "local" ROS parameters might confuse readers. It might be better to be explicit about what we mean, e.g. "We set 'slip_compliance_unitless_lateral' and 'slip_compliance_unitless_longitudinal' ROS parameters ...". Same for similar comments below. And/or adding documentation to the gazebo_ros wheel slip plugin would help.

gazebo_plugins/test/worlds/wheelslip_worlds/gazebo_ros_wheelslip_1.world
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" ?>
+<!--
+  Gazebo ROS wheel slip plugin demo
+  Try for example:
+  To change slip compliance of a group of wheels:
+    ros2 param set /trisphere_cycle_slip0/wheel_slip_rear slip_compliance_unitless_longitudinal 1.0
+  To get the current slip compliance of a group of wheels:
+    ros2 param get /trisphere_cycle_slip0/wheel_slip_rear slip_compliance_unitless_longitudinal
+-->
```
jacobperron: I think we can remove this comment since this world file isn't really meant to be a demo that people run.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,213 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gazebo/test/ServerFixture.hh>
+#include <gazebo_ros/node.hpp>
+#include <gazebo_ros/testing_utils.hpp>
```
jacobperron: Are these two includes needed or can they be removed?

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,213 @@
...
+    std::this_thread::sleep_for(1s);
+    auto parameters_received = parameters_client->get_parameters(parameter_names);
+    for (auto & parameter : parameters_received) {
+      ASSERT_EQ(parameter_pairs[parameter.get_name()], parameter.as_double());
+    }
+  }
+};
+
+
+TEST_F(GazeboRosWheelSlipTest, TestWorldFile_2)
```
jacobperron: Looking at the implementation of each test, it's difficult to know _what_ we are testing. The comment is good, but it might be more obvious if we renamed the tests to something more informative, e.g. `TestRosParametersOverrideSdf`.

jacobperron (commented):
I think we still need to document the new parameters in the [header file](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp) and how they interact with SDF parameters.

gazebo_plugins/test/worlds/wheelslip_worlds/gazebo_ros_wheelslip_1.world
adityapande-1995: Removed the comment in all world files

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
adityapande-1995: Removed headers

gazebo_plugins/test/worlds/wheelslip_worlds/gazebo_ros_wheelslip_2.world
adityapande-1995: I'm using it in test case 1 (where we change parameters through command line) and test case 5. One of the parameters is set to -3, so it can be rejected and changed to 0 internally.

gazebo_plugins/test/worlds/wheelslip_worlds/gazebo_ros_wheelslip_2.world
adityapande-1995: Removed

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
adityapande-1995: Done

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
adityapande-1995: Some of the tests are failing randomly if I remove this sleep (expected parameter is say 10.0, but we get 0.0 from the ``SyncParametersClient``). I think this is because the topic service becomes available after we declare the ``gazebo_ros::Node`` with the ``sdf``, but the parameters are set later in the plugin. If we query the parameters between these events, they will be returned as 0.

gazebo_plugins/test/worlds/wheelslip_worlds/gazebo_ros_wheelslip_2.world
jacobperron: But if I understand correctly, this world file is only used in test case 2.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
jacobperron: Ah, that's tricky. I'm worried that this arbitrary sleep could be flaky or introduce extra time unnecessarily to the tests. IMO, it would be better to wait until the parameters match the expected values or timeout (e.g. continually call get_parameters until our condition is satisfied or a timeout occurs).

adityapande-1995: Yeah that makes sense. I'll modify the tests accordingly

gazebo_plugins/test/CMakeLists.txt
adityapande-1995: Done

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
adityapande-1995: Done

gazebo_plugins/test/worlds/wheelslip_worlds/gazebo_ros_wheelslip_2.world
adityapande-1995: Removed unnecessary plugins in world files.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
adityapande-1995: Added better comments for each test case. Will update the header as well.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
adityapande-1995: Renamed test cases

adityapande-1995:
Done : Add explanation about parameters to the header file :heavy_check_mark:

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -24,9 +24,12 @@ namespace gazebo_plugins
 {
 class GazeboRosWheelSlipPrivate;
 
+/// Description:
+/// ------------
```
jacobperron: I think common tools like Doxygen usually take the first line in a comment as the "brief" or "summary" when rendered as html for example. I don't think we're generating html docs for gazebo_ros_pkgs (yet), but I would remove these lines for posterity.
```suggestion
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -24,9 +24,12 @@ namespace gazebo_plugins
...
 /// A plugin for adjusting wheel slip parameters in gazebo.
 /// The plugin can set separate longitudinal and lateral wheel slip compliance
-/// parameters for separate wheel links.
+/// parameters for separate wheel links. THe following ROS parameters can be specified in
+/// the world file (These are optional, you may not may not declare them):
```
jacobperron: style nit: prefer one line per sentence for documentation. This generally makes diffs more readable and in my experience does not have an affect on generated documentation.

I've also applied some wording suggestions:
```suggestion
/// parameters for separate wheel links.
/// The following ROS parameters can optionally be specified in the world file
```

I'm not sure what was meant by "you may not declare them". I guess you mean subclasses shouldn't _also_ try to declare these parameters, but that's a general rule for all ROS parameters so I'm not sure it's worth calling attention to it here.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -38,6 +41,57 @@ class GazeboRosWheelSlipPrivate;
 ///      - Description: Unitless slip compliance (slip / friction) in the
 ///           longitudinal direction. This value is applied to all wheels declared
 ///           in the WheelSlipPlugin.
+///
+///  3. slip_compliance_unitless_lateral/wheel_name
+///      - Type: double
+///      - Description: Unitless slip compliance (slip / friction) in the
+///           lateral direction. This value is applied to the wheel named 'wheel_name'
+///           declaredin in the WheelSlipPlugin.
```
jacobperron:
```suggestion
///           declared in in the WheelSlipPlugin.
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -38,6 +41,57 @@ class GazeboRosWheelSlipPrivate;
...
+///
+///  4. slip_compliance_unitless_longitudinal/wheel_name
+///      - Type: double
+///      - Description: Unitless slip compliance (slip / friction) in the
+///           longitudinal direction. This value is applied to the wheel named 'wheel_name'
+///           declared in the WheelSlipPlugin.
+///
+/// Precedence order and deafult values:
+/// -----------------------------------
```
jacobperron:
```suggestion
/// Precedence order and default values:
/// ------------------------------------
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -38,6 +41,57 @@ class GazeboRosWheelSlipPrivate;
...
+/// Slip compliance values can be declared in 3 ways:
+///  1. SDF parameters, for e.g. :
+///       <wheel link_name="wheel_front">
+///         <slip_compliance_lateral>1</slip_compliance_lateral>
+///         <slip_compliance_longitudinal>2</slip_compliance_longitudinal>
+///         <wheel_normal_force>77</wheel_normal_force>
+///       </wheel>
+///
+///      These if not specified, will default to 0.0. Any negative values will ignored and
+//       set to 0.0
```
jacobperron:
```suggestion
///      If not specified, these will default to 0.0.
///      Any negative values will be ignored and set to 0.0.
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -38,6 +41,57 @@ class GazeboRosWheelSlipPrivate;
...
+///  4. slip_compliance_unitless_longitudinal/wheel_name
+///      - Type: double
+///      - Description: Unitless slip compliance (slip / friction) in the
+///           longitudinal direction. This value is applied to the wheel named 'wheel_name'
+///           declared in the WheelSlipPlugin.
+///
+/// Precedence order and deafult values:
+/// -----------------------------------
+/// Slip compliance values can be declared in 3 ways:
+///  1. SDF parameters, for e.g. :
```
jacobperron:
```suggestion
///  1. SDF parameters, e.g. :
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -38,6 +41,57 @@ class GazeboRosWheelSlipPrivate;
...
+///       <wheel link_name="wheel_front">
+///         <slip_compliance_lateral>1</slip_compliance_lateral>
+///         <slip_compliance_longitudinal>2</slip_compliance_longitudinal>
+///         <wheel_normal_force>77</wheel_normal_force>
+///       </wheel>
+///
+///      These if not specified, will default to 0.0. Any negative values will ignored and
+//       set to 0.0
+///
+///  2. ROS parameters for each wheel, for e.g :
```
jacobperron:
```suggestion
///  2. ROS parameters for each wheel, e.g :
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -38,6 +41,57 @@ class GazeboRosWheelSlipPrivate;
...
+///       <ros>
+///         <parameter name="slip_compliance_unitless_lateral/wheel_front" type="double">10.0
+///           </parameter>
+///         <parameter name="slip_compliance_unitless_longitudinal/wheel_front" type="double">11.0
+///           </parameter>
+///       </ros>
+///
+///      These if not specified, default to SDF parameters set for each wheel. If specified, they
+///      will override the ones set as SDF parameters.
```
jacobperron:
```suggestion
///      If not specified, these default to SDF parameters set for each wheel.
///      If specified, they will override the values set as SDF parameters.
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -38,6 +41,57 @@ class GazeboRosWheelSlipPrivate;
...
+///
+///  3. ROS parameters for all wheels, for e.g :
```
jacobperron:
```suggestion
///  3. ROS parameters for all wheels, e.g :
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -38,6 +41,57 @@ class GazeboRosWheelSlipPrivate;
...
+///      <ros>
+///         <parameter name="slip_compliance_unitless_lateral" type="double">10.0</parameter>
+///         <parameter name="slip_compliance_unitless_longitudinal" type="double">11.0</parameter>
+///      </ros>
+///
+///     These if not specified will default to the last ones set in the SDF tags. In this case,
+///     they won't override anything.
```
jacobperron:
```suggestion
///     If not specified, these will default to the last values set in SDF tags.
```

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -47,6 +101,10 @@ class GazeboRosWheelSlipPrivate;
     <plugin name="projector" filename="libgazebo_ros_wheel_speed.so">
       <ros>
         <namespace>wheel_slip_front</namespace>
+        <parameter name="slip_compliance_unitless_lateral/wheel_front_left" type="double">10.0</parameter>
+        <parameter name="slip_compliance_unitless_longitudinal/wheel_front_left" type="double">11.0</parameter>
+        <parameter name="slip_compliance_unitless_lateral" type="double">10.0</parameter>
+        <parameter name="slip_compliance_unitless_longitudinal" type="double">11.0</parameter>
```
jacobperron: minor: I notice that the example values used for the ROS parameters here (and above) are an order of magnitude larger than the example values for SDF value above and below. I think we should change the example parameter values to be on a similar scale as the SDF example values. IIUC, a value of `10.0` is actually quite dramatic.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,276 @@
...
+          // Parameters don't match
+          flag_do_parameters_match = false;
+          break;
+        }
+      }
+
+      // Check the flag_do_parameters_match, if it was true, no need to query again
+      if (flag_do_parameters_match) {break;}
+      std::this_thread::sleep_for(0.2s);
+    }
```
jacobperron, line 77: Here's a solution where we avoid sleeping altogether (I haven't actually tested if this compiles :grimacing:):

```c++
    bool all_parameters_match = false;
    auto start_time = std::chrono::steady_clock::now();
    while (!all_parameters_match && (std::chrono::steady_clock::now() - start_time) < 1s) {
      all_parameters_match = true;

      // Query for parameters
      auto parameters_received = parameters_client->get_parameters(parameter_names);
      for (auto & parameter : parameters_received) {
        if (parameter_pairs[parameter.get_name()] != parameter.as_double()) {
          all_parameters_match = false;
          break;
        }
      }
    }
```

If the time we need to wait if very short, then I _think_ this sleepless wait is more efficient. I don't feel strongly one way or the other, just suggesting an alternative. We can leave the currently implementation as-is.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,276 @@
...
+    auto parameters_client = std::make_shared<rclcpp::SyncParametersClient>(
+      node, node_name);
+
+    // Wait for serive to show up
+    int counter = 0;
+    while (!parameters_client->wait_for_service(1s)) {
+      std::cout << "service not available, waiting again..." << std::endl;
+      counter++;
+      if (counter > 5) {FAIL();}
+    }
```
jacobperron: I think this is more or less equivalent to:
```suggestion
    ASSERT_TRUE(parameters_client->wait_for_service(5s))
```

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {"slip_compliance_unitless_longitudinal/wheel_front", 11.0}};
+
+  this->test_parameters(
+    "worlds/wheelslip_worlds/gazebo_ros_wheelslip_4.world",
+    "trisphere_cycle_slip0/wheel_slip_front",
+    parameter_pairs
+  );
+}
+
+TEST_F(GazeboRosWheelSlipTest, TestRosGloabalParamsOverrideAll)
```
jacobperron:
```suggestion
TEST_F(GazeboRosWheelSlipTest, TestRosGlobalParamsOverrideAll)
```

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {"slip_compliance_unitless_longitudinal/wheel_rear_left", 4},
+    {"slip_compliance_unitless_longitudinal/wheel_rear_right", 6}};
+
+  // TEST 1 : Verify parameters were set as per the SDF, negative values should be replaced by 0
+  int counter = 0;
+  while (!parameters_client->wait_for_service(1s)) {
+    std::cout << "service not available, waiting again..." << std::endl;
+    counter++;
+    if (counter > 5) {FAIL();}
+  }
```
jacobperron:
```suggestion
  ASSERT_TRUE(parameters_client->wait_for_service(5s));
```

jacobperron (commented):
Thanks for adding documentation and iterating with me!

I've left some wording suggestions and minor refactor suggestions for the tests.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
adityapande-1995: Done

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
adityapande-1995: Resolved

jacobperron: It doesn't look like anything changed in the docs; we're still using large values for ROS parameters and smaller values for SDF.

adityapande-1995: Oh sorry I changed it in the actual world files for test cases, not the docs. Will do

jacobperron (approved):
LGTM. Just one outstanding comment above related to documentation.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
adityapande-1995: Done

adityapande-1995:
Failing test is a known issue, and seems unrelated to this PR : https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1304

PR #1314, opened 2021-09-22
Fix unit tests: need to expect performance_metrics topic, adjust expected range minimum

lucasw:
Maybe `ros_network_default.test` has been broken since the performance_metrics topic was introduced in 1ab2668215c2bd582868cd988ce2e579037b188a ?

### DIFF
--- a/gazebo_plugins/scripts/test_range.py
+++ b/gazebo_plugins/scripts/test_range.py
@@ -12,7 +12,12 @@ class TestRangePlugin(unittest.TestCase):
 
   def test_inside_range(self):
     msg = rospy.wait_for_message('/sonar', Range)
-    self.assertTrue(msg.range < 0.25 and msg.range > 0.22, 'actual value: {0}'.format(msg.range))
+    # Nominal value is 0.225
+    # TODO(lucasw) could have this average over 10 samples or so
+    min_range = rospy.get_param("~min_range", 0.20)
+    max_range = rospy.get_param("~max_range", 0.25)
+    self.assertGreater(msg.range, min_range, 'range too small: {} < {}'.format(msg.range, min_range))
+    self.assertLess(msg.range, max_range, 'range too big: {} > {}'.format(msg.range, max_range))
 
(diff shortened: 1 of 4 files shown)


lucasw:
Hmm now a different test is failing

```
 [gazebo_plugins.rosunit-test_range_plugin/test_inside_range][FAILURE]-----------
2: False is not true : actual value: 0.21732348203659058
2:   File "/usr/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
2:     yield
2:   File "/usr/lib/python3.8/unittest/case.py", line 676, in run
2:     self._callTestMethod(testMethod)
2:   File "/usr/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
2:     method()
2:   File "/var/lib/jenkins/workspace/ros_gazebo_pkgs-ci-pr_any_noetic-focal-amd64/ws/src/gazebo_ros_pkgs/gazebo_plugins/scripts/test_range.py", line 15, in test_inside_range
2:     self.assertTrue(msg.range < 0.25 and msg.range > 0.22, 'actual value: {0}'.format(msg.range))
2:   File "/usr/lib/python3.8/unittest/case.py", line 765, in assertTrue
2:     raise self.failureException(msg)```

**Comment by lucasw** [comment]
#1290, #1289, #1280, #1271 and probably many more look to have been failing from the range or network checker tests, which this fixes.

The block laser test is timing out in `test_block_laser_clipping.py` on my system when I run 

```
catkin build gazebo_plugins --no-deps --catkin-make-args run_tests
```

maybe due to re-use of gazebo ports while running tests simultaneously, but I haven't seen it failing in CI (maybe due to single threading there).

I see these 

```
[Err] [Master.cc:95] EXCEPTION: Unable to start server[bind: Address already in use]. There is probably another Gazebo process running.

[Err] [Master.cc:95] EXCEPTION: Unable to start server[bind: Address already in use]. There is probably another Gazebo process running.
```

but they aren't always correlated with test failures.

Issue #1301, opened 2021-08-01
[ros2] gazebo plugins should publish /tf topic as /namespace/tf if the robot is spawned with namespace


sathak93:
gazebo_plugins publishes  /odom, /scan, /imu and other topics with namespaces but not the  /tf topic when spawned the robot with spawn_entity.py
![Screenshot from 2021-08-01 19-55-43](https://user-images.githubusercontent.com/57533054/127774562-d1184ff9-c5cb-4b1d-9394-ac5df6a15a5c.png)

bjsowa:
The `/tf` and `/tf_static` are always published on the global namespace. That's how the tf framework works.

sathak93:
@bjsowa  how to configure global namespace in launch file.
also  robot_state_publisher also uses tf framework why would it publishes with namespace if it is started with namespace?.
 if we spawn multiple robots how do we differentiate the /tf topipc from one another.
![Screenshot from 2021-08-03 20-52-01](https://user-images.githubusercontent.com/57533054/128042519-882ff98b-7981-43b5-b523-221c5494ca47.png)

SteveMacenski:
Also this is useful for separations for multi-robot configurations. The spawner we have in Nav2 creates the TF remap. Ideally, we'd like to delete that and use the gazebo_ros's but the blocking reason we can't is that we cannot remap the tf.

bjsowa:
> @bjsowa how to configure global namespace in launch file.
> also robot_state_publisher also uses tf framework why would it publishes with namespace if it is started with namespace?.
> if we spawn multiple robots how do we differentiate the /tf topipc from one another.
> ![Screenshot from 2021-08-03 20-52-01](https://user-images.githubusercontent.com/57533054/128042519-882ff98b-7981-43b5-b523-221c5494ca47.png)

I don't know how you launched the `robot_state_publisher` but when I tried running it under a namespace, the `/tf` topic stays under global namespace:

![rosgraph](https://user-images.githubusercontent.com/23124539/128492121-eab3f0aa-c254-4875-94fa-091cc6707833.png)

Are you sure you did not use remapping?

bjsowa:
If you look at the code of tf2_ros::TransformBroadcaster:

https://github.com/ros2/geometry2/blob/5f30fe8fcc9eacde580fb1374ae82182e7fb5bb6/tf2_ros/include/tf2_ros/transform_broadcaster.h#L58-L61

[...]

PR #1319, opened 2021-10-07
Set default namespace to model name if <namespace> is not specified

### LINKED ISSUE #1320: Gazebo plugins default to root namespace if <ros><namespace> is not specified
When robots are spawned, e.g. by inserting the models through the GUI, the plugin topics occupy the root (`/`) namespace if the `<ros><namespace>` is unspecified. This causes problems when trying to insert multiple robot models with the same SDF / URDF. The topics will all conflict. 

The change to use the root namespace was introduced in this PR to fix tests due to exceptions: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1130

It seem like the more desirable behavior is to default the namespace to the parent model name if `<ros><namespace>` unspecified, which is then [consistent with ros1](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/noetic-devel/gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h#L132-L139).

iche033:
Signed-off-by: Ian Chen <ichen@osrfoundation.org>

Currently the namespace for gazebo_plugins defaults to the root namespace (`/`) if the `<ros><namespace>` SDF element is not specified. 

This PR updates the namespace logic to be [consistent with ros1](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/noetic-devel/gazebo_plugins/include/gazebo_plugins/gazebo_ros_utils.h#L132-L139), in which the default namespace becomes the parent robot name if left unspecified. 
 
fixes #1320

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_video.cpp
+++ b/gazebo_plugins/src/gazebo_ros_video.cpp
@@ -197,9 +197,7 @@ void GazeboRosVideo::Load(
   gazebo::rendering::VisualPtr _parent,
   sdf::ElementPtr _sdf)
 {
-  impl_->rosnode_ = gazebo_ros::Node::Get(
-    _sdf,
-    _parent->GetRootVisual()->Name());
+  impl_->rosnode_ = gazebo_ros::Node::Get(_sdf, _parent);
 
   // Get QoS profiles
   const gazebo_ros::QoS & qos = impl_->rosnode_->get_qos();
--- a/gazebo_plugins/src/gazebo_ros_video.cpp
+++ b/gazebo_plugins/src/gazebo_ros_video.cpp
@@ -198,8 +198,8 @@ void GazeboRosVideo::Load(
   sdf::ElementPtr _sdf)
 {
   impl_->rosnode_ = gazebo_ros::Node::Get(
-      _sdf,
-      _parent->GetRootVisual()->Name());
+    _sdf,
+    _parent->GetRootVisual()->Name());
 
   // Get QoS profiles
   const gazebo_ros::QoS & qos = impl_->rosnode_->get_qos();
--- a/gazebo_plugins/src/gazebo_ros_video.cpp
+++ b/gazebo_plugins/src/gazebo_ros_video.cpp
@@ -197,7 +197,8 @@ void GazeboRosVideo::Load(
   gazebo::rendering::VisualPtr _parent,
   sdf::ElementPtr _sdf)
 {
-  impl_->rosnode_ = gazebo_ros::Node::Get(_sdf,
+  impl_->rosnode_ = gazebo_ros::Node::Get(
+      _sdf,
       _parent->GetRootVisual()->Name());
 
   // Get QoS profiles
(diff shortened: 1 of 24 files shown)


chapulina (commented):
The general idea sounds good to me. Just note that this changes the behavior of released plugins, which may break downstream users. Galactic is relatively young though, so it may not impact a lot of people.

Note that there are linter errors on CI.

scpeters:
should we target `ros2` first and then consider back-porting to `galactic`?

iche033:
I added a new param, named `<legacy_namespace>`, that defaults to `true` so that the current behavior of using the `/` namespace is preserved. With this new param, users can now switch to use the parent model name as the new namespace by setting it to `false`. 

Once this gets in, I can create a separate PR porting this fix to `ros2` and set `<legacy_namespace>` to false

padhupradheep:
without looking much into the code, will this also solve the #1301 issue related to the /tf topic ?

iche033:
> without looking much into the code, will this also solve the #1301 issue related to the /tf topic ?

probably not as it looks like the transform broadcaster publishes them to absolute namespace
https://github.com/ros2/geometry2/blob/89852f731710061fbd99d7d812523baac6f59a88/tf2_ros/include/tf2_ros/transform_broadcaster.h#L72-L73

gazebo_plugins/src/gazebo_ros_video.cpp
```diff
@@ -197,7 +197,9 @@ void GazeboRosVideo::Load(
   gazebo::rendering::VisualPtr _parent,
   sdf::ElementPtr _sdf)
 {
-  impl_->rosnode_ = gazebo_ros::Node::Get(_sdf);
+  impl_->rosnode_ = gazebo_ros::Node::Get(
+    _sdf,
+    _parent->GetRootVisual()->Name());
```
chapulina: Just curious, why not add a `Get` overload that takes in `VisualPtr`, like it was done for models and sensors?

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -66,6 +69,10 @@ class GAZEBO_ROS_NODE_PUBLIC Node : public rclcpp::Node
    *   <ros>
    *    <!-- Namespace of the node -->
    *    <namespace>/my_ns</namespace>
+   *    <!-- Legacy namespace behavior. True to default to the root namespace
+   *         if <namespace> is not specified, otherwise false to default
+   *         namespace to the the model name -->
```
chapulina, line 75: Maybe mention that it defaults to true if `legacy_namespace` isn't set?

chapulina (approved):
LGTM now that it's an opt-in feature

gazebo_ros/include/gazebo_ros/node.hpp
scpeters: suggested rephrasing:

~~~
Legacy behavior for setting namespace when <namespace> is not specified.
When <legacy_namespace> is unspecified or set to true, the root namespace `/` is the default.
When <legacy_namespace> is set to false, the default namespace is the model name.
~~~

iche033: done, updated doc. f24ac84

gazebo_plugins/src/gazebo_ros_video.cpp
iche033: yeah that works. It wasn't done in the ros1 version so I also didn't add an overload for it before. I just added it. f24ac84

PR #1325, opened 2021-11-01
PublishSimTime: changed Throttler to use simTime instead of realTime

### LINKED ISSUE #1324: ROS2 Nodes stay paused when stepping Gazebo simulation
We've noticed that when we have ROS2 nodes using `sim_time` interfacing with our Gazebo simulation, all of our nodes are unresponsive when stepping through the simulation.

The nodes behave completely as expected when running the simulation normally (unpaused). They "pause" as expected if we pause the simulation. But when we try and step the simulation, nothing happens. In Gazebo, we can see the simulation time increasing, as well as any dynamics within the simulation working as expected. But everything on the ROS front remains "paused". This is perhaps best illustrated (and perhaps caused) by the `/clock` topic. We can monitor the `/clock` topic, and confirm that it updates as expected when the simulation plays, but when we step the simulation manually, nothing is ever published. We can manually step 10 seconds ahead, and then if we unpause the simulation, the `/clock` starts publishing again with a 10 second jump. 

This is an issue for us as we'd like to have a gazebo-simulink co-simulation which works by stepping the gazebo sim for every time step in the simulink sim. But this behavior prevents anything else from working.

We've tested this in ROS2 foxy. Interestingly, in ROS1 Melodic, topics seem to continue to update if we manually step the simulation. So perhaps an implementation was changed or a feature has yet to be ported between the ROS1 and ROS2 versions of gazebo_ros_pkgs? 

Happy to help try and fix this, but not sure where to start looking. Any help would be welcome!

shonigmann:
Changed `Throttler` for the clock publisher in [gazebo_ros_init.cpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/3310d0d01cfcd4a2710d834d2abf8bb6a620a888/gazebo_ros/src/gazebo_ros_init.cpp#L271-L292) to use `simTime` instead of `realTime`. This should fix #1324 and should allow ROS applications to continue to cooperate with Gazebo when the simulation is paused and is being stepped. 

If this doesn't break anything, I can open new PRs for the rest of the ros2 distro-specific branches (foxy, galactic, etc) that have the same issue.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_init.cpp
+++ b/gazebo_ros/src/gazebo_ros_init.cpp
@@ -270,7 +270,7 @@ GazeboRosInitPrivate::GazeboRosInitPrivate()
 
 void GazeboRosInitPrivate::PublishSimTime(const gazebo::common::UpdateInfo & _info)
 {
-  if (!throttler_.IsReady(_info.realTime)) {
+  if (!throttler_.IsReady(_info.simTime)) {
     return;
   }
 


jmachowinski (approved):

Zyrin:
Bump, as we ran into this one. @scpeters please review this.

scpeters (approved):

scpeters:
thanks for your contribution!

PR #1326, opened 2021-11-03
adding namespaces to the tf_frames for ROS2

padhupradheep:
Currently, the robot_state_publisher comes with an argument called as the `frame_prefix` for adding a prefix to the frames. Also, even if there is a `frame_prefix` added to the robot_state_publisher, the plugins from this package does not broadcast the frames with the prefix. This missing feature, does not allow an ease of developing a multi-robot simulation. 

Assuming that the namespaces and the `frame_prefix` will be defined the same by the users, I added the namespaces to the frames. For example: `odom` will be broadcasted as `my_ns/odom` . 

Do let me know your thoughts ? Maybe there is also someother way to achieve this!

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_planar_move.cpp
+++ b/gazebo_plugins/src/gazebo_ros_planar_move.cpp
@@ -114,6 +114,8 @@ public:
 
   /// True to publish odom-to-world transforms.
   bool publish_odom_tf_;
+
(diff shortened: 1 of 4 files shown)


padhupradheep:
This can be easily achieved through xacros, no need to get into the core of the plugin

PR #1331, opened 2021-11-10
gazebo_ros_wheel_slip: publish wheel slip

audrow:
This PR adds a publisher to the `gazebo_ros_wheel_slip` plugin. The plugin uses the `gazebo::WheelSlipPlugin::getSlips()` method to get each wheel's slip and then computes the normalized wheel slip for each wheel. The lateral slip is given in degrees slipped and the longitudinal slip is given as a percentage of the current wheel's speed.

This PR also creates a message type `InstantSlip.msg` for the wheel slip, which has a similar structure to ``sensor_msgs::msg::JointState`` in that it uses a `name` string array to index float arrays where the lateral and wheel slip values are stored.

Also, this PR uses an approach to updating similar to the `gazebo_ros_joint_state_publisher` plugin. In that the user can use the SDF to set the update rate and a callback is called on each update. One difference between this implementation and the `gazebo_ros_joint_state_publisher` is that this PR uses a lambda function for the callback, rather than defining the callback in a pointer to an implementation (PIMPL). The reason for this approach is that the `gazebo::WheelSlipPlugin::getSlips()` function needs to be called, but the PIMPL doesn't have access to that function since it doesn't inherit from `gazebo::WheelSlipPlugin`.

I'm still thinking about how to add tests for this plugin. Thoughts and feedback are welcome.


gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -19,10 +19,17 @@
...
 #include <std_msgs/msg/bool.hpp>
+#include <rclcpp/rclcpp.hpp>
+#include "gazebo_msgs/msg/instant_slip.hpp"
 
 #include <memory>
 #include <string>
 #include <vector>
+#include <cmath>
+
+#define PI 3.14159265
```
scpeters: use `M_PI` from the standard headers or `IGN_PI` from `ignition/math/Helpers.hh`

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -114,6 +133,93 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+    #ifdef IGN_PROFILER_ENABLE
+      IGN_PROFILE_BEGIN("fill ROS message");
+    #endif
+
+    // Populate message
+    auto slip_msg = gazebo_msgs::msg::InstantSlip();
+
+    std::map<std::string, ignition::math::Vector3d> slips;
+    this->GetSlips(slips);
+    for(const auto& elem : slips)
```
scpeters: nit: I would say `wheel` instead of `elem` here, since `elem` makes me think of XML and SDFormat

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -114,6 +133,93 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+    {
+      auto name = elem.first;
+      auto slip = elem.second;
+      auto spin_speed = slip.Z();
+      auto long_vel = slip.X() + slip.Z();
+      auto lat_vel = slip.Y();
+      double long_slip;
+      if (spin_speed == 0) {
```
scpeters: I would check if it's near 0 within a specified tolerance

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -114,6 +133,93 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+        long_slip = 0;
+      } else {
+        long_slip = (long_vel-spin_speed) / spin_speed;
+      }
+      auto lat_slip = atan2(lat_vel, long_vel) * 180 / PI;
```
scpeters: radians are typically used according to https://www.ros.org/reps/rep-0103.html#units

scpeters (commented):
> I'm still thinking about how to add tests for this plugin. Thoughts and feedback are welcome.

you could use the demo world for a test:

* https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/worlds/gazebo_ros_wheel_slip_demo.world

scpeters:
can you review the CI results and fix the linting errors?

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Good call.

audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/fdeefa9cbd3b46447921e4b24df66fb59ce603c6.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Removed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/fdeefa9cbd3b46447921e4b24df66fb59ce603c6.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/cc362daa078258471e93f910e3ec48e7d52f1af8.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/87a345b02c112f6e8b952dd912f41f7869d09c9d.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -114,6 +131,100 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+    this->GetSlips(slips);
+    for(const auto& wheel : slips)
+    {
+      auto name = wheel.first;
+      auto slip = wheel.second;
+      auto spin_speed = slip.Z();
+      auto long_vel = slip.X() + slip.Z();
+      auto lat_vel = slip.Y();
+      double long_slip;
+      if (spin_speed < zero_wheel_spin_tolerance) {
```
scpeters: I'm not sure if spin_speed is always positive, so I would either use an absolute value function or ignition::math::equal(spin_speed, 0, zero_wheel_spin_tolerance) (from ignition/math/Helpers.hh)

audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/3e0dfc01a62dec2ffdd1a754970d14fc31b462a9.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -114,6 +132,103 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+    this->GetSlips(slips);
+    for(const auto& wheel : slips)
+    {
+      auto name = wheel.first;
+      auto slip = wheel.second;
+      auto spin_speed = slip.Z();
+      auto long_vel = slip.X() + slip.Z();
+      auto lat_vel = slip.Y();
+      double long_slip;
+      if (abs(spin_speed) < zero_wheel_spin_tolerance) {
```
scpeters: I think you want `fabs` instead of `abs`

scpeters:
> can you review the CI results and fix the linting errors?

looks like a few uncrustify errors still

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: I think that [`abs` is overloaded for floating point values](https://www.cplusplus.com/reference/cmath/abs/), but I've changed it in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/5f587ea13c08992e745c97cb67cd9224e3f8d4a6.

audrow:
@scpeters, no need to review just yet. I'll look into why the message isn't getting passed correctly in the test. (It was working on my local machine.)

audrow:
The failing test seems to be related to https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1304.

audrow:
@scpeters, this should be ready for review now.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
scpeters: I believe `std::abs` is overloaded for floating-point types but the C function `::abs` only handles `int`

https://stackoverflow.com/a/33738702

gazebo_msgs/msg/InstantSlip.msg
```diff
@@ -0,0 +1,3 @@
+string[] name
```
scpeters: can you add documentation to this message file? you can use the joint states message as a reference:

* https://github.com/ros2/common_interfaces/blob/master/sensor_msgs/msg/JointState.msg

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -256,6 +272,102 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+  if (!_sdf->HasElement("zero_wheel_spin_tolerance")) {
+    RCLCPP_INFO(
+      impl_->ros_node_->get_logger(), "Missing <zero_wheel_spin_tolerance>, defaults to %f",
+      zero_wheel_spin_tolerance);
+  } else {
+    zero_wheel_spin_tolerance = _sdf->GetElement("zero_wheel_spin_tolerance")->Get<double>();
+  }
+
+  double update_rate = 100.0;
+  if (!_sdf->HasElement("update_rate")) {
```
scpeters: if this only applies to the publisher, can you rename it as `publisher_update_rate`? The underlying `WheelSlipPlugin` itself still runs with every physics step

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -256,6 +272,102 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
       }
     }
   }
+
+  double zero_wheel_spin_tolerance = 1.0e-3;
+  if (!_sdf->HasElement("zero_wheel_spin_tolerance")) {
```
scpeters: nit: rename `zero_wheel_spin_tolerance` -> `wheel_spin_tolerance`

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -256,6 +272,102 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+  impl_->last_update_time_ = _model->GetWorld()->SimTime();
+
+  impl_->slip_publisher_ = impl_->ros_node_->create_publisher<gazebo_msgs::msg::InstantSlip>(
+    "wheel_slip", 10);
+
+  auto on_update_callback = [this, zero_wheel_spin_tolerance](
+    const gazebo::common::UpdateInfo & info) {
+
+    #ifdef IGN_PROFILER_ENABLE
+      IGN_PROFILE("instant slip callback");
```
scpeters: nit: use `"GazeboRosWheelSlip publisher"`

scpeters (changes requested):
Can you rename the message file from `InstantSlip.msg` to `WheelSlip.msg`?

gazebo_msgs/msg/InstantSlip.msg
audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/74450d05243da4f06905d7551590e167e905c534.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Good call, fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/53c792ba286876e9e477153ee02cdb5f8fa825e6.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/53c792ba286876e9e477153ee02cdb5f8fa825e6.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/53c792ba286876e9e477153ee02cdb5f8fa825e6.

gazebo_plugins/test/worlds/gazebo_ros_wheel_slip.world
```diff
@@ -0,0 +1,69 @@
...
+        <wheel link_name="wheel_front">
+          <slip_compliance_lateral>1</slip_compliance_lateral>
+          <slip_compliance_longitudinal>1</slip_compliance_longitudinal>
+          <wheel_normal_force>77</wheel_normal_force>
+        </wheel>
+      </plugin>
+      <plugin name="wheel_slip_rear" filename="libgazebo_ros_wheel_slip.so">
+        <ros>
+          <namespace>trisphere_cycle_slip</namespace>
+       </ros>
```
scpeters: nit: indentation

gazebo_plugins/test/worlds/gazebo_ros_wheel_slip.world
```diff
@@ -0,0 +1,69 @@
...
+    </include>
+
+    <include>
+      <uri>model://trisphere_cycle</uri>
+      <name>trisphere_cycle_slip</name>
+      <pose>0 2 0  0 0 0</pose>
+      <plugin name="wheel_slip_front" filename="libgazebo_ros_wheel_slip.so">
+        <ros>
+          <namespace>trisphere_cycle_slip</namespace>
+       </ros>
```
scpeters: nit: indentation

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -45,6 +51,17 @@ class GazeboRosWheelSlipPrivate
...
+  rclcpp::Publisher<gazebo_msgs::msg::WheelSlip>::SharedPtr slip_publisher_;
+
+  /// Period in seconds
+  double update_period_;
+
+  /// Keep last time an update was published
+  gazebo::common::Time last_update_time_;
+
+  /// Pointer to the update event connection.
+  gazebo::event::ConnectionPtr update_connection_;
```
scpeters: please add `publisher_` as a prefix to `update` in these 3 new variable names to match the changes in 53c792ba286876e9e477153ee02cdb5f8fa825e6

scpeters (commented):
can you document the new SDFormat XML parameters in the header file?

* https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp

gazebo_plugins/test/worlds/gazebo_ros_wheel_slip.world
audrow: Good catch. I fixed this and the file that I referenced in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/3a57611aca307213d3fc0ad9b16a1622e81107f5.

gazebo_plugins/test/worlds/gazebo_ros_wheel_slip.world
audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/3a57611aca307213d3fc0ad9b16a1622e81107f5.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/be39f37ef3703a961f54c66110e75315f5d3b825.

audrow:
> can you document the new SDFormat XML parameters in the header file?

I documented the new parameters in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/91cc8fc64bd048de2362eb3c65d12683805ebf4b.

scpeters:
code is looking good; I'll test it tomorrow

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -256,6 +272,105 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+        auto name = wheel.first;
+        auto slip = wheel.second;
+        auto spin_speed = slip.Z();
+        auto long_vel = slip.X() + slip.Z();
+        auto lat_vel = slip.Y();
+        double long_slip;
+        if (fabs(spin_speed) < wheel_spin_tolerance) {
+          long_slip = 0;
+        } else {
+          long_slip = (long_vel - spin_speed) / spin_speed;
```
scpeters: I think this doesn't follow the typical sign convention for longitudinal slip, which corresponds to positive slip during acceleration and negative slip during deceleration

see equation 1 in the following paper for example:

* http://www-cdr.stanford.edu/dynamic/WheelSlip/SlmillerGerdesACC.pdf

I noticed in the demo world that the published values are negative, when the vehicles are both driving against gravity, which is like accelerating so I would expect a positive slip value

audrow:

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -256,6 +272,105 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+        auto name = wheel.first;
+        auto slip = wheel.second;
+        auto spin_speed = slip.Z();
+        auto long_vel = slip.X() + slip.Z();
+        auto lat_vel = slip.Y();
+        double long_slip;
+        if (fabs(spin_speed) < wheel_spin_tolerance) {
+          long_slip = 0;
+        } else {
+          long_slip = -1.0 * (long_vel - spin_speed) / spin_speed;
```
scpeters: nit: this might be easier to read as `long_slip = (spin_speed - long_vel) / spin_speed;`

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -258,6 +263,57 @@ TEST_F(GazeboRosWheelSlipTest, TestSetParameters)
...
+  // Check that we receive the latest joint state
+  ASSERT_NE(nullptr, latestMsg);
+
+  ASSERT_LT(0u, latestMsg->name.size());
+  ASSERT_GT(4u, latestMsg->name.size());
+  ASSERT_EQ(latestMsg->name.size(), latestMsg->lateral_slip.size());
+  ASSERT_EQ(latestMsg->name.size(), latestMsg->longitudinal_slip.size());
+
+  EXPECT_NEAR(0.0, latestMsg->lateral_slip[0], tol);
+  ASSERT_LT(0.0, fabs(latestMsg->longitudinal_slip[0]));
```
scpeters: this expectation doesn't make sense to me. won't `0` always be less than an absolute value?

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -258,6 +263,57 @@ TEST_F(GazeboRosWheelSlipTest, TestSetParameters)
...
+  ASSERT_GT(0.7, fabs(latestMsg->longitudinal_slip[0]));
```
scpeters: given our discussion of the sign convention for longitudinal slip, I recommend an expectation that doesn't use absolute value here. And I think we expect it to be positive for this test? If possible, `EXPECT_NEAR(0.3 /* or whatever */, latestMsg->longitudinal_slip[0], generous_tolerance);` would be nice.

side note: failing an `ASSERT_*` will immediately exit the test, so I only use that for things like checking for null pointers that I'm about to dereference. I recommend `EXPECT_*` for most other expectations

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
audrow: My intention here is to make sure that it's not 0, which may be the case if the wheel speed is below a certain tolerance. I think it may be better to use `EXPECT_NE` here.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
audrow: In the simulation, I believe the tricycle's front wheel gives a different sign for the longitudinal wheel slip than the rear wheels. This is why I use the absolute value. I assume this has to be fixed in the tricycle model, or maybe I can add another parameter for flipping the sign of the calculated slip.

The wheel slip publisher publishes information for a random set of the wheels with the plugin. Alternatively, I could check for the name of each wheel in the test and then test its slip with `EXPECT_NEAR`, if you think this would be better.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: I changed this to be closer to the equation in the pdf you linked above (`long_vel` is in the denominator) and used your suggested styling. The line now reads: `(spin_speed - long_vel) / long_vel`. (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/f1a0f88fc78ab539bb823e7aec1632b744cea4a6)

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
scpeters: the controller is only applying torque to the rear wheels, so that's why the slip values are different:

https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/files#diff-fd5ababb339b088418f3745d7340ba6b7cee97a16c5d34d9e8abbe3aa72e1461R48-R58

if you want to get a single wheel-slip message with all three wheels, you can unify the `gazebo_ros_wheel_slip` plugin instances (since there's currently two plugins). Then you could check the wheel name and set expectations accordingly.

Or, you can enable a velocity controller for the front wheel as well. I think there's a few options.

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
audrow: Actually, I think it makes more sense to have `spin_speed` in the denominator. For example, if the slip is perfect so that the wheel has no longitudinal velocity (i.e., within the tolerance), then the slip would be said to be zero, but it should be 100%. What do you think? It is a little different than the pdf's equation.

scpeters: there are multiple definitions of wheel slip. During acceleration and at low speeds, it's common to use `spin_speed` in the denominator. At high speeds and during braking, it's common to use `long_vel`. Elsewhere in WheelSlipPlugin we are using the version of slip with `spin_speed` in the denominator, so it may be more consistent to stick with `spin_speed` in the denominator

* https://github.com/osrf/gazebo/blob/gazebo11_11.9.1/plugins/WheelSlipPlugin.cc#L512-L537

scpeters: here's a link to the paper referenced in the WheelSlipPlugin comment:

http://www.astro.mech.tohoku.ac.jp/~yoshida/paperlist/ICRA02-yoshida.pdf

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
audrow: Makes sense, thank you.

gazebo_plugins/test/worlds/gazebo_ros_wheel_slip.world
```diff
@@ -0,0 +1,66 @@
...
+        </wheel>
+      </plugin>
+      <plugin name="joint_control" filename="libJointControlPlugin.so">
+        <controller type="position">
+          <joint>wheel_front_steer</joint>
+          <target>0</target>
+          <pid_gains>9 0 0.1</pid_gains>
+        </controller>
+        <controller type="velocity">
+          <joint>wheel_front</joint>
```
scpeters: I believe the proper joint name here is `wheel_front_spin`

https://github.com/osrf/gazebo_models/blob/bd7bc758fdf82be65b545d614d6a85359ff0e47c/trisphere_cycle/model.sdf#L279

gazebo_plugins/test/worlds/gazebo_ros_wheel_slip.world
```diff
@@ -0,0 +1,66 @@
...
+          <joint>wheel_rear_left_spin</joint>
+          <joint>wheel_rear_right_spin</joint>
+          <target>6.0</target>
+          <pid_gains>9 0 0</pid_gains>
+        </controller>
+        <controller type="force">
+          <joint>wheel_front</joint>
```
scpeters, line 51: `wheel_front_spin`

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
```diff
@@ -258,6 +263,57 @@ TEST_F(GazeboRosWheelSlipTest, TestSetParameters)
...
+  ASSERT_NE(nullptr, latestMsg);
+
+  ASSERT_LT(0u, latestMsg->name.size());
+  ASSERT_GT(4u, latestMsg->name.size());
+  ASSERT_EQ(latestMsg->name.size(), latestMsg->lateral_slip.size());
+  ASSERT_EQ(latestMsg->name.size(), latestMsg->longitudinal_slip.size());
+
+  EXPECT_NEAR(0.0, latestMsg->lateral_slip[0], tol);
+  EXPECT_NE(0.0, latestMsg->longitudinal_slip[0]);
+  EXPECT_GT(0.7, fabs(latestMsg->longitudinal_slip[0]));
```
scpeters: with those changes to the joint name, I'm seeing all three longitudinal wheel slips near `0.2`. How do you feel about replacing this `EXPECT_GT` with an `EXPECT_NEAR(0.2, ...`? You could repeat it for all 3 slip values as well

gazebo_plugins/test/worlds/gazebo_ros_wheel_slip.world
audrow: I'm not sure why but it seems to be `wheel_front` for me in my local testing. Otherwise, I don't see the front wheel in the publisher. `wheel_front` matches the demo world here: https://github.com/ros-simulation/gazebo_ros_pkgs/blob/f642be7a206cfecddf4aaf09c033562a201c6592/gazebo_plugins/worlds/gazebo_ros_wheel_slip_demo.world#L35.

gazebo_plugins/test/test_gazebo_ros_wheelslip.cpp
audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/6530b1fa1cbe8db13647e032167c2e00ef080758.

gazebo_plugins/test/worlds/gazebo_ros_wheel_slip.world
audrow: I see, good catch.

audrow: Fixed in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1331/commits/7608cef133764af35d83862c5b6cbf186ce31667.

scpeters (approved):
looks good!

thanks for patience with all my suggestions

please remember to edit the commit message when using squash-merge

scpeters:
> please remember to edit the commit message when using squash-merge

feel free to post it as a comment if you want a proofread before committing

audrow:
> feel free to post it as a comment if you want a proofread before committing

@scpeters, I've changed the title. What do you think? Also, how about the following for the commit body?

> This commit adds a publisher to the wheel slip plugin. The publisher publishes the lateral and longitudinal wheel slip for links that use the wheel slip plugin. The publication rate can be set through an SDF parameter. Also, this commit adds `WheelSlip.msg` to the `gazebo_msgs` package.

scpeters:
> @scpeters, I've changed the title. What do you think?

I made a slight edit by prefixing with `gazebo_ros_wheel_slip:`

> Also, how about the following for the commit body?
> 
> > This commit adds a publisher to the wheel slip plugin. The publisher publishes the lateral and longitudinal wheel slip for links that use the wheel slip plugin. The publication rate can be set through an SDF parameter. Also, this commit adds `WheelSlip.msg` to the `gazebo_msgs` package.

Thanks, how about the following:

~~~
The wheel slip plugin now publishes a WheelSlip message
containing the lateral and longitudinal wheel slip for each
wheel link that uses the plugin. The publication rate can be
configured through an SDFormat parameter. The new
`WheelSlip.msg` is added to `gazebo_msgs`.
~~~

Issue #1305, opened 2021-08-20
Ros2: Set Gazebo clock publish rate


HarunTeper:
Hello,

the default publish rate of the gazebo clock node is 10Hz. So far, it is most convenient to use the following approach to use Ros2 and Gazebo in a launch file for me:

gazebo_launch_cmd = IncludeLaunchDescription(
        PythonLaunchDescriptionSource(os.path.join(pkg_gazebo_ros, 'launch', 'gazebo.launch.py')),
        launch_arguments={'world': world}.items()
    )
However, looking at the launch files and some other threads, I can't manage to set the publish_rate for the gazebo clock node. Is it possible to include the parameter .yaml file in the launch arguments or to include a parameter {'publish_rate': '10'} to the list of parameters, so that it can be set by the launch file.

allenh1:
This is a duplicate of #1211 and #1247... Would be really great if this could be addressed. It makes the gazebo clock unusable for time-sensitive calculations.

john-maidbot:
I had this same issue. Pretty frustrating :sweat_smile: If anyone else finds this, there is a workaround (it would be a good solution if not for the error message, but it worked for me on ros2 galactic, gazebo version 11.9.0): 
https://answers.ros.org/question/378362/how-to-set-gazebo-clock-publish-rate-foxy/

allenh1:
@john-maidbot I have a pull request (see #1332) for dealing with this... It works for all my purposes, but would require you to build from source.

jacobperron:
https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1391 has made it more convenient to set ROS parameters (including `publish_rate`) from the launch file.

pinkponk:
For anyone else having problems with this and sees this issue, I finally got it to work with this:


```python
    gazebo_params_path = os.path.join(
                    get_package_share_directory(package_name),'config','gazebo_params.yaml')
    gazebo_world_path = os.path.join(
                    get_package_share_directory(package_name),'worlds','AAAA.world')

    gazebo = IncludeLaunchDescription(
                PythonLaunchDescriptionSource([os.path.join(
                    get_package_share_directory('gazebo_ros'), 'launch', 'gazebo.launch.py')]),
                    launch_arguments={
                        'params_file': gazebo_params_path,
                                      'world': gazebo_world_path }.items()
         )
```

Dont forget to include the config dir in the CMakeList.txt
```
install(
[...]

Issue #1247, opened 2021-03-06
ROS2 plugin Cannot set parameters (e.g. publish_rate)


joshnewans:
**Description**

The ROS2 plugin for Gazebo (`libgazebo_ros_init`?) does not seem to accept parameters, which is required to change the clock frequency.
There is an existing issue, #1211, discussing increasing the default clock frequency, and in that thread myself and another user (tim-fan) noted that regardless of the default, the `publish_rate` parameter cannot be set/changed despite [this guide](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-ROS-Clocks-and-sim-time#clock-rate) indicating that it should be possible.

**To Reproduce**

Follow the guide at https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-ROS-Clocks-and-sim-time#clock-rate
1. Create a YAML file with the appropriate parameters
2. Launch gazebo with the params file set `gazebo -s libgazebo_ros_init.so __params:=/tmp/params.yaml`
3. Check `ros2 param get /gazebo publish_rate` or `ros2 topic hz /clock`

**Expected behavior**

The `publish_rate` and `/clock` rate should match whatever is set in the params file. Instead it is always 10Hz.

**Environment (please complete the following information):**

 - OS: Foxy Docker (Ubuntu 20.04)
 - Gazebo version: Gazebo 11.3
 - ROS version: Foxy
  - gazebo_ros_pkgs: Debian install
 - glx server in use: NVIDIA Corporation

**Additional context**

This seems to be where the parameter is processed, I'm just not sure how to direct it there.
Also, as mentioned in those threads, this parameter should be exposed by the launch files, but I guess that can be a separate issue once this is resolved.

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/8a320668a5a6c84242547e13e04dbd5e1029c063/gazebo_ros/src/gazebo_ros_init.cpp#L167-L171

allenh1:
Is there any update on this? I'm stuck with this at the moment, and need to change this in a launch file to 500 Hz.

jacobperron:
[...]

PR #1332, opened 2021-11-11
[ROS 2] Allow Setting Publish Rate for /clock

allenh1:
This generates a params file with the desired clock rate, then passes the file to `gzserver` as the contents of `extra_gazebo_args`.

Signed-off-by: Hunter L. Allen <hunter.allen@ghostrobotics.io>

---

This adds `empy` as a dependency (which I need to add to the `package.xml`).

Other than that, this allows one to pass a publish rate from an external launch file, as requested in #1211, #1247, and #1305.


jacobperron:
Thanks for the PR! We ended up taking a slightly different approach in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1391

PR #1338, opened 2021-12-09
wheel_slip: set lateral slip to zero at low speed

scpeters:
The `gazebo_ros_wheel_slip` plugin publishes instantaneous wheel slip as of #1331. It currently truncates the longitudinal slip to `0` when the `spin_speed` is sufficiently small to avoid division by `0`. Lateral slip does not have a problem with dividing by `0` because it uses the `atan2` function, but it still generates significant slip values even when the vehicle is at rest. This updates the logic for the slip publisher to also truncate the lateral slip to `0` when the `spin_speed` is small. A test is added using a world file with a model at rest.

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
+++ b/gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
@@ -341,12 +341,14 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr 
         auto long_vel = slip.X() + slip.Z();
         auto lat_vel = slip.Y();
         double long_slip;
+        double lat_slip;
         if (fabs(spin_speed) < wheel_spin_tolerance) {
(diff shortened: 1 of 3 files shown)


gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -341,12 +341,14 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
         auto long_vel = slip.X() + slip.Z();
         auto lat_vel = slip.Y();
         double long_slip;
+        double lat_slip;
         if (fabs(spin_speed) < wheel_spin_tolerance) {
           long_slip = 0;
+          lat_slip = 0;
```
audrow, line 347: Good call to set the lateral slip here, too.

audrow (approved):
Looks like a good change to me!

PR #1349, opened 2022-01-28
Fix spawn_entity_demo: use executable

fmauch:
No big deal, but when I tried using the `spawn_entity_demo.launch.py` on my galactic system, it failed on me.

Since Foxy `node_executable` was marked [deprecated](https://github.com/ros2/launch_ros/blob/d2c726b5e8ce1aa99a9ae6c4881d2fa691ce0a64/launch_ros/launch_ros/actions/node.py#L140) and in Galactic it got removed. I assume the `ros2` branch should reflect that. This should also be applied to the `galactic` branch.


scpeters:
thanks for catching this!

PR #1355, opened 2022-02-06
gazebo_ros: fix cpplint and flake8 errors

scpeters:
There are lots of cpplint errors since a recent change in rolling. This fixes the new `build_order` errors in the `gazebo_ros` package. `gazebo_plugins` will be addressed in a separate pull request.

### DIFF
--- a/gazebo_ros/include/gazebo_ros/conversions/geometry_msgs.hpp
+++ b/gazebo_ros/include/gazebo_ros/conversions/geometry_msgs.hpp
@@ -15,15 +15,16 @@
 #ifndef GAZEBO_ROS__CONVERSIONS__GEOMETRY_MSGS_HPP_
 #define GAZEBO_ROS__CONVERSIONS__GEOMETRY_MSGS_HPP_
 
+#include <ignition/math/Pose3.hh>
+#include <ignition/math/Quaternion.hh>
+#include <ignition/math/Vector3.hh>
+
 #include <geometry_msgs/msg/point.hpp>
 #include <geometry_msgs/msg/point32.hpp>
 #include <geometry_msgs/msg/pose.hpp>
 #include <geometry_msgs/msg/quaternion.hpp>
 #include <geometry_msgs/msg/transform.hpp>
 #include <geometry_msgs/msg/vector3.hpp>
-#include <ignition/math/Pose3.hh>
-#include <ignition/math/Quaternion.hh>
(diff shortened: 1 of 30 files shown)


gazebo_ros/include/gazebo_ros/conversions/geometry_msgs.hpp
```diff
@@ -15,15 +15,16 @@
...
+
 #include <geometry_msgs/msg/point.hpp>
 #include <geometry_msgs/msg/point32.hpp>
 #include <geometry_msgs/msg/pose.hpp>
 #include <geometry_msgs/msg/quaternion.hpp>
 #include <geometry_msgs/msg/transform.hpp>
 #include <geometry_msgs/msg/vector3.hpp>
-#include <ignition/math/Pose3.hh>
-#include <ignition/math/Quaternion.hh>
-#include <ignition/math/Vector3.hh>
```
jacobperron, line 26: I guess cpplint thinks these are c-system headers? The other option is the use double-quotes for all includes not from the standard library. I don't have a preference either way though.

jacobperron (approved):
LGTM

gazebo_ros/include/gazebo_ros/conversions/geometry_msgs.hpp
scpeters: I guess the new version of `cpplint` thinks that `.hh` headers are C headers instead of C++

I can switch to double quotes if you think that's a better option

jacobperron: I don't have a preference; I don't think there's an objective reason to use double-quotes. If we use double-quotes, I think we can keep the `.hh` files _below_ any standard library headers :shrug:. The size of the diff might actually turn out to be larger if we switch to double-quotes.

chapulina:

jacobperron:
I've rebased the changes on `ros2` now that ROS CI is unblocked.

jacobperron:
In retrospect, I think cpplint is at fault for considering files with the .hh extension to be C headers instead of C++ headers. I've proposed a patch upstream: https://github.com/ament/ament_lint/pull/374

Note, even with the upstream patch, there are still a number of include order lint errors to address.

jacobperron:
flake8 fixes were merged in #1380

PR #1365, opened 2022-03-25
Initialize wheel slip parameters directly from SDF

jacobperron:
Do not rely on the parameter_events topic for initialization.
Otherwise, it's easy to race with other nodes that want to query or update the wheel slip parameters.

This commit changes the behavior of the parameters affecting all wheels.
These parameters are no longer initialized if a user does not provide a value, i.e. they are
'unset' until a value is set. Previously, they defaulted to the value set on an individual wheel.

For purposes of initializing values, parameters setting individual wheels now have higher precendence
than the parameters for all wheels.

I also bumped the number of retries (timeout) in the tests, since I noticed a larger delay with certain
RMWs (ie. rmw_connextdds).

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
@@ -77,7 +77,16 @@ class GazeboRosWheelSlipPrivate;
 ///      If not specified, these will default to 0.0.
 ///      Any negative values will ignored and set to 0.0.
 ///
-///  2. ROS parameters for each wheel, e.g :
+///  2. ROS parameters for all wheels, e.g :
+///      <ros>
+///         <parameter name="slip_compliance_unitless_lateral" type="double">0.1</parameter>
+///         <parameter name="slip_compliance_unitless_longitudinal" type="double">0.2</parameter>
+///      </ros>
+///
+///     If not specified, these will default to the last values set in SDF tags.
+///     If these are specified, they override SDF parameters for each wheel.
+///
+///  3. ROS parameters for each wheel, e.g :
 ///       <ros>
 ///         <parameter name="slip_compliance_unitless_lateral/wheel_front" type="double">0.1
 ///           </parameter>
@@ -85,20 +94,12 @@ class GazeboRosWheelSlipPrivate;
 ///           </parameter>
 ///       </ros>
 ///
-///      If not specified, these default to SDF parameters set for each wheel.
-///      If specified, they will override the values set as SDF parameters.
-///
-///  3. ROS parameters for all wheels, e.g :
-///      <ros>
-///         <parameter name="slip_compliance_unitless_lateral" type="double">0.1</parameter>
-///         <parameter name="slip_compliance_unitless_longitudinal" type="double">0.2</parameter>
-///      </ros>
-///
-///     If not specified, these will default to the last values set in SDF tags.
-///     If these are specified, they override both SDF and the ROS parameters for each wheel.
+///      If not specified, these default to the value of the ROS parameter for all wheels.
+///      If specified, they will override values set as SDF parameters or ROS parameters for all
+///      wheels.
 ///
 ///   Precedence order :
-///   ROS Params for all wheels > ROS params for individual wheels > SDF parameters
+///   ROS Params for individual wheels > ROS params for all wheels > SDF parameters
 ///   Check out the test cases for more information and expected behaviour.
 ///
 /// See the WheelSlipPlugin documentation at the following location for more details:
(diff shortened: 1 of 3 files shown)


jacobperron:
@jennuine if you could review this too, that would be great.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -77,28 +77,29 @@ class GazeboRosWheelSlipPrivate;
...
 ///
-///  2. ROS parameters for each wheel, e.g :
+///  2. ROS parameters for all wheels, e.g :
+///      <ros>
+///         <parameter name="slip_compliance_unitless_lateral" type="double">0.1</parameter>
+///         <parameter name="slip_compliance_unitless_longitudinal" type="double">0.2</parameter>
+///      </ros>
+///
+///     If not specified, these will default to the last values set in SDF tags.
+///     If these are specified, they override SDF parameters for each wheel.
```
adityapande-1995: *for all wheels

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
```diff
@@ -87,39 +103,54 @@ void GazeboRosWheelSlip::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
...
+        wheel_name);
+      this->impl_->map_slip_longitudinal_default_[wheel_name] = slip_longitudinal;
+    }
+  }
+
+  // Check for "global" slip parameters and use them for individual wheels if provided
+  impl_->ros_node_->declare_parameter(
+    "slip_compliance_unitless_lateral", rclcpp::PARAMETER_DOUBLE);
+  impl_->ros_node_->declare_parameter(
+    "slip_compliance_unitless_longitudinal", rclcpp::PARAMETER_DOUBLE);
```
adityapande-1995, line 125: Does this mean the parameter ``slip_compliance_unitless_longitudinal`` will have no default value ? Will the parameters return 0 if we do a "get value" on these parameters ?

adityapande-1995: In this PR : https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1308, we had explicit set their default values due to some weird behavior.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
```diff
@@ -77,28 +77,29 @@ class GazeboRosWheelSlipPrivate;
...
-///  3. ROS parameters for all wheels, e.g :
-///      <ros>
-///         <parameter name="slip_compliance_unitless_lateral" type="double">0.1</parameter>
-///         <parameter name="slip_compliance_unitless_longitudinal" type="double">0.2</parameter>
-///      </ros>
-///
-///     If not specified, these will default to the last values set in SDF tags.
-///     If these are specified, they override both SDF and the ROS parameters for each wheel.
+///      If not specified, these default to the value of the ROS parameter for all wheels.
+///      If specified, they will override values set as SDF parameters or ROS parameters for all
```
adityapande-1995: *for each wheel

gazebo_plugins/src/gazebo_ros_wheel_slip.cpp
jacobperron: The parameters will return "not set", which I think is reasonable if the user did not supply any values. 

I don't recall what the undesired behavior was in #1308, but this PR is certainly changing behavior so maybe the concern no longer applies?

gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp
jacobperron: I'm trying to communicate that these parameters for individual wheels will override parameters that apply to all wheels. The language is a bit confusing, maybe we can clarify the language.

jacobperron: I've attempted to clarify the documentation, PTAL: [aa06988](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1365/commits/aa0698835c18ff71a70226e86bc52f4503113f78)

jacobperron:

jennuine (approved):
LGTM

adityapande-1995:

jacobperron:
CI is broken for Rolling until a bug is resolved upstream: https://bugs.launchpad.net/ubuntu/+source/gazebo/+bug/1965780

adityapande-1995 (approved):
Tests look good !

PR #1367, opened 2022-03-31
gazebo_ros: Add inline keyword to template definitions

GerardHH:
Hello maintainers, 

I recently ran into the issue that I was referencing some of these template definitions in multiple translation units. This would result in linker errors. Marking these inline should prevent others to run into the same issue as me.

### DIFF
--- a/gazebo_ros/include/gazebo_ros/conversions/builtin_interfaces.hpp
+++ b/gazebo_ros/include/gazebo_ros/conversions/builtin_interfaces.hpp
@@ -27,7 +27,8 @@ namespace gazebo_ros
 /// \brief Specialized conversion from an Gazebo Time to a ROS Time message.
 /// \param[in] in Gazebo Time to convert.
 /// \return A ROS Time message with the same value as in
-template<> inline
+template<>
+inline
 builtin_interfaces::msg::Time Convert(const gazebo::common::Time & in)
 {
   builtin_interfaces::msg::Time time;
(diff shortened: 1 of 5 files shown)


JRBeltman (approved):
this follows the standard.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf2-a-h-file-must-not-contain-object-definitions-or-non-inline-function-definitions
https://timsong-cpp.github.io/cppwp/n4659/temp.expl.spec#12

gazebo_ros/include/gazebo_ros/conversions/builtin_interfaces.hpp
```diff
@@ -27,7 +27,7 @@ namespace gazebo_ros
 /// \brief Specialized conversion from an Gazebo Time to a ROS Time message.
 /// \param[in] in Gazebo Time to convert.
 /// \return A ROS Time message with the same value as in
-template<>
+template<> inline
```
jacobperron: Looks like our linters prefer the `inline` keyword on the next line, e.g. 
```suggestion
template<>
inline
```

Same for all other instances.

jacobperron (commented):
Thanks for the fix! Besides the linter issue, LGTM

gazebo_ros/include/gazebo_ros/conversions/builtin_interfaces.hpp
jacobperron: I went ahead and pushed a commit to address this ([7a79eff](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1367/commits/7a79eff53a95fa43209b8cf3f31cdc5c6a485e82)).

wjwwood:
Another way to do this (maybe better) would have been to move the implementation into a `.cpp` file.

JvE-Lely:
When will this be released as a version? Since in `v3.7.0` this patch is not included.

scpeters:
> When will this be released as a version? Since in `v3.7.0` this patch is not included.

It has been ported to `rolling` in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1388, but is not yet released, as you have noted

PR #1371, opened 2022-04-24
gazebo_plugins: GPS sensor plugin publishing velocity

MarcelDudek:
Since Foxy supports Gazebo 11 by default, GPS sensor implements VelocityEast, VelocityNorth and VelocityUp methods
that allow for velocity simulation. Based on previous commits this was not implemented in Gazebo 9.0 and therefore velocity support was not implemented for GPS plugin.

Added publisher for velocity data simulated by sensor. It outputs data as `geometry_msgs::mgs::Vector3Stamped`. `NavFix` unfortunately doesn't support velocity data and there is no suitable message in `sensors_msgs`.

Velocity data is in ENU coordinates.

velocity_output


jacobperron (approved):
LGTM, thanks for the addition!

PR #1376, opened 2022-04-28
Fixes gazebo shutdown error when using nested launch files

adityapande-1995:
Resolves https://github.com/ros2/launch/issues/545

Originally, ``gzserver`` and ``gzclient`` were launched using ``shell=True``. 

This means the launch process executes a shell, which executes the server and client. When launch_testing exits, it sends SIGINT to the children shell processes, which do not forward SIGINT to the respective server and client children. These processes remain alive even after launch exits, as children of ``systemd``. 

Check out https://linuxconfig.org/how-to-propagate-a-signal-to-child-processes-from-a-bash-script for reference.

## Testing
The following test launch file should exit properly and kill both the client and server processes : 
```
import sys
import time
import unittest

import launch
from launch.actions import IncludeLaunchDescription
from launch.launch_description_sources import PythonLaunchDescriptionSource
from launch_ros.substitutions import FindPackageShare
from launch_testing.actions import ReadyToTest
import launch_testing.markers
import pytest


@pytest.mark.launch_test
def generate_test_description():
    return launch.LaunchDescription(
        [
            IncludeLaunchDescription(
                PythonLaunchDescriptionSource([FindPackageShare('gazebo_ros'), '/launch/gazebo.launch.py'])
            ),
            ReadyToTest(),
        ])


class MyTestFixture(unittest.TestCase):

    def test_something(self):
        time.sleep(10.0)
        print('END TEST', file=sys.stderr)
```


ivanpauno (approved):
LGTM!
I'm not sure if there was any reason to use `shell=True`.

adityapande-1995:

jacobperron (commented):
When I try the launch a complex Gazebo world with these changes, it fails to load plugins:

```
[gzserver-1] [Err] [Plugin.hh:212] Failed to load plugin  libgazebo_ros_init.so:  libgazebo_ros_init.so: cannot open shared object file: No such file or directory
[gzserver-1] [Err] [Plugin.hh:212] Failed to load plugin  libgazebo_ros_factory.so:  libgazebo_ros_factory.so: cannot open shared object file: No such file or directory
[gzserver-1] [Err] [Plugin.hh:212] Failed to load plugin  libgazebo_ros_force_system.so:  libgazebo_ros_force_system.so: cannot open shared object file: No such file or directory
```
I guess that the changes are somehow affecting command line argument parsing?

jacobperron:
Actually, the errors occur in an empty world too. You just need to pass `verbose:=true` to see the errors:

```
ros2 launch gazebo_ros gazebo.launch.py verbose:=true
```

adityapande-1995:
Hmm, I'm not sure why that is happening, as the environment variables are included here in ``gzserver.launch.py`` : 
```
    env = {
        'GAZEBO_MODEL_PATH': model,
        'GAZEBO_PLUGIN_PATH': plugin,
        'GAZEBO_RESOURCE_PATH': media
    }
```

jacobperron:
I'm not sure either. The difference between `shell=True` and `shell=False` is that the former calls [subprocess_shell](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_shell) and the later calls [subprocess_exec](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_exec), both are documented to treat the `env` argument the same as [subprocess.Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen).

jacobperron (approved):
LGTM

adityapande-1995:
Not sure what is happening with CI though

jacobperron:
We're working on fixing CI issues (e.g. https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1380). Please stand by.

jacobperron:
Rebased on `ros2`

gazebo_ros/launch/gzserver.launch.py
```diff
@@ -219,6 +219,6 @@ def _arg_command(arg):
 
 # Add gazebo_ros plugins if true
 def _plugin_command(arg):
-    cmd = ['"-s libgazebo_ros_', arg, '.so" if "true" == "', LaunchConfiguration(arg), '" else ""']
+    cmd = ['"-s', 'libgazebo_ros_', arg, '.so" if "true" == "', LaunchConfiguration(arg), '" else ""']
```
jacobperron: flake8 is complaining that this line is too long.

adityapande-1995:
The errors in CI are unrelated to this PR : 
```
 ALSA lib confmisc.c:855:(parse_card) cannot find card '0'
10: ALSA lib conf.c:5178:(_snd_config_evaluate) function snd_func_card_inum returned error: No such file or directory
10: ALSA lib confmisc.c:422:(snd_func_concat) error evaluating strings
10: ALSA lib conf.c:5178:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
10: ALSA lib confmisc.c:1334:(snd_func_refer) error evaluating name
10: ALSA lib conf.c:5178:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
10: ALSA lib conf.c:5701:(snd_config_expand) Evaluate error: No such file or directory
10: ALSA lib pcm.c:2664:(snd_pcm_open_noupdate) Unknown PCM default
10: AL lib: (EE) ALCplaybackAlsa_open: Could not open playback device 'default': No such file or directory
```

jacobperron (approved):

ivanpauno:
Are we sure the issue commented [here](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1376#pullrequestreview-959608416) was not related to the PR?
It would important to double check before backporting, as that could cause many issues.

adityapande-1995:
Yeah it was caused due to earlier changes in this PR, fixed in this commit :  https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1376/commits/e5b5676b1bdb85d16eed38e1133ee4d20b6c3ce1 , the space character was causing the issue.

Arne-Christoph:
Hey all!
If I use these changes and try to change the publish frequency of the /clock topic via the extra_gazebo_args parameter

e.g.
"extra_gazebo_args",
 default_value="--ros-args --params-file gazebo_params.yaml",

this is no longer working.
If I change the shell parameter back to true it is working again. Does someone has a clue what is going wrong?

@adityapande-1995 @ivanpauno have you noticed something like this?

jacobperron:
@Arne-Christoph For visibility, it would be better to open up an issue with the exact commands that used to work, describing the previous and new behavior.

Arne-Christoph:
alright. will do so.

PR #1380, opened 2022-05-03
fix: modify some tests

wep21:
modify some header include orders and change some brackets to double quotes to pass the cpplint test.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_init.cpp
+++ b/gazebo_ros/src/gazebo_ros_init.cpp
@@ -162,7 +162,7 @@ void GazeboRosInit::Load(int argc, char ** argv)
   // the simulation is paused), late subscribers can receive the previously published message(s).
   impl_->clock_pub_ = impl_->ros_node_->create_publisher<rosgraph_msgs::msg::Clock>(
     "/clock",
-    rclcpp::QoS(rclcpp::KeepLast(10)).transient_local());
+    rclcpp::ClockQoS());
 
 #ifdef GAZEBO_ROS_HAS_PERFORMANCE_METRICS
   impl_->performance_metrics_pub_ =
--- a/gazebo_ros/test/test_sim_time.cpp
+++ b/gazebo_ros/test/test_sim_time.cpp
@@ -60,12 +60,10 @@ TEST_F(TestSimTime, TestClock)
   executor.add_node(node);
 
   std::vector<ClockMsg::SharedPtr> msgs;
-  std::vector<rclcpp::Time> times;
   auto sub = node->create_subscription<ClockMsg>(
     "/clock", rclcpp::ClockQoS(),
     [&](ClockMsg::SharedPtr _msg) {
       msgs.push_back(_msg);
-      times.push_back(node->now());
     });
 
   unsigned int sleep{0};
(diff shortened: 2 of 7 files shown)


wep21:
@jacobperron @j-rivero Could you review this PR?

scpeters:
so the `.hh` headers need to be enclosed in `""` instead of `<>`?

jacobperron:
This overlaps with https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1355

Also, I'm proposing a change to ament_cpplint to fix a bug where .hh files are considered C system headers (which may affect this PR): https://github.com/ament/ament_lint/pull/374

jacobperron:
> so the .hh headers need to be enclosed in "" instead of <>?

I think this is only a workaround, to trick cpplint into thinking they are not C system headers. After https://github.com/ament/ament_lint/pull/374, using double-quotes shouldn't be necessary.

wep21:
How about  once ignoring cpplint include order until https://github.com/ament/ament_lint/pull/374 is merged?

jacobperron:
> How about once ignoring cpplint include order until https://github.com/ament/ament_lint/pull/374 is merged?

Sounds good to me. Thanks for working on it!

gazebo_ros/test/test_sim_time.cpp
```diff
@@ -62,7 +62,7 @@ TEST_F(TestSimTime, TestClock)
   std::vector<ClockMsg::SharedPtr> msgs;
   std::vector<rclcpp::Time> times;
   auto sub = node->create_subscription<ClockMsg>(
-    "/clock", rclcpp::SystemDefaultsQoS(),
+    "/clock", rclcpp::ClockQoS(),
```
jacobperron, line 64: Maybe we should also update the publisher QoS to use the same QoS:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/9954bb9f5c02e463e1227aa3f756247e8c6756df/gazebo_ros/src/gazebo_ros_init.cpp#L163-L165

gazebo_ros/test/test_sim_time.cpp
```diff
@@ -88,7 +88,7 @@ TEST_F(TestSimTime, TestClock)
       EXPECT_GT(rclcpp::Time(msgs[i]->clock), rclcpp::Time(msgs[i - 1]->clock)) <<
         rclcpp::Time(msgs[i]->clock).nanoseconds() << "    " <<
         rclcpp::Time(msgs[i - 1]->clock).nanoseconds();
-      EXPECT_GT(times[i], times[i - 1]) << times[i].nanoseconds() << "    " <<
+      EXPECT_GE(times[i], times[i - 1]) << times[i].nanoseconds() << "    " <<
```
wep21: I don't know how to ensure that test clock callback is executed after rclcpp internal clock callback, so I changed the condition of sim time test.

gazebo_ros/test/test_sim_time.cpp
wep21: @jacobperron I fixed publisher QoS at https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1380/commits/261df1f52c2cf17231bc9dc34fd8050591e1d1ef.

wep21:
@ros-pull-request-builder

jacobperron:

gazebo_ros/test/test_sim_time.cpp
jacobperron: I'm not sure the test is well-formed since it depends on the order in which the executor processes callbacks. It might be better to just assert that messages are received on the clock topic instead of also trying to test that the node's use_sim_time logic is working (seems out of scope for this test). So, I'd be okay with just removing this check.

jacobperron (approved):

jacobperron:
Looks like there are some, unrelated, flaky tests. I'll try triggering CI one more time.

jacobperron:

gazebo_ros/test/test_sim_time.cpp
wep21: @jacobperron I removed the test of use_sim_time logic at https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1380/commits/1864f19f02fae83dfbfc777bb8e14cf5570ec4dd.

jacobperron (approved):

gazebo_ros/src/gazebo_ros_init.cpp
```diff
@@ -162,7 +162,7 @@ void GazeboRosInit::Load(int argc, char ** argv)
   // the simulation is paused), late subscribers can receive the previously published message(s).
   impl_->clock_pub_ = impl_->ros_node_->create_publisher<rosgraph_msgs::msg::Clock>(
     "/clock",
-    rclcpp::QoS(rclcpp::KeepLast(10)).transient_local());
+    rclcpp::ClockQoS());
```
jacobperron, line 165: I'm having second thoughts about this change; I didn't realize the durability setting for `rclcpp::ClockQoS` is volatile (instead of transient local). I think it might be better to revert this change so we keep the same behavior for anyone depending on the behavior from early versions.

PR #1391, opened 2022-05-17
Expose launch parameter for `params_file`

### LINKED ISSUE #1211: Increase /clock frequency

### LINKED ISSUE #1247: ROS2 plugin Cannot set parameters (e.g. publish_rate)

### LINKED ISSUE #1305: Ros2: Set Gazebo clock publish rate

ihasdapie:
This PR exposes a `params_file` launch argument for `gazebo` with which the clock rate (or any other ROS parameters) can be set.

For example:

``` yaml
# params.yml
gazebo:
    ros__parameters:
        publish_rate: 400.0
```

`ros2 launch gazebo_ros gazebo.launch.py params_file:=params.yml`

See [wiki](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration%3A-ROS-Clocks-and-sim-time) for more

- Closes #1305 
- Closes #1211 
- Closes #1247

### DIFF
--- a/gazebo_ros/launch/gzserver.launch.py
+++ b/gazebo_ros/launch/gzserver.launch.py
@@ -53,12 +53,12 @@ def generate_launch_description():
         _plugin_command('force_system'), ' ',
         # Wait for (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/941)
         # _plugin_command('force_system'), ' ',
-        _arg_command('profile'), ' ', LaunchConfiguration('profile'), ' ', 
+        _arg_command('profile'), ' ', LaunchConfiguration('profile'), ' ',
         # convenience parameter for params file
         '--ros-args',
         PythonExpression([
             '"--params-file', '" if "" != "',
-            LaunchConfiguration("params_file"), '" else ""' ]),
+            LaunchConfiguration("params_file"), '" else ""']),
         LaunchConfiguration("params_file"),
         '--',
         LaunchConfiguration('extra_gazebo_args'),
@@ -158,7 +158,6 @@ def generate_launch_description():
             'extra_gazebo_args', default_value='',
             description='Extra arguments to be passed to Gazebo'
         ),
-        
         DeclareLaunchArgument(
             'params_file', default_value='',
             description='Path to ROS 2 yaml parameter file'
--- a/gazebo_ros/launch/gzserver.launch.py
+++ b/gazebo_ros/launch/gzserver.launch.py
@@ -55,11 +55,12 @@ def generate_launch_description():
         # _plugin_command('force_system'), ' ',
         _arg_command('profile'), ' ', LaunchConfiguration('profile'), ' ', 
         # convenience parameter for params file
+        '--ros-args',
         PythonExpression([
-            '"--ros-args', ' --params-file', '" if "" != "',
+            '"--params-file', '" if "" != "',
             LaunchConfiguration("params_file"), '" else ""' ]),
-        LaunchConfiguration("params_file"), ' ',
-
+        LaunchConfiguration("params_file"),
+        '--',
         LaunchConfiguration('extra_gazebo_args'),
     ]
 
@@ -160,7 +161,7 @@ def generate_launch_description():
         
         DeclareLaunchArgument(
             'params_file', default_value='',
-            description='Path to ROS2 yaml parameter file'
+            description='Path to ROS 2 yaml parameter file'
         ),
 
         # Specific to gazebo_ros
--- a/gazebo_ros/launch/gzserver.launch.py
+++ b/gazebo_ros/launch/gzserver.launch.py
@@ -53,7 +53,13 @@ def generate_launch_description():
         _plugin_command('force_system'), ' ',
         # Wait for (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/941)
         # _plugin_command('force_system'), ' ',
-        _arg_command('profile'), ' ', LaunchConfiguration('profile'),
+        _arg_command('profile'), ' ', LaunchConfiguration('profile'), ' ', 
+        # convenience parameter for params file
+        PythonExpression([
+            '"--ros-args', ' --params-file', '" if "" != "',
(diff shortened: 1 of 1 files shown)


gazebo_ros/launch/gzserver.launch.py
```diff
@@ -151,6 +157,11 @@ def generate_launch_description():
             'extra_gazebo_args', default_value='',
             description='Extra arguments to be passed to Gazebo'
         ),
+        
+        DeclareLaunchArgument(
+            'params_file', default_value='',
+            description='Path to ROS2 yaml parameter file'
```
jacobperron:
```suggestion
            description='Path to ROS 2 yaml parameter file'
```

gazebo_ros/launch/gzserver.launch.py
```diff
@@ -217,6 +228,7 @@ def _arg_command(arg):
     return py_cmd
 
 
+
```
jacobperron: nitpick: remove this extra newline

gazebo_ros_pkgs.repos
```diff
@@ -3,10 +3,6 @@ repositories:
     type: git
     url: https://github.com/ros-simulation/gazebo_ros_pkgs.git
     version: ros2
-  image_common:
-    type: git
-    url: https://github.com/ros-perception/image_common.git
-    version: ros2
```
jacobperron: I think it would be better to make this change in a separate pull request.

gazebo_ros/launch/gzserver.launch.py
```diff
@@ -53,7 +53,13 @@ def generate_launch_description():
...
         # Wait for (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/941)
         # _plugin_command('force_system'), ' ',
-        _arg_command('profile'), ' ', LaunchConfiguration('profile'),
+        _arg_command('profile'), ' ', LaunchConfiguration('profile'), ' ', 
+        # convenience parameter for params file
+        PythonExpression([
+            '"--ros-args', ' --params-file', '" if "" != "',
+            LaunchConfiguration("params_file"), '" else ""' ]),
+        LaunchConfiguration("params_file"), ' ',
+
```
jacobperron: I think we better terminate `--ros-args` with `--`, otherwise if we launch with `extra_gazebo_args` set, it may be confused for ROS arguments.

Also, it looks like if we drop the extra spaces this works without setting `shell=True` on `ExecuteProcess`.

```suggestion
        '--ros-args',
        PythonExpression([
            '"--params-file', '" if "" != "',
            LaunchConfiguration("params_file"), '" else ""' ]),
        LaunchConfiguration("params_file"),
        '--',
```

jacobperron (commented):
Thanks! I've left some feedback below.

jacobperron (approved):
LGTM!

jacobperron:
@osrf-jenkins run tests

jacobperron:
The warnings from ros2_gazebo_pkgs-ci-pr job look unrelated.
Same for the test failure in the Hpr job.

I think we can merge this :)

PR #1394, opened 2022-05-20
Multiple nodes with same name issue

### LINKED ISSUE #1303: Avoid creating multiple nodes with the same name
We have a specific use-case where we create a custom plugin that inherits from `GazeboRosCameraPlugin`. This child class needs a reference to a ROS node, but if we use `gazebo_ros::Node::Get(sdf)` then we end up with two ROS nodes with the same name: the node created by the parent and the node created by the child. Both are using the SDF to determine the name of the node.

I think in general, if multiple gazeob_ros plugins happen to use the same name in SDF, then we will end up with multiple nodes with the same name. This is generally discouraged in ROS 2 and can result in ambiguous scenarios, e.g. when setting parameters remotely, they will be arbitrarily set on one of the nodes with the same name.

To workaround this issue, we introduced a getter for the node so that children can access it (instead of creating a new node): https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1299. However, this does not fix the problem; it only provides a workaround.

To properly support this use case we should either, 1) forbid creating multiple nodes with the same name or 2) make it easy to avoid creating multiple nodes with the same name.

If we go with option (2), we could achieve this by maintaining a global map of node names to node objects (alongside the other globals in `gazebo_ros::Node`). Then whenever a user tries to get a new node, we can first check if one already exists with the same name and give them a handle to it.  However, this method may run into thread-safety issues.

deepanshubansal01:
The `gazebo_ros_plugins` while instantiating the rclcpp::Node object uses the node name as  the plugin name specified in sdf/world file. However, this can create problems in cases we have multiple plugins with same name. 

This PR forbids creating nodes with same name by keeping track of global node names. In case a given name exists already, a rclcpp error is shown and the gazebo is shutdown safely.  

The full description of the original issue can be found here #1303

### DIFF
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -27,6 +27,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <unordered_set>
 
 namespace gazebo_ros
 {
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -87,9 +87,11 @@ public:
    * </plugin>
    * \endcode
    * \param[in] _sdf An SDF element in the style above or containing a <ros> tag in the style above
-   * \return A shared pointer to a new #gazebo_ros::Node
+   * \param[in] _node_name: An optional node_name to overwrite plugin name being used as node name.
+   * \return A shared pointer to a new #gazebo_ros::Node. A nullptr is returned in case multiple
+   * nodes have the same name resulting in a ROS error and gazebo crash.
    */
-  static SharedPtr Get(sdf::ElementPtr _sdf);
+  static SharedPtr Get(sdf::ElementPtr _sdf, std::string _node_name="");
 
   /// Create a #gazebo_ros::Node and add it to the global #gazebo_ros::Executor.
   /**
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -88,8 +88,8 @@ public:
    * \endcode
    * \param[in] _sdf An SDF element in the style above or containing a <ros> tag in the style above
    * \param[in] _node_name: An optional node_name to overwrite plugin name being used as node name.
-   * \return A shared pointer to a new #gazebo_ros::Node. A nullptr is returned in case multiple
-   * nodes have the same name resulting in a ROS error and gazebo crash.
+   * \return A shared pointer to a new #gazebo_ros::Node. A nullptr is returned and an error message
+   * is logged in case multiple nodes have the same name.
    */
   static SharedPtr Get(sdf::ElementPtr _sdf, std::string _node_name = "");
 
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -30,6 +30,9 @@
 
 namespace gazebo_ros
 {
+
+class NodeLookUp;
+
 /// ROS Node for gazebo plugins
 /**
  * \class Node node.hpp <gazebo_ros/node.hpp>
@@ -140,6 +143,9 @@ private:
   /// QoS for node entities
   gazebo_ros::QoS qos_;
 
+  /// ros node lookup
+  static NodeLookUp static_node_lookup_;
+
   /// Locks #initialized_ and #executor_
   static std::mutex lock_;
 
@@ -217,5 +223,20 @@ Node::SharedPtr Node::CreateWithArgs(Args && ... args)
 
   return node;
 }
+
+// Class to hold the global map of gazebo_ros::Node::SharedPtr objects
+class NodeLookUp{
+   public:
+    // Methods need to be protected by internal mutex
+    void add_node(const std::string& node_name, Node::SharedPtr ros_node);
+    Node::SharedPtr get_node(const std::string& node_name);
+    void remove_node(const std::string& node_name);
+
+ private:
+    /// map of node names vs objects
+    std::unordered_map<std::string, Node::SharedPtr> map_;
+    std::mutex internal_mutex_;
+};
+
 }  // namespace gazebo_ros
 #endif  // GAZEBO_ROS__NODE_HPP_
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -31,8 +31,8 @@
 
 namespace gazebo_ros
 {
-// forward declare NodeLookUp
-class NodeLookUp;
+// forward declare ExistingNodes
+class ExistingNodes;
 
 /// ROS Node for gazebo plugins
 /**
@@ -145,7 +145,7 @@ private:
   gazebo_ros::QoS qos_;
 
   /// ros node lookup
-  static NodeLookUp static_node_lookup_;
+  static ExistingNodes static_node_lookup_;
 
   /// Locks #initialized_ and #executor_
   static std::mutex lock_;
@@ -226,12 +226,12 @@ Node::SharedPtr Node::CreateWithArgs(Args && ... args)
 }
 
 // Class to hold the global set of tracked node names.
-class NodeLookUp
+class ExistingNodes
 {
 public:
   // Methods need to be protected by internal mutex
   void add_node(const std::string & node_name);
-  bool is_node_name_in_set(const std::string & node_name);
+  bool check_node(const std::string & node_name);
   void remove_node(const std::string & node_name);
 
 private:
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -225,17 +225,19 @@ Node::SharedPtr Node::CreateWithArgs(Args && ... args)
 }
 
 // Class to hold the global set of tracked node names.
-class NodeLookUp{
-   public:
-    // Methods need to be protected by internal mutex
-    void add_node(const std::string& node_name);
-    bool is_node_name_in_set(const std::string& node_name);
-    void remove_node(const std::string& node_name);
+class NodeLookUp
+{
+public:
+  // Methods need to be protected by internal mutex
+  void add_node(const std::string & node_name);
+  bool is_node_name_in_set(const std::string & node_name);
+  void remove_node(const std::string & node_name);
 
- private:
-    /// set of tracked node names
-    std::unordered_set<std::string> set_;
-    std::mutex internal_mutex_;
+private:
+  /// set of tracked node names
+  std::unordered_set<std::string> set_;
+  std::mutex internal_mutex_;
+
 };
 
 }  // namespace gazebo_ros
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -239,6 +239,5 @@ private:
   std::unordered_set<std::string> set_;
   std::mutex internal_mutex_;
 };
-
 }  // namespace gazebo_ros
 #endif  // GAZEBO_ROS__NODE_HPP_
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -238,7 +238,6 @@ private:
   /// set of tracked node names
   std::unordered_set<std::string> set_;
   std::mutex internal_mutex_;
-
 };
 
 }  // namespace gazebo_ros
--- a/gazebo_ros/include/gazebo_ros/node.hpp
+++ b/gazebo_ros/include/gazebo_ros/node.hpp
@@ -30,7 +30,7 @@
 
 namespace gazebo_ros
 {
-
+// forward declare NodeLookUp
 class NodeLookUp;
 
 /// ROS Node for gazebo plugins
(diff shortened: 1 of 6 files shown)


gazebo_ros/src/node.cpp
```diff
@@ -164,4 +183,24 @@ rclcpp::Logger Node::internal_logger()
   return rclcpp::get_logger("gazebo_ros_node");
 }
 
+void NodeLookUp::add_node(const std::string& node_name, Node::SharedPtr ros_node) {
+  std::lock_guard<std::mutex> guard(this->internal_mutex_);
+  this->map_.emplace(node_name, ros_node);
+}
+
+Node::SharedPtr NodeLookUp::get_node(const std::string& node_name) {
```
adityapande-1995: Now that I think about it, we don't require a ``get_node`` method in this container, as we are not going to return a node with the name name if that exists already. In fact, we can change the container to just a ``std::vector`` instead of a ``map``

adityapande-1995 (changes requested):
Good work !
Could you add a short description in the PR ? You could write a short explanation of the problem, the solution that this PR provides, and a link to the issue that was relevant to this PR. We should also mention that this is a behavior breaking change (when you have multiple plugins with same name).

It would be nice to add a test as well, you can look at this file : https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_ros/test/test_node.cpp, which tests partial features of the ``Node`` class.

gazebo_ros/src/node.cpp
deepanshubansal01: Thanks, makes sense. 
Instead of `vector` maybe we can have `unordered_set` I guess then? Probably will be more efficient for search but we can discuss that later as well.

adityapande-1995 (commented):
The CI also check for linting errors, formatting, and style. You can run ``ament_uncrustify some_file.cpp`` to check those errors locally. For instance, here it suggests changing the indent : https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-pr_any_rolling-jammy-amd64/27/testReport/junit/gazebo_ros/uncrustify/include_gazebo_ros_node_hpp/

gazebo_ros/src/node.cpp
adityapande-1995: Sound good to me, that would be more efficient.

adityapande-1995:
These tests have duplicate nodes for some reason, we should resolve these : 

[projectroot.test.test_gazebo_ros_p3d](https://build.ros2.org/job/Rpr__gazebo_ros_pkgs__ubuntu_jammy_amd64/51/testReport/junit/projectroot/test/test_gazebo_ros_p3d/)
[projectroot.test.test_gazebo_ros_force](https://build.ros2.org/job/Rpr__gazebo_ros_pkgs__ubuntu_jammy_amd64/51/testReport/junit/projectroot/test/test_gazebo_ros_force/)
[projectroot.test.test_gazebo_ros_wheelslip](https://build.ros2.org/job/Rpr__gazebo_ros_pkgs__ubuntu_jammy_amd64/51/testReport/junit/projectroot/test/test_gazebo_ros_wheelslip/)

adityapande-1995 (commented):
We should add ``#include <unordered_set>`` at the start. The general policy is, "Include what you use"

gazebo_ros/test/test_node_lookup.cpp
```diff
@@ -0,0 +1,53 @@
...
+    "<sdf version='1.6'>"
+    "<world name='default'>"
+    "<model name='model_1'>"
+    "<plugin name='node_1' filename='libnode_name.so'/>"
+    "</model>"
+    "<model name='model_2'>"
+    "<plugin name='node_1' filename='libnode_name.so'/>"
+    "</model>"
+    "</world>"
+    "</sdf>";
```
adityapande-1995: It would be nice to indent this sdf file : 
```
<world name='default'>
  <sdf version='1.6'>
...
```

gazebo_ros/test/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ set(tests
   test_gazebo_ros_properties
   test_gazebo_ros_state
   test_node
+  test_node_lookup
```
adityapande-1995: Maybe we can call this ``test_reject_duplicate_nodes``

gazebo_ros/src/node.cpp
```diff
@@ -164,4 +183,22 @@ rclcpp::Logger Node::internal_logger()
...
+{
+  std::lock_guard<std::mutex> guard(this->internal_mutex_);
+  this->set_.erase(node_name);
+}
+
+bool NodeLookUp::is_node_name_in_set(const std::string & node_name)
+{
+  std::lock_guard<std::mutex> guard(this->internal_mutex_);
+  return this->set_.find(node_name) != this->set_.end();
+}
```
adityapande-1995: I'd prefer the class name to be more descriptive, say ``ExistingNodes``, and the ``is_node_name_in_set`` could be called ``check_node``.

gazebo_ros/src/node.cpp
```diff
@@ -97,6 +102,17 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
...
   }
+  // set full node name
+  full_name = ns + "/" + name;
+
+  // check if node with the same name exists already
+  if (static_node_lookup_.is_node_name_in_set(full_name)) {
+    RCLCPP_ERROR(
+      internal_logger(),
+      "Found multiple nodes with same name: %s. This is due to different plugins with same name, "
+      "either change the plugin name or use a unique namespace", full_name.c_str());
```
adityapande-1995: This *might* be due to nodes with same name, as cases like inheritance can also cause this error, we might want to include that in this message

adityapande-1995 (commented):
Looking good !

gazebo_ros/test/CMakeLists.txt
deepanshubansal01: yeah, makes sense.

gazebo_ros/src/node.cpp
```diff
@@ -98,13 +103,35 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
...
+  } else {
+    full_name = ns + "/" + name;
+  }
+
+  // check if node with the same name exists already
+  if (static_existing_nodes_.check_node(full_name)) {
+    RCLCPP_ERROR(
+      internal_logger(),
+      "Found multiple nodes with same name: %s. This might be due to multiple plugins using same "
+      "name to solve this either change one of the the plugin names or use a different namespace. "
```
jacobperron:
```suggestion
      "name. Try changing one of the the plugin names or use a different ROS namespace. "
```

gazebo_ros/src/node.cpp
```diff
@@ -98,13 +103,35 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
...
+      "The error might also result from a custom plugin inheriting from one of the GazeboRosPlugin "
```
jacobperron:
```suggestion
      "This error might also result from a custom plugin inheriting from another gazebo_ros plugin. "
```

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -217,5 +224,20 @@ Node::SharedPtr Node::CreateWithArgs(Args && ... args)
 
   return node;
 }
+
+// Class to hold the global set of tracked node names.
+class ExistingNodes
```
jacobperron, line 231: Since this is more of an implementation detail, can this definition live in the cpp file?

gazebo_ros/src/node.cpp
```diff
@@ -98,13 +103,35 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
...
+      "this can be solved by accessing node object of the parent class itself instead of creating "
+      "a new node object for custom plugin.", full_name.c_str());
```
jacobperron: My concern here is that we don't provide a way for the user to get the ROS node from the parent class (at least I don't think we do).  So I'm not sure that we should suggest this.

Perhaps, the best thing we can do is make it convenient for plugins to create nodes with their own name. E.g. Change the signature of the function to `Get(sdf::ElementPtr sdf, std::string node_name = "")` (or provide an overload) where `node_name` use used instead of the plugin name.

gazebo_ros/src/node.cpp
```diff
@@ -98,13 +103,35 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
...
+    return nullptr;  // this makes the gazebo shutdown safely
```
jacobperron: I guess this is okay for plugins that check for `nullptr`. We should document this new behavior in the docs for this function.

gazebo_ros/include/gazebo_ros/node.hpp
deepanshubansal01: yeah, I will move the definition to cpp.

deepanshubansal01: The definition for this is already in cpp file should I move the class declaration as well to cpp?

gazebo_ros/src/node.cpp
deepanshubansal01: Where should I document this.

gazebo_ros/src/node.cpp
deepanshubansal01: I am moving forward with having an optional argument `Get(sdf::ElementPtr sdf, std::string node_name = "")` which will override the plugin name in case the default node name is provided.

gazebo_ros/include/gazebo_ros/node.hpp
jacobperron: I think technically the methods are defined in the cpp file, but the class is still being defined here. I was referring to moving this class definition to the cpp file, though maybe that's not possible since we are defining a member of this type above :thinking:

gazebo_ros/src/node.cpp
jacobperron: Here: https://github.com/ros-simulation/gazebo_ros_pkgs/blob/851babe02ab5d9d758c9f129ef273bad71e5ea9c/gazebo_ros/include/gazebo_ros/node.hpp#L90

gazebo_ros/include/gazebo_ros/node.hpp
deepanshubansal01: I think write now the class declaration is in header and definition in cpp.

gazebo_ros/src/node.cpp
deepanshubansal01: I have added this now.

gazebo_ros/src/node.cpp
deepanshubansal01: @jacobperron @adityapande-1995 I have changed the ROS ERROR message being displayed according to the new optional node name argument but might need some minor modifications.

gazebo_ros/src/node.cpp
```diff
@@ -98,13 +108,36 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
...
+    full_name = ns + name;
+  } else {
+    full_name = ns + "/" + name;
+  }
+
+  // check if node with the same name exists already
+  if (static_existing_nodes_.check_node(full_name)) {
+    RCLCPP_ERROR(
+      internal_logger(),
+      "Found multiple nodes with same name: %s. This might be due to multiple plugins using same "
```
jacobperron:
```suggestion
      "Found multiple nodes with same name: %s. This might be due to multiple plugins using the same "
```

gazebo_ros/src/node.cpp
```diff
@@ -98,13 +108,36 @@ Node::SharedPtr Node::Get(sdf::ElementPtr sdf)
...
+      "name. Try changing one of the the plugin names or use a different ROS namespace. "
+      "This error might also result from a custom plugin inheriting from another gazebo_ros plugin "
+      "and the custom plugin trying to access the ROS node object hence creating multiple nodes "
+      "with same name. To solve this try providing the optional node_name argument in "
+      "gazebo_ros::Node::Get() function. ", full_name.c_str());
+    return nullptr;  // this makes the gazebo shutdown safely
```
jacobperron: nitpick: technically this doesn't cause gazebo to shutdown, it is up to the plugin to check for nullptr and fail gracefully.
```suggestion
    return nullptr;
```

gazebo_ros/include/gazebo_ros/node.hpp
```diff
@@ -83,9 +87,11 @@ class GAZEBO_ROS_NODE_PUBLIC Node : public rclcpp::Node
    * </plugin>
    * \endcode
    * \param[in] _sdf An SDF element in the style above or containing a <ros> tag in the style above
-   * \return A shared pointer to a new #gazebo_ros::Node
+   * \param[in] _node_name: An optional node_name to overwrite plugin name being used as node name.
+   * \return A shared pointer to a new #gazebo_ros::Node. A nullptr is returned in case multiple
+   * nodes have the same name resulting in a ROS error and gazebo crash.
```
jacobperron: nitpick: Correct me if I'm wrong, but this doesn't actually cause Gazebo to crash; it depends on how the plugin handles the nullptr.

```suggestion
   * \return A shared pointer to a new #gazebo_ros::Node.
   * A nullptr is returned and an error message is logged in case multiple nodes have the same name.
```

jacobperron (commented):
Thanks for adding tests :)

I have a couple minor comments left.

gazebo_ros/include/gazebo_ros/node.hpp
deepanshubansal01: yeah, make sense.

deepanshubansal01: From what I observed from testing is that the gazebo doesn't even started in the first place and the overall process exited cleanly.

deepanshubansal01:
> Thanks for adding tests :)
> 
> I have a couple minor comments left.

No problem. Thanks a lot for the review!

jacobperron (approved):
LGTM!

PR #1396, opened 2022-05-23
remove image_common from gazebo_ros_pkgs.repos

ihasdapie:
`image_common` is already provided in [`ros2.repos`](https://github.com/ros2/ros2/blob/077c53b9a27100bcb4e3ec94549d5a9e6690fea5/ros2.repos#L66-L69)


jacobperron (approved):
LGTM

I think the ros2_gazebo_pkgs and Hpr jobs are failing for reasons unrelated to this PR.

PR #1401, opened 2022-06-09
feat: make cmd topic configurable

mart-karelics:
This MR add the feature to set the `command_topic` for the `gazebo_ros_planar_move` plugin

It can now be used like the other paramters:
```xml
    <gazebo>
        <plugin name="object_controller" filename="libgazebo_ros_planar_move.so">
            <command_topic>/cmd_vel</command_topic>
            <odometry_frame>odom</odometry_frame>
            <publish_rate>20.0</publish_rate>
            <update_rate>20.0</update_rate>
            <publish_odom>True</publish_odom>
            <publish_odom_tf>False</publish_odom_tf>
            <robot_base_frame>base_link</robot_base_frame>
            <covariance_x>0.00001</covariance_x>
            <covariance_y>0.00001</covariance_y>
            <covariance_yaw>0.00001</covariance_yaw>
        </plugin>
    </gazebo>
```

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_planar_move.cpp
+++ b/gazebo_plugins/src/gazebo_ros_planar_move.cpp
@@ -150,17 +150,19 @@ void GazeboRosPlanarMove::Load(gazebo::physics::ModelPtr _model, sdf::ElementPtr
   }
   impl_->last_update_time_ = impl_->world_->SimTime();
 
-  // Update rate
+  // Publish rate
   auto publish_rate = _sdf->Get<double>("publish_rate", 20.0).first;
-  if (update_rate > 0.0) {
+  if (publish_rate > 0.0) {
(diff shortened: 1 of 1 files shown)


mart-karelics:
@jacobperron Anything I can do speed up merging this?

jacobperron (commented):
Sorry for the delayed review.

Can we instead just remap the topic name instead of adding another sdf tag?
For example,

```xml
<plugin name="object_controller" filename="libgazebo_ros_planar_move.so">
  <ros>
    <remapping>cmd_vel:=new_topic</remapping>
  </ros>
</plugin>
```

I haven't confirmed if the above example works, but I think this is exactly the use-case remapping is intended for.

mart-karelics:
Thanks for the reply, i have confirmed the above example works if I add an `/` before the `cmd_vel`

PR #1407, opened 2022-07-06
Set camera optical centre from gazebo directly instead of SDF tag

deepanshubansal01:
This PR makes make changes to set the camera optical centre from gazebo directly instead of taking optical centre(cx, cy) values from SDF tag. 

Signed-off-by: deepanshu <deepanshubansal01@gmail.com>

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera.cpp
@@ -405,7 +405,7 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
       impl_->camera_[i]->LensDistortion()->SetCrop(border_crop);
 
       cx = impl_->camera_[i]->LensDistortion()->Center().X();
-      cy =  impl_->camera_[i]->LensDistortion()->Center().Y();
+      cy = impl_->camera_[i]->LensDistortion()->Center().Y();
 
       distortion_k1 = impl_->camera_[i]->LensDistortion()->K1();
       distortion_k2 = impl_->camera_[i]->LensDistortion()->K2();
--- a/gazebo_plugins/src/gazebo_ros_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera.cpp
@@ -392,8 +392,8 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
     auto hack_baseline = _sdf->Get<double>("hack_baseline", 0.0).first;
 
     // Get Optical center from camera
-    double cx = (static_cast<double>(width[i]) + 1.0) / 2.0;
-    double cy = (static_cast<double>(height[i]) + 1.0) / 2.0;
+    double default_cx = (static_cast<double>(width[i]) + 1.0) / 2.0;
+    double default_cy = (static_cast<double>(height[i]) + 1.0) / 2.0;
 
     // Get distortion from camera
     double distortion_k1{0.0};
@@ -404,8 +404,8 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
     if (impl_->camera_[i]->LensDistortion()) {
       impl_->camera_[i]->LensDistortion()->SetCrop(border_crop);
 
-      cx = impl_->camera_[i]->LensDistortion()->Center().X();
-      cy = impl_->camera_[i]->LensDistortion()->Center().Y();
+      default_cx = impl_->camera_[i]->LensDistortion()->Center().X();
+      default_cy = impl_->camera_[i]->LensDistortion()->Center().Y();
 
       distortion_k1 = impl_->camera_[i]->LensDistortion()->K1();
       distortion_k2 = impl_->camera_[i]->LensDistortion()->K2();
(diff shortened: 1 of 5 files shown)


gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -407,6 +404,9 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
     if (impl_->camera_[i]->LensDistortion()) {
       impl_->camera_[i]->LensDistortion()->SetCrop(border_crop);
 
+      cx = impl_->camera_[i]->LensDistortion()->Center().X();
+      cy =  impl_->camera_[i]->LensDistortion()->Center().Y();
```
adityapande-1995: Uncrustify doesn't seem happy with this line

adityapande-1995: Could we add a test for this patch ?

adityapande-1995 (commented):
I'm not sure of the deprecation strategy here, but it might be helpful to at least throw a warning or a debug statement saying that the sdf parameter will be ignored and instead the optical centre would be taken from Gazebo.

deepanshubansal01:
> I'm not sure of the deprecation strategy here, but it might be helpful to at least throw a warning or a debug statement saying that the sdf parameter will be ignored and instead the optical centre would be taken from Gazebo.

Makes sense.

adityapande-1995 (commented):
Based on the offline discussion, we should change the behavior so that the default value is taken from gazebo, and if the sdf tags are provided that should override the default values. This way we don't need to deprecate the tags. Also, test cases would be nice.

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -398,6 +391,9 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
     auto border_crop = _sdf->Get<bool>("border_crop", true).first;
     auto hack_baseline = _sdf->Get<double>("hack_baseline", 0.0).first;
 
+    double default_cx = (static_cast<double>(width[i]) + 1.0) / 2.0;
+    double default_cy = (static_cast<double>(height[i]) + 1.0) / 2.0;
```
jacobperron, line 395: Do we need to compute defaults from the image size if we're just overriding these from `LensDistortion()->Center()` below?

deepanshubansal01: I checked the behaviour and we do need to compute default `cx, cy` from image size reason being the `LensDistortion()->Center()` thing only works if their is a `distortion` tag provided in world file. 

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/612692e3c6a00d662b5bf02461e80b928e7dc119/gazebo_plugins/src/gazebo_ros_camera.cpp#L407

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -407,13 +403,21 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
     if (impl_->camera_[i]->LensDistortion()) {
       impl_->camera_[i]->LensDistortion()->SetCrop(border_crop);
 
+      // Get Optical center from camera
+      default_cx = impl_->camera_[i]->LensDistortion()->Center().X();
+      default_cy = impl_->camera_[i]->LensDistortion()->Center().Y();
```
deepanshubansal01, line 408: Based on unit testing it looks like the way we are getting the` cx, cy` from lens distortion is also wrong. I guess the `center` tag of `distortion`  tag takes values from **0 to 1** and we need to multiple by corresponding height and width to get `cx, cy.`

Also, if no `center` tag is provided in distortion tag it defaults to **0.5**.

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/612692e3c6a00d662b5bf02461e80b928e7dc119/gazebo_plugins/test/worlds/gazebo_ros_camera_distortion_barrel.world#L37

deepanshubansal01: cc @adityapande-1995 @jacobperron

jacobperron: Is there another SDF element that we can check? [This ticket](https://github.com/gazebosim/sdformat/issues/1071) implies that there is a `<camera><lens><intrinsics><cx>` element that is also valid. What do the intrinsic camera center values default to?

deepanshubansal01: Based on testing it seems like there isn't any other SDF element which we can use. There is a `ProjectionMatrix()` in gazebo which we can probably use to get some information. However, the `ProjectionMatrix` is calculated in a weird **normalised** way from OGRE. In case when no `<camera><lens><intrinsics><cx>` is provided the gazebo calculates the projection matrix using other parameters(image size, focal length and fov). 

Here is where calculation for projection matrix is done in case **when no intrinsic are provided.**
https://github.com/OGRECave/ogre/blob/36c8fe6ae25fc36d039dc542669809e365cb21fb/OgreMain/src/OgreFrustum.cpp#L325-L336

The OGRE uses `makeProjectionMatrix()` to compute the **normalised projection matrix**. 
https://ogrecave.github.io/ogre/api/13/class_ogre_1_1_math.html#a23c7798ef5a11dea62318b5be2983c2d
https://github.com/OGRECave/ogre/blob/36c8fe6ae25fc36d039dc542669809e365cb21fb/OgreMain/src/OgreFrustum.cpp#L369

deepanshubansal01:
Closing this is favour of new PR see https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1410

PR #1409, opened 2022-07-20
gazebo_ros_camera: configure projection matrix from sdf

ihasdapie:
This PR allows for the projection matrix in `gazebo_ros_camera` to be configured from SDF.

Signed-off-by: Brian Chen <brian.chen@openrobotics.org>

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
@@ -59,9 +59,15 @@ class GazeboRosCameraPrivate;
       <triggered>true</triggered>
 
       <!-- Set some projection matrix fields-->
+      <!-- Projection matrix principal point cx-->
       <P_cx>0</P_cx>
+      <!-- Projection matrix principal point cy-->
       <P_cy>320.5</P_cy>
+      <!-- Projection matrix focal length fy-->
+      <P_fy>0</P_fy>
+      <!-- Projection matrix translation Tx, Ty between stereo cameras-->
       <Tx>240.5</Tx>
+      <Ty>0</Ty>
 
       <hack_baseline>0.07</hack_baseline>
     </plugin>
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
@@ -58,6 +58,11 @@ class GazeboRosCameraPrivate;
       <!-- Set to true to turn on triggering -->
       <triggered>true</triggered>
 
+      <!-- Set some projection matrix fields-->
+      <P_cx>0</P_cx>
+      <P_cy>320.5</P_cy>
+      <Tx>240.5</Tx>
+
       <hack_baseline>0.07</hack_baseline>
     </plugin>
   \endcode
--- a/gazebo_plugins/src/gazebo_ros_camera.cpp
+++ b/gazebo_plugins/src/gazebo_ros_camera.cpp
@@ -447,14 +447,14 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
 
     // camera_ projection matrix (same as camera_ matrix due
     // to lack of distortion/rectification) (is this generated?)
-    camera_info_msg.p[0] = focal_length;
+    camera_info_msg.p[0] = _sdf->Get<double>("fx_p", focal_length).first;
     camera_info_msg.p[1] = 0.0;
-    camera_info_msg.p[2] = cx;
-    camera_info_msg.p[3] = -focal_length * hack_baseline;
+    camera_info_msg.p[2] = _sdf->Get<double>("cx'", cx).first;
+    camera_info_msg.p[3] = _sdf->Get<double>("Tx", -focal_length * hack_baseline).first;
     camera_info_msg.p[4] = 0.0;
-    camera_info_msg.p[5] = focal_length;
-    camera_info_msg.p[6] = cy;
-    camera_info_msg.p[7] = 0.0;
+    camera_info_msg.p[5] = _sdf->Get<double>("fy_p", focal_length).first;
+    camera_info_msg.p[6] = _sdf->Get<double>("cy'", cy).first;
+    camera_info_msg.p[7] = _sdf->Get<double>("Ty", 0).first;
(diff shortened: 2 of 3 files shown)


gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -447,14 +447,14 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
 
     // camera_ projection matrix (same as camera_ matrix due
     // to lack of distortion/rectification) (is this generated?)
-    camera_info_msg.p[0] = focal_length;
+    camera_info_msg.p[0] = _sdf->Get<double>("fx_p", focal_length).first;
     camera_info_msg.p[1] = 0.0;
-    camera_info_msg.p[2] = cx;
-    camera_info_msg.p[3] = -focal_length * hack_baseline;
+    camera_info_msg.p[2] = _sdf->Get<double>("cx'", cx).first;
```
jacobperron: Is a single-quote character valid in SDF? I'm trying to imagine what setting this in practice looks like,

```xml
<cx'>100</cx'>
```

I would make the names clear that they're specifically for overriding the projection matrix, P, e.g. `P_cx`, `P_cy`, `P_fx`, `P_fy`, etc.

jacobperron:
For posterity, the idea behind this PR is to allow users to configure the projection matrix, P, independently of the intrinsics matrix, K. I think this makes sense in the case of stereo cameras, where P may be different from K due to the transform between the two cameras.

gazebo_plugins/src/gazebo_ros_camera.cpp
ihasdapie: Yup that'd be a good idea. I've pushed changes to reflect that.

gazebo_plugins/src/gazebo_ros_camera.cpp
```diff
@@ -447,13 +447,13 @@ void GazeboRosCamera::Load(gazebo::sensors::SensorPtr _sensor, sdf::ElementPtr _
 
     // camera_ projection matrix (same as camera_ matrix due
```
jacobperron: We're also interested in exposing the R matrix parameters above. I'm not sure what is the best way to set a matrix from SDF.

ihasdapie: It appears that in gazebo classic and gazebo matrices are set elementwise with a new sdf tag for each settable entry. 
Unless there's a better way we could just do this.

```suggestion
    camera_info_msg.r[0] = _sdf->Get<double>("r_0", 1.0).first;
    camera_info_msg.r[1] = _sdf->Get(double>("r_1", 0.0).first
    // ... and so forth
```

jacobperron: Maybe @adityapande-1995 or @scpeters can recommend an approach.

ihasdapie: I _think_ we could also do something like this:
`<rectification_matrix>1 0 0 0 1 0 0 0 1</rectification_matrix>`

And while we're at it make the other matricies use this syntax as well.

scpeters: > I _think_ we could also do something like this: `<rectification_matrix>1 0 0 0 1 0 0 0 1</rectification_matrix>`
> 
> And while we're at it make the other matricies use this syntax as well.

yes, I think this would be easier than specifying all the matrix parameters individually, as long as we document the order of parameters. I recommend this approach

scpeters: should we block this until we add the `rectification_matrix` or merge as is?

ihasdapie: I think it is reasonable to merge as-is and then we can add `rectification_matrix` and the other matricies in a new PR

scpeters:
this looks good to me; we just need to add documentation of the SDFormat parameters to [gazebo_ros_camera.hpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp) and example usage in [gazebo_ros_camera_demo.world](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/worlds/gazebo_ros_camera_demo.world)

ihasdapie:
> this looks good to me; we just need to add documentation of the SDFormat parameters to [gazebo_ros_camera.hpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp) and example usage in [gazebo_ros_camera_demo.world](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/worlds/gazebo_ros_camera_demo.world)

@scpeters Resolved in  https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1409/commits/1470786adc393e771f9ae60c4907d81d51cf5599. 

The values I put in for the the example usage and in the demo world are the same as before i.e. the default calculated values using the existing tags so no behaviour changes.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
```diff
@@ -58,6 +58,11 @@ class GazeboRosCameraPrivate;
       <!-- Set to true to turn on triggering -->
       <triggered>true</triggered>
 
+      <!-- Set some projection matrix fields-->
+      <P_cx>0</P_cx>
+      <P_cy>320.5</P_cy>
+      <Tx>240.5</Tx>
```
scpeters, line 69: can you also add `Ty` and `P_fy` and a brief description of each parameter?

gazebo_plugins/src/gazebo_ros_camera.cpp
kjeppesen1: Here is the `rectification_matrix` PR: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1423

gazebo_plugins/include/gazebo_plugins/gazebo_ros_camera.hpp
scpeters: 24cf9b721557b046227bbe015d35ecca1bf79b7b, thanks

scpeters (approved):

PR #1414, opened 2022-08-04
Fix console errors and warnings when launching gzserver

### LINKED ISSUE #1411: Launch parameter errors in gzserver.launch.py
I'm seeing some weird console errors using the `ros2` branch with the following launch command:

~~~
$ ros2 launch gazebo_ros gzserver.launch.py verbose:=true physics:=ode
...
[Err] [Server.cc:472] Could not open file[]
[Wrn] [Server.cc:381] Falling back on worlds/empty.world
...
[Err] [Server.cc:530] Unregistered physics engine [ ], the default will be used instead
~~~

jacobperron:
Fixes #1411 

Since we are no longer launching with `shell=True`, we do not need the extra spaces between command arguments (d369be816b0f7d21b44591d163871ba81a0dab90).
In particular, they were causing an error to be printed to the console if the `physics` option was given.

If one of the substitutions in the gzserver command results in an empty string,
then we see the following console error and warning:

    [Err] [Server.cc:472] Could not open file[]
    [Wrn] [Server.cc:381] Falling back on worlds/empty.world

This is because gzserver is parsing any empty string as the world argument.
To fix the issue, we explicitly launch an empty world if no launch argument for the
world is given. ([b617a29](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1414/commits/b617a290f88a124e5ba0bca1da1a642273d3fd06))


jacobperron:
hmm... looks like this breaks the params_file launch argument. I'm looking into it.

adityapande-1995 (commented):
Can we have a small test case for this, using say ``physics:=ode`` ?

jacobperron:
Yes, good idea. I think we can add a launch test checking for expected output.

jacobperron:
> hmm... looks like this breaks the params_file launch argument. I'm looking into it.

I realized that the main issue was with the 'world' argument. So, instead of trying to filter out empty strings in the argument list, instead I've just copied the empty world file from Gazebo to launch explicitly by default. This avoids gzserver from incorrectly parsing another argument as the world file.

jacobperron:
@adityapande-1995 I've added a test that checks the launch output. PTAL ([3eab7c1](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1414/commits/3eab7c10106724fe12b4d9d30802c276dfc2a271)).

I don't think there's a convenient way to check the output of an included launch file with `launch_pytest`, so we should probably add something upstream to make this more ergonomic.

jacobperron:
I've created a launch issue for adding better facilities for checking process output: https://github.com/ros2/launch/issues/653

wjwwood (approved):

scpeters:
LGTM

PR #1424, opened 2022-08-26
Add Proximity sensor option

tonynajjar:
Title says it all

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
+++ b/gazebo_plugins/src/gazebo_ros_ray_sensor.cpp
@@ -76,6 +76,9 @@ public:
   /// brief Radiation type to report when output type is range
   uint8_t range_radiation_type_;
 
+  /// If set, the range sensor becomes a proximity sensor and this its threshold.
+  double proximity_threshold_{0.0};
+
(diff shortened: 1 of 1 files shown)


iliabaranov:
Does the existing range sensor not work?

https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/Range.html

It has specific definition for fixed distance ranging, which is essentially a proximity sensor

tonynajjar:
What you linked is the Range message. Can you link me to the range sensor?

iliabaranov:
Sure thing :)

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/231a7219b36b8a6cdd100b59f66a3df2955df787/gazebo_plugins/src/gazebo_ros_range.cpp#L286

tonynajjar:
Thanks. It's not available in ROS2, it has been replaced by the ray sensor, see [here](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Ray-sensors#gazebo_ros_range).

I didn't look in too much detail, but it seems to me like even the ROS1 range sensor does not support thresholding it to simulate a proximity sensor

Issue #1239, opened 2021-02-20
[ROS2] Ackermann Drive Plugin fails silently if the wheels have gemetries other than SphereShape or CylinderShape


SivertHavso:
**Description**

`gazebo_ros_ackermann_drive` should warn users if the wheel radius is an invalid value.

The plugin finds the radius by calling GetRadius() on sphere and cylinder objects but this fails silently if the wheel geometry is a mesh.

**Screenshots**
The relevant parts of the code is:
```cpp
  impl_->wheel_radius_ = impl_->CollisionRadius(
    impl_->joints_[GazeboRosAckermannDrivePrivate::REAR_RIGHT]->GetChild()->GetCollision(id));
```

```cpp
double GazeboRosAckermannDrivePrivate::CollisionRadius(const gazebo::physics::CollisionPtr & _coll)
{
  if (!_coll || !(_coll->GetShape())) {
    return 0;
  }
  if (_coll->GetShape()->HasType(gazebo::physics::Base::CYLINDER_SHAPE)) {
    gazebo::physics::CylinderShape * cyl =
      dynamic_cast<gazebo::physics::CylinderShape *>(_coll->GetShape().get());
    return cyl->GetRadius();
  } else if (_coll->GetShape()->HasType(gazebo::physics::Base::SPHERE_SHAPE)) {
    gazebo::physics::SphereShape * sph =
      dynamic_cast<gazebo::physics::SphereShape *>(_coll->GetShape().get());
    return sph->GetRadius();
  }
  return 0;
}
```
The return value from the CollisionRadius() method is never checked.

**Expected behavior**

The limitation should be documented, and the plugin should output an error if `impl_->wheel_radius_` is 0 or negative.

**Additional context**
I'll attempt to make a pull request to fix it soon

Yadunund:
@SivertHavso I have opened this [PR](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1427) to address the problem.

PR #1427, opened 2022-09-14
[ROS 2] Fix how wheel properties are retrieved in Ackermann Plugin

Yadunund:
This PR fixes https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1239 by retrieving all required geometric parameters from SDF params as opposed to calculating them from meshes with certain implicit assumptions of mesh geometry. The parameter name `track_width` was chosen instead of `wheel_separation` as this usually nomenclature for automobiles which the `ackermann` plugin will most likely be used with.

The current implementation has implicit assumptions of what the collision mesh of the wheel should be but this is not apparent to users who may spend a lot of time figuring out why their gazebo model is not responding as required. 

Other plugins in this repo like the [diff_drive](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/0440b0ec23c8b4254d17b83f87e7e9dc0eb7b602/gazebo_plugins/include/gazebo_plugins/gazebo_ros_diff_drive.hpp#L72-L74) also gets the parameters from the SDF and not via meshes.

Signed-off-by: Yadunund <yadunund@gmail.com>

### DIFF
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
@@ -40,6 +40,11 @@ class GazeboRosAckermannDrivePrivate;
       <update_rate>100.0</update_rate>
 
       <!-- wheels -->
+      <wheel_radius>0.3</front_left_joint>
+      <wheel_base>2.0</wheel_base>
+      <track_width>1.2</track_width>
+
+      <!-- joints -->
       <front_left_joint>front_left_wheel_joint</front_left_joint>
       <front_right_joint>front_right_wheel_joint</front_right_joint>
       <rear_left_joint>rear_left_wheel_joint</rear_left_joint>
--- a/gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
+++ b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
@@ -39,12 +39,12 @@ class GazeboRosAckermannDrivePrivate;
 
       <update_rate>100.0</update_rate>
 
-      <!-- wheels -->
+      <!-- kinematics -->
       <wheel_radius>0.3</front_left_joint>
       <wheel_base>2.0</wheel_base>
       <track_width>1.2</track_width>
 
-      <!-- joints -->
+      <!-- wheels -->
       <front_left_joint>front_left_wheel_joint</front_left_joint>
       <front_right_joint>front_right_wheel_joint</front_right_joint>
       <rear_left_joint>rear_left_wheel_joint</rear_left_joint>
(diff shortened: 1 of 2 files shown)


gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
```diff
@@ -39,6 +39,11 @@ class GazeboRosAckermannDrivePrivate;
 
       <update_rate>100.0</update_rate>
 
+      <!-- kinematics -->
+      <wheel_radius>0.3</front_left_joint>
+      <wheel_base>2.0</wheel_base>
+      <track_width>1.2</track_width>
```
j-rivero, line 56: Could we please include a comment to indicate that this means "wheel_seperation" in ackermann language? Should help people to understand what really is just from reading.

j-rivero (commented):
> The current implementation has implicit assumptions of what the collision mesh of the wheel should be but this is not apparent to users who may spend a lot of time figuring out why their gazebo model is not responding as required.

I agree with this and also think that @siverhavso is right on issue #1239 about a total silent in the presence of errors. The PR looks to me like quite disrupting with the current behavior supported (getting info from meshes) so a more conservative approaches could be:
 1. Check if kinematic information is provided in the SDF (current pr code) and if not fallback to the old behavior and implement an error message if both fail.
 1. Try using the current approach of getting info from meshes and fallback to this PR code if GetRadius fails.
 1. Implement a ROS param to tell the plugin if someone wants the current approach or the new one in this PR.

 Not an strong opinion but going with 3 defaulting with current code could help us to keep compatibility and consistency with diff_drive.

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
Yadunund: Done [7410afe](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1427/commits/7410afe37274be91661275230afbfe1d6dd630cb)

Yadunund:
Sorry for the long overdue changes @j-rivero.

I refactored the implementation to default to the existing behavior of inferring properties from the collision mesh file. If the radius computation fails, an error is printed to inform the users that we only support collision meshes that are cylinders or spheres. Alternatively, they can set `wheel_params_from_collision_mesh` param to `false` and manually specify the values that the plugin needs.

cschindlbeck:
What is left to do here?

gazebo_plugins/include/gazebo_plugins/gazebo_ros_ackermann_drive.hpp
```diff
@@ -39,6 +39,22 @@ class GazeboRosAckermannDrivePrivate;
...
+
+      <!-- By default this plugin will attempt to infer the wheel radius,
+      wheelbase and track width from the geometry and poses of the links defined
+      as front and rear wheels. This only works when the collision meshes are
+      cylinder or sphere primitives. If the collision mesh for the model
+      needs to be more complex, you must set this param to false and specific
+      the values of the wheel params below. -->
+      <wheel_params_from_collision_mesh>true</wheel_params_from_collision_mesh>
+      <!-- The radius of the wheel in meters. Used when wheel_params_from_collision_mesh is false. -->
+      <wheel_radius>0.3</front_left_joint>
```
cschindlbeck, line 52: This should be

```suggestion
      <wheel_radius>0.3</wheel_radius>
```

cschindlbeck (commented):
In the hpp the tag wheel_radius is not correct, see below

PR #1428, opened 2022-09-26
gazebo_ros: rename python package

Aposhian:
Currently `gazebo_ros` installs a Python package called "scripts", which is terribly generic and collides easily with other packages. This renames it to a python package called `gazebo_ros`. All usages in this repo are updated, but if there are external users of this then it will break them.


Aposhian:
Does `gazebo_ros_paths.py` even need to be installed as a "program" though?

PR #1429, opened 2022-10-05
Fix deprecation warning in spawn_entity.py

jacobperron:


audrow (approved):

jacobperron:
Looks like we need to fix some other c++ deprecation warnings as well.

scpeters:
looks like some failing tests in `Rpr__gazebo_ros_pkgs__ubuntu_jammy_amd64` that are different each time

scpeters (approved):
CI failures are unrelated

PR #1453, opened 2023-02-02
gazebo_ros_factory: more reliable model spawning

scpeters:
Use the [World::InsertModelString](https://github.com/gazebosim/gazebo-classic/blob/gazebo11_11.12.0/gazebo/physics/World.cc#L2558-L2565) API instead of the gazebo-transport `~/factory` topic to spawn models. This is more reliable, since gazebo transport may lose messages. You can test this with the `spawn_entity.py` script.

For reference, the `~/factory` callback ([World::OnFactoryMsg](https://github.com/gazebosim/gazebo-classic/blob/gazebo11_11.12.0/gazebo/physics/World.cc#L1587-L1592)) receives a `gazebo::msgs::Factory` and pushes it to the World's `dataPtr->factoryMsgs` data structure to be processed in a separate thread.

The [World::InsertModelString](https://github.com/gazebosim/gazebo-classic/blob/gazebo11_11.12.0/gazebo/physics/World.cc#L2558-L2565) receives a `std::string`, wraps it in a `gazebo::msgs::Factory` and pushes the message to the `dataPtr->factoryMsgs` data structure. So the C++ API has the same data endpoint but copies the data directly instead of using gazebo transport. This c++ API can be used in the `gazebo_ros_factory` plugin because it is loaded and running in the `gzserver` process.

### DIFF
--- a/gazebo_ros/src/gazebo_ros_factory.cpp
+++ b/gazebo_ros/src/gazebo_ros_factory.cpp
@@ -97,9 +97,6 @@ public:
   /// Gazebo node for communication.
   gazebo::transport::NodePtr gz_node_;
 
-  /// Publishes model factory messages.
-  gazebo::transport::PublisherPtr gz_factory_pub_;
-
   /// Publishes light factory messages.
   gazebo::transport::PublisherPtr gz_factory_light_pub_;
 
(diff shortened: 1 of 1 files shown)


audrow (approved):
LGTM!

PR #1456, opened 2023-03-02
Refactor gzserver launch args, add --initial_sim_time

scpeters:
This refactors some of the argument handling in `gzserver.launch.py` and adds support for the `--initial_sim_time` parameter, which was recently added in https://github.com/gazebosim/gazebo-classic/pull/3294. It is straightforward to review the changes by commit:

* https://github.com/ros-simulation/gazebo_ros_pkgs/commit/00c0c1e1249a4ab0f7a13ee8d6bd549f3a429a5e: Move the functionality of the `_arg_command` helper with a `condition` to a separate `_conditional_command` helper function and update the `params_file` support to use `_conditional_command`.
* https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1456/commits/efa494f3285bd8516b1e32af925a72559e5232a7: Remove the `condition` argument from `_arg_command` and add a `join_with` argument instead that defaults to a space `' '`. Also, since all the remaining uses of `_arg_command` also pass the corresponding `LaunchConfiguration()` right afterward, change the return value of `_arg_command` to a list joining the `--command` string with the `LaunchConfiguration()` value. This simplifies the specification of the `cmd` variable in the launch description.
* https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1456/commits/c6f90bcb365d63286fa4fa05fea6e5fdb89c2f03: Add support for the `--initial_sim_time` argument, using `join_with='='` so that it will be passed as `--initial_sim_time=[time]`. This is a valid alternative to `--initial_sim_time [time]` but allows old versions of gazebo to parse it all as a single argument and ignore it as unrecognized. Otherwise the `--initial_sim_time` may be ignored, but the `[time]` argument may linger and cause unintended parsing issues.

### DIFF
--- a/gazebo_ros/launch/gzserver.launch.py
+++ b/gazebo_ros/launch/gzserver.launch.py
@@ -55,11 +55,16 @@ def generate_launch_description():
         # Wait for (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/941)
         # _plugin_command('force_system'), ' ',
         _arg_command('profile'), LaunchConfiguration('profile'),
-        # convenience parameter for params file
-        _arg_command('ros-args', condition=LaunchConfiguration('params_file')),
-        _arg_command('params-file', condition=LaunchConfiguration('params_file')),
+        # Support a yaml params_file:
+        #   If a params file is to be used, it needs to be preceded by --ros-args
+        #   and followed by a trailing --
+        # This provides the leading --ros-args if a params_file is specified.
+        _conditional_command('ros-args', LaunchConfiguration('params_file')),
+        # These two lines provide the --params-file [params_file] arguments
+        _conditional_command('params-file', LaunchConfiguration('params_file')),
         LaunchConfiguration('params_file'),
-        _arg_command('', condition=LaunchConfiguration('params_file')),
+        # This provides the trailing -- if a params_file is specified.
+        _conditional_command('', LaunchConfiguration('params_file')),
         LaunchConfiguration('extra_gazebo_args'),
     ]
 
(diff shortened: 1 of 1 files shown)


gazebo_ros/launch/gzserver.launch.py
```diff
@@ -220,12 +232,16 @@ def _boolean_command(arg):
...
-def _arg_command(arg, condition=None):
-    if condition:
-        cmd = ['"--', arg, '" if "" != "', condition, '" else ""']
-    else:
-        cmd = ['"--', arg, '" if "" != "', LaunchConfiguration(arg), '" else ""']
+# Add string command and argument if not empty
+def _arg_command(arg, join_with=" "):
+    cmd = ['"--', arg, '" if "" != "', LaunchConfiguration(arg), '" else ""']
+    py_cmd = PythonExpression(cmd)
+    return (py_cmd, join_with, LaunchConfiguration(arg))
```
wjwwood: It might be harmless but in the case that the launch configuration is an empty string you'll still be adding the join with. I think anyway.

wjwwood: And actually if the join with is = that would be a problem no?

scpeters: I knew I was forgetting something 🤦

scpeters: I amended https://github.com/ros-simulation/gazebo_ros_pkgs/commit/eb1d266e70ba86fcb3f0aa7ed263d05d5650730b to https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1456/commits/efa494f3285bd8516b1e32af925a72559e5232a7, and I think this fixes the handling of `join_with` when the argument is empty

audrow (approved):
Looks good to me!

PR #1462, opened 2023-03-10
bugfix negative velocities in diff drive odom messages fixed

maxbader:
Hi
I fixed a bug in the diff_drive plugin. The odometry was treating negative velocities as positive.
Greetings


agoeckner:
This is a useful fix. Does anyone maintain this repository?

maxbader:
@agoeckner 
you can use my https://github.com/tuw-robotics/tuw_gazebo if you are just interested in the diff drive plugin

PR #1499, opened 2023-08-02
Add ROS Package Paths to Gazebo Resource Paths

DLu:
I've had it with meshes not loading when I spawn URDF into Gazebo. 

This PR aims to eventually 
 * Extract all the ROS package names from mesh filenames of the form `package://package_name/whatever`
 * Determine which paths aren't on the Gazebo Resource Path and add as necessary, using [server service calls](https://gazebosim.org/api/sim/7/classgz_1_1sim_1_1Server.html#details)

However, I'm still learning to do calls to gazebo services, so that part is a WIP and I would appreciate assistance with. 

(Note: I've developed this code with Foxy but the PR is for the `ros2` branch)

### DIFF
--- a/gazebo_ros/src/gazebo_ros_factory.cpp
+++ b/gazebo_ros/src/gazebo_ros_factory.cpp
@@ -14,8 +14,10 @@
 
 #include "gazebo_ros/gazebo_ros_factory.hpp"
 
+#include <ament_index_cpp/get_package_share_directory.hpp>
 #include <gazebo/common/Events.hh>
 #include <gazebo/common/Plugin.hh>
+#include <gazebo/msgs/MessageTypes.hh>
 #include <gazebo/physics/Entity.hh>
 #include <gazebo/physics/Model.hh>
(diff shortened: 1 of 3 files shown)


gazebo_ros/src/gazebo_ros_factory.cpp
```diff
@@ -383,6 +423,32 @@ void GazeboRosFactoryPrivate::AddNamespace(
...
+  std::unordered_set<std::string> paths;
+
+  /*gazebo::msgs::Empty request;
+  gazebo::msgs::StringMsg_V response;
+  gz_node_->Request("/gazebo/resource_paths/get", 5000, request, response);
+  for (const std::string& path : response)
+  {
+      paths.insert(path);
+  }*/
+
```
DLu, line 435: IfIHadAny

mjcarroll: Perfect.

DLu:
Superceded by https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1500

PR #372, opened 2015-12-08
Update gazebo_ros_planar_move

k-okada:
this may solves #121 and added linear/accel limit
as normal robot platform, it stops the robot if the cmd_vel did not
come for several seconds.

the default behavior is not changed.

gazebo_ros_planar_move: if cmd_vel does not come to joint_state_idel_sec, it stops the robot
gazebo_ros_planar_move : add velocity and acceleration limit
add recovery motion gazebo_ros_planar_move

PR #1504, opened 2023-08-21
Fix floating effect with gazebo_ros_planar_move

### LINKED ISSUE #121: Problem of the gazebo_ros_planar_move plugin
Hi!  I use ros hydro + gazebo 1.9.1, my os is Ubuntu 12.04 32bit. The gazebo_ros_planar_move plugin can not make the model in touch with the ground when moving, which means that the model just "floats" above the ground when it is put there (by spawn or manual translation in gazebo gui) with this plugin (when without this plugin, the model falls down quickly as expected). I suppose the problem comes from that the linear velocity of z direction of the model is set to zero, and it is also not suitable to set this velocity to some constant value, then the model would stick to the ground, and it is quite difficult for the model to move away. I understand that this is just a simple plugin for the model motion control in gazebo, but I still hope there could be some possible solutions for this issue. Thanks for any help!

fmrico:
Hi,

This PR is related to #372 and fixes #121.

It is ok to command planar speed commands to the robot (x, y, yaw), but if the world is not plane, the robot will float descending slopes.

Current behavior:
[Screencast 21-08-23 14:24:40.webm](https://github.com/pal-robotics-forks/gazebo_ros_pkgs/assets/3810011/33d88405-5423-4cc7-9af6-191b54d2cc52)

Behavior with this PR:
[Screencast 21-08-23 14:22:55.webm](https://github.com/pal-robotics-forks/gazebo_ros_pkgs/assets/3810011/51eafa65-3a6e-42a7-acf8-134d1fc4a032)

I hope it helps

### DIFF
--- a/gazebo_plugins/src/gazebo_ros_planar_move.cpp
+++ b/gazebo_plugins/src/gazebo_ros_planar_move.cpp
@@ -256,7 +256,7 @@ void GazeboRosPlanarMovePrivate::OnUpdate(const gazebo::common::UpdateInfo & _in
     double g_vel;
     ignition::math::Vector3d g = model_->GetWorld()->Gravity();
 
-    if(abs(g.Z()) > 0.0) {
+    if (abs(g.Z()) > 0.0) {
       g_vel = delta_pose.Z() / seconds_since_last_update + g.Z() * seconds_since_last_update;
       lin_vel.Z() = lin_vel.Z() + g_vel;
     }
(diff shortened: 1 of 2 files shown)


fmrico:
@ahcorde :pleading_face:

ahcorde (changes requested):
linters https://build.ros2.org/job/Ipr__gazebo_ros_pkgs__ubuntu_jammy_amd64/11/testReport/junit/gazebo_plugins/uncrustify/src_gazebo_ros_planar_move_cpp/

fmrico:
CI failed, but I don't see the error in the output :thinking:

PR #1531, opened 2024-06-25
Publish odom twist in robot-local coordinates

Martin-Oehler:
Hello everyone,
according to [nav_msgs/Odometry](https://docs.ros.org/en/noetic/api/nav_msgs/html/msg/Odometry.html), the twist message should be specified in the coordinate frame given by the (local) child_frame_id. Currently, the plugin publishes twist in the (global) header.frame_id coordinate frame.

I added a boolean parameter to change this behavior to the documented one. It defaults to false to not break legacy code.

Best regards,
Martin


ahcorde (approved):

hobbeshunter:
Thanks for the fix.

Since the default behavior is the wrong behavior (contrary to the documentation in nav_msgs/Odometry) I would appreciate a warning message.

Issue #1532, opened 2024-06-28
Release only gazebo_msgs on Jazzy


j-rivero:
Gazebo Classic is currently not available on the platforms supported by ROS 2 Jazzy (mainly Ubuntu Noble) and EOL in the next moth of January 2025 so the new release of this whole repository have little sense with this context.

However there is interest into having the `gazebo_msgs` release out for ROS 2 Jazzy since this "Gazebo" messages can be used a "simulation" messages by other simulations in the ROS ecosystem, OC3D for example. While the optimal solution would be to create a simulator neutral package for messages, the release of `gazebo_msgs` (with a warning message to avoid new Gazebo users to use them) can facilitate migrations. 

//cc @azeey @adamdbrw

j-rivero:
Released https://index.ros.org/p/gazebo_msgs/github-ros-simulation-gazebo_ros_pkgs/#jazzy

PR #1533, opened 2024-06-28
Add a warning about the use of gazebo_msgs without Gazebo Classic

j-rivero:
Related to https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1532.

Add a warning for ROS 2 distributions different than humble and iron (current supported ones) that keeps Gazebo users aware of the real goal of the package (being msgs for Gazebo classic).

### DIFF
--- a/gazebo_msgs/CMakeLists.txt
+++ b/gazebo_msgs/CMakeLists.txt
@@ -2,6 +2,20 @@ cmake_minimum_required(VERSION 3.5)
 
 project(gazebo_msgs)
 
+if(NOT DEFINED IGNORE_GAZEBO_MSGS_WARNING)
+  if(DEFINED ENV{ROS_DISTRO})
+    set(ROS_DISTRO $ENV{ROS_DISTRO})
+    if(NOT (ROS_DISTRO STREQUAL "iron" OR
+            ROS_DISTRO STREQUAL "humble"))
+      message(WARNING "This gazebo_msgs package hosts messages designed initially for Gazebo Classic "
+        "which is not available on ROS 2 Jazzy which is unavailable on ROS 2 Jazzy. "
+        "The new Gazebo uses the ros_gz bridge message vailable at:"
+        "https://github.com/gazebosim/ros_gz/tree/ros2/ros_gz_bridge#bridge-communication-between-ros-and-gazebo\n"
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

gazebo_msgs/CMakeLists.txt
```diff
@@ -2,6 +2,20 @@ cmake_minimum_required(VERSION 3.5)
 
 project(gazebo_msgs)
 
+if(NOT DEFINED IGNORE_GAZEBO_MSGS_WARNING)
+  if(DEFINED ENV{ROS_DISTRO})
+    set(ROS_DISTRO $ENV{ROS_DISTRO})
+    if(NOT (ROS_DISTRO STREQUAL "iron" OR
+            ROS_DISTRO STREQUAL "humble"))
+      message(WARNING "This gazebo_msgs package hosts messages designed initially for Gazebo Classic "
+        "which is not available on ROS 2 Jazzy which is unavailable on ROS 2 Jazzy. "
```
azeey: Is this what you meant?
```suggestion
        "which is not available on ROS 2 Jazzy which is unavailable on Ubuntu Noble. "
```

gazebo_msgs/CMakeLists.txt
```diff
@@ -2,6 +2,20 @@ cmake_minimum_required(VERSION 3.5)
 
 project(gazebo_msgs)
 
+if(NOT DEFINED IGNORE_GAZEBO_MSGS_WARNING)
```
azeey: Most users will not be building this package from source so they will not be seeing this warning. Ideally, we would print the warning when the downstream application is doing a `find_package`.

azeey: Chris suggested using `CONFIG_EXTRAS` with `ament_package` , e.g https://github.com/ros2/rclcpp/blob/rolling/rclcpp_components/CMakeLists.txt#L176

j-rivero: good catch. [7037648](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1533/commits/70376483630d04361461f4e00e13a1920ac48a99) Still need to test it, closing by today.

azeey: Okay, thanks. I'd be okay with removing this and only doing the `CONFIG_EXTRAS` thing, but I'll leave it up to you.

azeey (approved):

